PR #3, opened 2018-04-04
[rce] Fix relValue passing

muteor:
Was trying out the link tool and noticed that you could not disable the nofollow option because `relValue` was not passed into the LinkPanel props.

This PR, adds `relValue` to anywhere that uses `anchorTarget` as they are always used together anyway.

These two props could probably be refactored later to be part of config rather than the main props, I didn't do that for now to keep the change small. Another option would be to combine the prop into something like:

```
linkConfig: { anchorTarget, relValue }
```

This could then live in either the main editor props interface or the config interface. Main interface may make sense because this setting is applicable to all/most blocks.


bchmn:
we definitely want to move those into the config prop further down the line, but this works for now.  thanks!

PR #11, opened 2018-04-18
Handle keyboard shortcuts

artur-charcenko:
Adds support for default keyboard shortcuts like CMD+B - bold.
[Full list](https://github.com/facebook/draft-js/blob/master/src/component/utils/getDefaultKeyBinding.js#L67)

### DIFF
--- a/rich-content-editor/src/RichContentEditor/RichContentEditor.jsx
+++ b/rich-content-editor/src/RichContentEditor/RichContentEditor.jsx
@@ -1,7 +1,7 @@
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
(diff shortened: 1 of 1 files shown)


rich-content-editor/src/RichContentEditor/RichContentEditor.jsx
```diff
@@ -14,6 +14,8 @@ import { normalizeInitialState } from '~/Utils';
 import styles from '~/Styles/rich-content-editor.scss';
 import draftStyles from '~/Styles/draft.scss';
 
+const IGNORED_COMMANDS = ['code'];
```
artur-charcenko: Ignoring `code` shortcut for now, because we don't have styles for code block

PR #12, opened 2018-04-19
Avoid mutating entity data

artur-charcenko:
When `pubsub.update('componentData', data);` is called it mutates entity data and causes autosave related problems in blog.


lxgreen:
thanks!

PR #32, opened 2018-05-23
[rce] improve(props passing) Pass restProps to the underlying editor

muteor:
I need onFocus/onBlur props to be passed to the editor, went for just passing restProps through. Can add them explicitly if you want or use lodash/omit to filter them.

### DIFF
--- a/packages/editor/src/RichContentEditor/RichContentEditor.jsx
+++ b/packages/editor/src/RichContentEditor/RichContentEditor.jsx
@@ -292,7 +292,32 @@ class RichContentEditor extends Component {
   };
 
   renderEditor = () => {
-    const { helpers, placeholder, editorKey } = this.props;
+    const { 
+      helpers, 
+      editorKey, 
+      tabIndex,
+      placeholder,
+      spellCheck,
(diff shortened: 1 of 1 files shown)


muteor:
Ok have updated with the known set of props based on the typescript interface, I have left out the handle* props for now.

packages/editor/src/RichContentEditor/RichContentEditor.jsx
```diff
@@ -302,12 +327,30 @@ class RichContentEditor extends Component {
...
-        placeholder={placeholder || ''}
-        readOnly={!!readOnly}
-        helpers={helpers}
-        spellCheck
         editorKey={editorKey}
         keyBindingFn={keyBindingFn}
+        helpers={helpers}
+        tabIndex={tabIndex}
+        placeholder={placeholder || ''}
+        readOnly={!!readnly}
```
bchmn: typo - should be `readOnly`

muteor: hehe how did that even happen :)

Fixed.

PR #33, opened 2018-05-28
Finalize html plugin

artur-charcenko:


packages/common/src/Styles/global.scss
```diff
@@ -41,6 +41,13 @@
   display: none;
 }
 
+.pluginContainer,
+.pluginContainerReadOnly {
+  @include phone {
+    max-width: 100%;
```
artur-charcenko: @bchmn I wanted to apply `max-width: 100%` only for mobile, but I guess it's not the right way to do it

packages/common/src/Styles/text-input.scss
```diff
@@ -0,0 +1,30 @@
+@import '../../../common/src/Styles/palette';
```
bchmn: `@import palette`

packages/common/src/Utils/urlValidators.js
```diff
@@ -16,3 +16,5 @@ export const isVideoUrl = url => [isYoutube, isFacebook, isVimeo].some(f => f(ur
 export const isValidUrl = url => linkify.test(url);
 
 export const normalizeURL = url => (linkify.match(url) || [{}])[0].url;
+
+export const isHttpsUrl = url => /^https:/.test(url);
```
bchmn: shouldn't this also test if it's a validUrl?
`https://foo` is `true` at the moment despite it not being a valid url.
I think it should either check for url validity:
`export const isHttpsUrl = url => isValidUrl(url) && /^https:/.test(url);`
or rename it to `isHttps` or `startsWithHttps`

packages/common/src/Base/createBaseToolbar.jsx
```diff
@@ -283,7 +287,7 @@ export default function createToolbar({ buttons, theme, pubsub, helpers, isMobil
           );
         default:
           return (
-            <BaseToolbarButton
```
bchmn: I don't understand this change, `BaseToolbarButton` is never rendered now.

packages/common/src/Utils/urlValidators.js
artur-charcenko: I need to show different errors for when url is invalid and when it doesn't start with https. So I'm gonna rename it to `startsWithHttps`

PR #50, opened 2018-06-27
mentions plugin

vybu:
Implementation of `"dumb"` mentions plugin, which uses `draft-js-mention-plugin` as a base. Doesn't include data fetching and navigation. However idea is to provide helpers for TPAs which can be used in configuration of plugin to get member fetching and navigation capabilities.


packages/plugin-mentions/src/tpa-utils/membersAreaNavigationHandler.js
```diff
@@ -0,0 +1,13 @@
+/* eslint-disable no-console */
+export const createMembersAreaNavigationHandler = (wixSdk, santaMembersAppDefId) => mention =>
```
bchmn: I'm not sure what this is for but I'm pretty sure it doesn't belong in the plugin.

packages/plugin-mentions/src/createMentionsPlugin.js
```diff
@@ -0,0 +1,78 @@
...
+}
+*/
+
+const defaultSettings = {
+  mentionPrefix: '@',
+  mentionTrigger: '@',
+  getMentionLink: () => '#',
+};
+
+const createMentionsPlugin = (config = {}) => {
```
bchmn: `createMentionPlugin` is already defined here by the external plugin, please rename to `createExternalMentionsPlugin`

PR #69, opened 2018-07-26
Adds basic Code Block plugin

sarunas:


lxgreen:
Hi
Everything is cool, just one note -- as far as I understand, you've added the `DefaultInlineToolbarButton` component to the `common` package similarly to the `LinkButton` found in `common/Components`. However, there's a spacial reason why the `LinkButton` is found there -- every plugin that needs a link, reuses it (at the moment, these are `image-plugin`, 'gallery-plugin`, and `link-plugin`). In the case of `code-block-plugin`, it is the only `DefaultInlineToolbarButton` consumer at the moment. So, there's no real reason to put it in `common` package. Instead, it should be implemented inside your plugin package.
The ability to add a plugin inline toolbar button was developed a couple of days ago, so I understand a need in common InlineToolbarButton for plugin development reuse. We'll add it in the following releases.

sarunas:
IMHO, `InlineToolbarButton` should reside in common package, cause the styling and components from `common`. Anyway, we can refactor`LinkButton` to use `InlineToolbarButton` internaly - i think it makes more sense. Cause if I make my own `CodeBlockButton` I still need to reference styles and components from `common` package.

lxgreen:
Hey,

Thanks for the `InlineToolbarButton` and `LinkButton` refactoring!

_#sarunas_the_great_contributor_

lxgreen:
BTW, at the moment, all the PRs should be merged to the `develop` branch. We are working on some Open Source practices adoption; later, the PRs will go directly to `master`.

PR #140, opened 2018-09-17
Wix rich content plugin giphy

FerasAMasood:


packages/plugin-giphy/src/toolbar/giphyApiInputModal.jsx
```diff
@@ -36,7 +36,7 @@ export default class GiphyApiInputModal extends Component {
     const { styles } = this;
     const { t, theme } = this.props;
     const searchTag = this.state.searchTag;
-    let navbar;
+    let navbar = null;
     const backButton = <div className={styles.backButton} onClick={this.onCloseRequested} role="button" tabIndex="0" onKeyPress={null} />;
```
FerasAMasood: please change this element into <a> or <button> to get rid of onkeypress forced by eslint

packages/plugin-giphy/statics/styles/giphy-api.scss
```diff
@@ -65,26 +69,24 @@
...
       margin-right: 0px;
      }
     cursor: pointer;
   }
   
-  .spinner{
-      @include phone{
+  .spinner {
+      @include phone {
         padding-top: 35vh;
```
FerasAMasood: please position in relevant to its container and not the view port

packages/plugin-giphy/statics/styles/giphy-api-input-modal.scss
```diff
@@ -7,14 +7,19 @@
...
 }
 
+.group_form{
+  border-bottom: 2px solid #cccccc;
+  width: 227px;
+  padding-bottom: 5px;
+}
 .search {
-  width: 90%;
+  width: auto;
```
FerasAMasood: Text input shall not have an auto width, it should have the maximum width you can give

packages/plugin-giphy/statics/styles/giphy-api-input-modal.scss
```diff
@@ -60,10 +66,12 @@
   border-left: solid 1px #ffffff;
```
FerasAMasood: what do we need a left border here?

packages/plugin-giphy/statics/styles/giphy-api-input-modal.scss
```diff
@@ -60,10 +66,12 @@
...
 }
 .giphyUrlInputModal_textInput {
-  top: 0;
-  left: 0;
+  // top: 0;
+  // left: 0;
   // margin: 8px 0 6px 0;
```
FerasAMasood: Please don't push a commented code, if you don't need it remove it

packages/plugin-giphy/statics/styles/giphy-api-input-modal.scss
```diff
@@ -60,10 +66,12 @@
...
-  margin-bottom: 6px;
+  border-bottom: 2px solid #cccccc;
+  width: 229px;
+  padding-bottom: 5px;
   input[type="url"] {
```
FerasAMasood: do we have an input of type url?

packages/plugin-giphy/statics/styles/giphy-api-input-modal.scss
```diff
@@ -11,6 +11,11 @@
   border-bottom: 2px solid #cccccc;
```
FerasAMasood: choose a more descriptive name

packages/plugin-giphy/statics/styles/giphy-api-input-modal.scss
```diff
@@ -11,6 +11,11 @@
...
   width: 227px;
   padding-bottom: 5px;
+
+  @include phone {
+    width: 98%;
+
```
FerasAMasood: why 98%?

FerasAMasood:
@lxgreen flay-out modal was added to `common` to be a new type of modal and become a reusable component, the part that is specific to  `giphy` plugin is the component rendered inside it and there logic is contained in  `giphy`plugin.
Also, as  `giphy` plugin we don't have access to button click event, unless we had our code in  `common`, event is needed to calculate the position of the modal.

lxgreen:
@FerasAMasood, I hear what you say. However, the fly-out modal should be a common modal type only if more plugins are going to use it. At the moment, when we have only a single scenario of fly-out, we cannot predict the future requirements of the common fly-out modal. 

Regarding the click event, its only used to get the `event.target` which is the button element itself. The button element can be found without the click event data, for example by using the `findDOMNode` from `react-dom` package.

nzanabeet:
@lxgreen , Can you please navigate us to a start point , Let's say I want to move the fly out calculations to giphy plugin, We know that the `insert-buttons.js` that is passed to `createBasePlugin` takes a `modalElement', now the modal element (fly out modal) position should be calculated based on click event on the Footer Toolbar insert button  (& replace button of the inline toolbar in another scenario) or the giphy button position itself in respect to the editor. How exactly can I access such event or position from giphy plugin side without going into common an manipulate the `onClick`  function that is attached to the insert button !  (More details please ..)

lxgreen:
@nzanabeet, you can use React `ref` that allows you to access the button DOM element. Once you got it, you can call `getBoundingClientRect()` to get the button coordinates and size. Another (less preferred) option  is to use `findDOMNode` from `react-dom` library to access the button DOM element.

PR #154, opened 2018-10-14
modalStylesFn and inline style for modal

lxgreen:

### DIFF
--- a/packages/plugin-video/src/toolbar/insert-video-buttons.jsx
+++ b/packages/plugin-video/src/toolbar/insert-video-buttons.jsx
@@ -21,7 +21,7 @@ export default ({ helpers, t }) => {
       componentData: DEFAULTS,
       toolbars: [TOOLBARS.FOOTER, TOOLBARS.SIDE],
       modalElement: VideoURLInputModal,
-      modalStyles: getModalStyles({ customStyles: modalCustomStyle, fullScreen: false, inline: false }),
+      modalStyles: getModalStyles({ customStyles: modalCustomStyle, fullScreen: false }),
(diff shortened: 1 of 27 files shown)


packages/plugin-video/src/toolbar/insert-video-buttons.jsx
```diff
@@ -21,7 +21,13 @@ export default ({ helpers, t }) => {
       componentData: DEFAULTS,
       toolbars: [TOOLBARS.FOOTER, TOOLBARS.SIDE],
       modalElement: VideoURLInputModal,
-      modalStyles: getModalStyles({ customStyles: modalCustomStyle, fullScreen: false }),
+      modalStyles: getModalStyles({ customStyles: modalCustomStyle, fullScreen: false, inline: false }),
```
bchmn: `getModalStyles` has `inline: false` by default, this change is unnecessary

packages/plugin-video/src/toolbar/insert-video-buttons.jsx
```diff
@@ -21,7 +21,13 @@ export default ({ helpers, t }) => {
...
+      // modalStylesFn: ({ buttonRef }) => {
+      //   const modalStyles = getModalStyles({ customStyles: modalCustomStyle, fullScreen: false });
+      //   const { top, left } = buttonRef.getBoundingClientRect();
+      //   const modalTop = top > 172 ? top - 172 : top + 172;
+      //   return { ...modalStyles, content: { ...modalStyles.content, top: modalTop, left, margin: 0, position: 'absolute' } };
+      // },
```
bchmn: why do we need this comment here?

PR #157, opened 2018-10-15
[viewer] add RTL support

audriusk-wix:


packages/viewer/statics/rich-content-viewer.scss
```diff
@@ -182,4 +182,7 @@
 
 .preview {
   white-space: pre-wrap !important;
+  &.rtl {
```
lxgreen: please make it flat (yep, meticulous)

PR #179, opened 2018-11-01
Dynamic html-component width

lxgreen:


CHANGELOG.md
```diff
@@ -10,6 +10,19 @@
 
 ## [Unreleased]
 
+### :rocket: New Feature
+
+- `common`
+  - [#179](https://github.com/wix-incubator/rich-content/pull/179) `mapStoreDataToButtonProps` API exposed (to be used in toolbar button structures)
+  - [#179](https://github.com/wix-incubator/rich-content/pull/179) `width` and `height` props added to the plugin config
```
bchmn: is this a change to `common`?

packages/plugin-html/src/HtmlComponent.jsx
```diff
@@ -1,41 +1,84 @@
...
+        this.props.componentData.config.height = this.props.settings.height;
+      } else {
+        this.props.componentData.config.height = INIT_HEIGHT;
+      }
+    }
+  }
+
+  render() {
+    const { styles, props } = this;
+    validate(props.componentData, schema);
```
bchmn: not specific to this PR but perhaps we should skip validation in production? WDYT?

CHANGELOG.md
lxgreen: yup, the values are used in [baseComponent](https://github.com/wix-incubator/rich-content/blob/57d28d0b262849f7265d4d0d6b0976d1789188ba/packages/common/src/Base/createBaseComponent.jsx#L238)

packages/plugin-html/src/HtmlComponent.jsx
lxgreen: hmm good idea. usually all the data changes start when we work on plugin, and it is redundant in production, as data and schema are aligned by then. agree to skip in prod

PR #192, opened 2018-11-18
Giphy viewer fix

nzanabeet:
- Remove container styles to keep width to hight ratio of each giphy

### DIFF
--- a/packages/plugin-giphy/src/giphy-component.jsx
+++ b/packages/plugin-giphy/src/giphy-component.jsx
@@ -60,7 +60,6 @@ GiphyComponent.propTypes = {
   componentData: PropTypes.object.isRequired,
   blockProps: PropTypes.object.isRequired,
   onClick: PropTypes.func.isRequired,
-  theme: PropTypes.object.isRequired
 };
 
 export { GiphyComponent as Component };
--- a/packages/plugin-giphy/src/giphy-component.jsx
+++ b/packages/plugin-giphy/src/giphy-component.jsx
@@ -1,6 +1,5 @@
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
-import classNames from 'classnames';
 import { mergeStyles } from 'wix-rich-content-common';
 import GiphyViewer from './giphy-viewer';
 import { GIPHY_TYPE } from './constants';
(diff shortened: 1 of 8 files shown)


packages/plugin-giphy/src/giphy-component.jsx
```diff
@@ -45,15 +44,12 @@ class GiphyComponent extends Component {
   };
 
   render() {
-    const { styles } = this;
-    const { className, onClick } = this.props;
-    const containerClassNames = classNames(styles.giphy_container, className || '');
```
bchmn: you're still importing css and using `mergeStyles` but you no longer consume the styles. I'm assuming you're using the styles and theme in the viewer portion of this plugin and if that's the case you no longer need the css import or the call to ` mergeStyles`

packages/plugin-giphy/src/giphy-viewer.jsx
```diff
@@ -21,7 +21,7 @@ class GiphyViewer extends Component {
   render() {
     const { componentData, theme, ...rest } = this.props; // eslint-disable-line no-unused-vars
```
bchmn: `...rest` is never used, either use it or remove the variable

packages/plugin-giphy/src/index.js
```diff
@@ -2,3 +2,4 @@ export { createGiphyPlugin } from './createGiphyPlugin';
 export { default as GiphyViewer } from './giphy-viewer';
 export { GIPHY_TYPE } from './constants';
 export { Modals, ModalsMap } from './modals';
+export { typeMapper as giphyTypeMapper } from './typeMapper';
```
bchmn: typeMapper should be exported from your `viewer.js` not the main index file (see other plugins for reference) - remove the export from here.

packages/plugin-giphy/src/giphy-component.jsx
bchmn: you also no longer use/need `classNameStrategies.js` - remove it

packages/plugin-giphy/statics/styles/giphy-viewer.scss
```diff
@@ -1,22 +1,10 @@
...
-  left: 0;
-  width: 100%;
-  height: 100%;
-  border: 0;
-  z-index: 0;
+ display: block;
+ width: 100%;
 }
 
 .giphy_container {
```
bchmn: this class is never applied, there's no point in leaving it blank in the css file - remove it.

packages/plugin-giphy/src/giphy-component.jsx
```diff
@@ -65,13 +58,9 @@ class GiphyComponent extends Component {
...
 GiphyComponent.propTypes = {
   componentData: PropTypes.object.isRequired,
-  componentState: PropTypes.object.isRequired,
-  store: PropTypes.object.isRequired,
   blockProps: PropTypes.object.isRequired,
   onClick: PropTypes.func.isRequired,
-  className: PropTypes.string.isRequired,
-  theme: PropTypes.object.isRequired,
-  t: PropTypes.func
+  theme: PropTypes.object.isRequired
```
bchmn: this prop is never used - remove it

PR #202, opened 2018-12-05
Adds code block viewer support

sarunas:

### DIFF
--- a/packages/viewer/src/Preview.jsx
+++ b/packages/viewer/src/Preview.jsx
@@ -153,10 +153,19 @@ Preview.propTypes = {
   theme: PropTypes.object,
   isMobile: PropTypes.bool,
   textDirection: PropTypes.oneOf(['rtl', 'ltr']),
-  decorators: PropTypes.arrayOf(PropTypes.shape({
-    component: PropTypes.func.isRequired,
-    strategy: PropTypes.func.isRequired,
-  })),
+  decorators: PropTypes.arrayOf(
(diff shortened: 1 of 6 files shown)


bchmn:
@sarunas does this resolve performance issues?

sarunas:
Although I have no data to prove - the decorator itself should be more performant - cause it will use static component (will not create new one for each decoration). Anyway this pull is part of another effort - migrating to viewer.

packages/viewer/src/Preview.jsx
```diff
@@ -153,10 +153,19 @@ Preview.propTypes = {
   theme: PropTypes.object,
   isMobile: PropTypes.bool,
   textDirection: PropTypes.oneOf(['rtl', 'ltr']),
-  decorators: PropTypes.arrayOf(PropTypes.shape({
-    component: PropTypes.func.isRequired,
-    strategy: PropTypes.func.isRequired,
-  })),
+  decorators: PropTypes.arrayOf(
+    PropTypes.oneOfType([
+      PropTypes.shape({
```
bchmn: when is this data structure used?

sarunas: Its used here:
https://github.com/wix-incubator/rich-content/pull/202/files#diff-0c1e1a195837ab2a16e4ec14542dd969R83

And its a definition of:
https://github.com/facebook/draft-js/blob/master/src/model/decorators/DraftDecoratorType.js

PR #204, opened 2018-12-10
Mentions viewer

sarunas:


sarunas:
@lxgreen @bchmn please write feedback regarding viewer plugin approach

PR #205, opened 2018-12-16
block headers

bchmn:


packages/common/src/Utils/getNormalizedHeaderBlock.js
```diff
@@ -1,64 +1,68 @@
...
-  };
-  inlineStyleRanges.push(...getCombinedRanges(wrapRange, existingInlineRanges.headerRanges));
+  const { headerRanges, otherRanges } = getInlineStyleRanges(block);
+  const inlineStyleRanges = [...otherRanges];
+  if (isEmpty(headerRanges)) {
+    return block;
+  } else {
 
-  return { ...block, type: 'unstyled', inlineStyleRanges };
+    console.log(block.text, block.text.length, length); //eslint-disable-line
```
artur-charcenko: Is it intentional?

bchmn: oops

packages/common/src/Utils/getNormalizedHeaderBlock.js
```diff
@@ -1,64 +1,66 @@
...
-  if (combined.length > 0) {
-    const lastCombined = combined[combined.length - 1];
-    const lastCombinedEnd = lastCombined.offset + lastCombined.length;
-    if (lastCombinedEnd < wrapRange.length) {
-      combined.push({ ...wrapRange, offset: lastCombinedEnd, length: wrapRange.length - lastCombinedEnd });
+  } else {
+    const headerLength = headerRanges.reduce((headerLength, { length }) => headerLength += length, 0); //eslint-disable-line no-param-reassign
+    const containsOnlyHeaders = headerLength === blockText.length;
+    if (containsOnlyHeaders) {
+      if (headerRanges.some(({ style }) => style === INLINE_HEADING.THREE)) {
```
artur-charcenko: I would extract `headerRanges.some(({ style }) => style === INLINE_HEADING.THREE)` into a separate function. It would be more readable.

artur-charcenko: It also looks like there's no need to separate header ranges from the rest of inline ranges. You can do `headerRanges.some(({ style }) => style === INLINE_HEADING.THREE)` on all ranges. Or am I missing something?

packages/viewer/src/Preview.jsx
```diff
@@ -44,9 +43,9 @@ const getList = (ordered, mergedStyles, textDirection) =>
           const dataEntry = blockProps.data.length > i ? blockProps.data[i] : {};
           return withTextAlignment(
             <li className={mergedStyles[`${className}List`]} key={blockProps.keys[i]}>
-              <p className={mergedStyles.elementSpacing}>
+              <div className={mergedStyles.elementSpacing}>
```
artur-charcenko: Shouldn't we render paragraphs inside list items?

PR #212, opened 2018-12-18
optional auto link

shaulgo:


packages/plugin-link/src/LinkDecorator.jsx
```diff
@@ -33,4 +29,4 @@ Link.propTypes = {
   settings: PropTypes.object,
 };
 
-export { findLinkEntities as Strategy, Link as Component };
+export { Link as Component };
```
bchmn: this file no longer contains a decorator, only a component - please rename to `LinkComponent.jsx`

shaulgo: Done

PR #224, opened 2019-01-07
prevent hashtag links in code-blocks

shaulgo:


packages/plugin-hashtag/src/decorator.jsx
```diff
@@ -6,7 +6,8 @@ import styles from '../statics/hashtag.scss';
 
 const findHashtagEntities = (contentBlock, callback) => {
   const text = contentBlock.getText();
-  if (!text || !text.match(hashtagRegexes.hashSigns)) {
+  const type = contentBlock.getType();
+  if (type === 'code-block' || !text || !text.match(hashtagRegexes.hashSigns)) {
```
bchmn: this solves the issue specifically for the code-block plugin, another plugin which manipulates text would encounter the same issue.  lets check for valid types instead:
```
const validBlockTypes = ['unstyled', 'ordered-list-item', 'unordered-list-item', 'blockquote', 'header-one', 'header-two', 'header-three']; //this should be defined outside the function to avoid unnecessary reallocations
if (!validBlockTypes.includes(type) || !text || !text.match(hashtagRegexes.hashSigns)) {
  return [];
}
```

shaulgo: I'm not sure which solution is better. In both ways, we will forget to add types to the list. 
But, seems like most types are good with it. So, there will be fewer times that we forget and need to make changes. 
Currently, are there any other types that have this problem? It seems that it should grow into a list of **in**valid types.

PR #234, opened 2019-01-18
[video] feat(upload thumbnail)

mantaspa:


CHANGELOG.md
```diff
@@ -10,6 +10,11 @@
 
 ## [Unreleased]
 
+### :rocket: New Feature
+
+- `video`
+  - upload thumbnail
```
bchmn: Please add a link to the PR like so:
https://github.com/wix-incubator/rich-content/blame/master/CHANGELOG.md#L17

mantaspa: > When is the thumbnail used? I see it's added to componentData but when is it displayed?

After updateEntity invocation it gets stored in DB. It is more like metadata, which we use for a post cover image. Also this thumbnail could be later used as a react-player thumbnail by passing it via 'light' prop.

bchmn: understood
how does one upload/choose a thumbnail? I don't see where thumbnail is set in state

mantaspa: At the moment media manager returns video Url and 4 images, so one of them might be used as a thumbnail.

mantaspa: Isn't it responsibility of the consumer to implement file upload for live site, which will handle file uploading and thumbnail generation?

bchmn: it is the consumers responsibility I'm just trying to understand if this is a breaking change.  If the consumer doesn't change anything and updates to this version of the video plugin will anything break for them?

mantaspa: If a consumer doesn't use relative video URL functionality which we released on Thursday, then it does not affect a consumer. If a consumer uses relative video URL functionality, he will get warning (schema validation) because thumbnail field is missing.

mantaspa: Sorry, the consumer shouldn't get a warning, because by  default properties in schema are not required.

bchmn: got it, thanks for the explanation 👍

PR #270, opened 2019-02-27
[*] configure dotenv for rollup so classnames are not lost in development

ehabshs:
Hey,

generateScopedName in rollup.config.js depends on an environment variable to  generate classnames, however rollup is not yet configured to read the env variables so we are losing the classnames in development, i have configured dotenv so rollup can load the env variables.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -70,14 +70,14 @@
 			}
 		},
 		"@babel/generator": {
-			"version": "7.3.0",
-			"resolved": "https://registry.npmjs.org/@babel/generator/-/generator-7.3.0.tgz",
-			"integrity": "sha512-dZTwMvTgWfhmibq4V9X+LMf6Bgl7zAodRn9PvcPdhlzFMbvUutx74dbEv7Atz3ToeEpevYEJtAwfxq/bDCzHWg==",
+			"version": "7.3.4",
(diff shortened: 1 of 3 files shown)


bchmn:
I don't understand why this is necessary, the `.env` file doesn't include `NODE_ENV`, it only includes the api keys so there's no point loading it in the rollup config.
rollup.config.js get the `NODE_ENV` since we run it using npm scripts:
https://github.com/wix-incubator/rich-content/blob/master/packages/common/package.json#L28

ehabshs:
Hey,

Sorry it is my fault, indeed it is not needed, what happened that on my device and bashar's device the watch script was building for production so the classnames were cut . For some reason our package jsons were not configured correctly, the watch was building for production, so i thought that it is because the env variables were not loaded correctly. 

Btw when i build a new plugin where i should get my template for the package json? should i get it from the packages/common? or just from any other plugin?

bchmn:
No problem, glad it's sorted.
The divider plugins [package.json](https://github.com/wix-incubator/rich-content/blob/master/packages/plugin-divider/package.json) is a great starting point since it has 0 dependencies other than `wix-rich-content-common`.  By the way, we plan on releasing a scaffolding for plugins in the near future.

PR #274, opened 2019-03-10
Core/context

lxgreen:
This PR contains the following changes:

- [common]
  - `Context` component is implemented as a `React.createContext()` wrapper. The `Context.Provider` API stores the contextual data such as `theme`, `t`, `config`, etc. The `Context.type` and `Context.Consumer` APIs allow to consume the contextual data within class or stateless function components, respectively.

- [editor]
  - `Context.Provider` is integrated into `RCE`

- [viewer]
  - `Context.Provider` is integrated into `RCV`

- [plugins]
  - every plugin's `Component` and `Viewer` components now rely on Context

- [examples/viewer]
  - `giphy`, `map`, and `soundCloud` examples added

- [map]
  - redundant code removed


packages/plugin-divider/src/components/divider-component.jsx
```diff
@@ -1,18 +1,17 @@
...
-import { mergeStyles, validate } from 'wix-rich-content-common';
+import { mergeStyles, validate, Context } from 'wix-rich-content-common';
 import isEqual from 'lodash/isEqual';
 import { getType, getConfig } from '../toolbar/selectors';
 import DividerLine from './divider-line';
 import schema from '../../statics/data-schema.json';
 import styles from '../../statics/styles/divider-viewer.scss';
 
-class DividerComponent extends PureComponent {
+class DividerComponent extends Component {
```
bchmn: `PureComponent` doesn't have access to `context`?

packages/plugin-divider/src/components/divider-component.jsx
```diff
@@ -49,9 +48,9 @@ class DividerComponent extends PureComponent {
           type={type}
           width={editorWidth}
           size={size}
-          isMobile={this.props.isMobile}
           alignment={alignment}
           styles={styles}
+          contextType={DividerComponent.contextType || Context.type}
```
bchmn: why is the `||` necessary here?
`DividerComponent.contextType = Context.type;`

packages/plugin-divider/src/components/divider-component.jsx
lxgreen: i'll validate it, maybe it is unnecessary. There was an issue with dynamic context passing to some Viewer components. Basically, the context relies on "pre-typed" component composition,  but the RCV renders all the components in a generic `<Component ...>` way. So, when the contextType is assigned in runtime like `Component.contextType = Context.type` it works with class-based components, but when they are stateless, or contain any stateless components like DividerLine, this assignment above has no effect, and the contextType has to be passed  as a prop to the stateless component.

packages/plugin-divider/src/components/divider-component.jsx
lxgreen: See my previous comment. Within the editor, the line `DividerComponent.contextType = Context.type;` does work (contextType assigned at the first script evaluation), however the viewer has to assign the contextType in runtime. So, the contextType is found on one of the two.

packages/plugin-divider/src/components/divider-component.jsx
lxgreen: PureComponent is back

PR #279, opened 2019-03-18
[infra - mobileToolbar] change mobile toolbar from fixed to sticky

ehabshs:
This fixes the mobileToolbar position bug, by changing the position of the toolbar to sticky

Thanks


bchmn:
fixed is good
*DO NOT MERGE*

PR #297, opened 2019-05-01
Lists - add rtl support

shaulgo:
**breaking change!!!!!** Might need to fix theming of the list's elements.


shaulgo:
breaking change!!!!! Might need to fix theming of the list's elements.

PR #311, opened 2019-06-05
Fix link underline + versioning

lxgreen:
### :rocket: New Feature
- `common`
  - versioning utils added: `getCurrent`, `compare`, `evaluate`
  - `normalizeInitialState` refactored; now considers versions; adds the underline style to links with version < 3.4.7
- `link`
  - underline style can be removed from link
### :bug: Bug Fix
- `text-color`
  - color applied to link underline decoration correctly


lxgreen:
waiting for #313

PR #325, opened 2019-07-01
[file upload plugin] allow for async file url resolution

vybu:
We want to be able to resolve file url on demand, this way we can handle private file downloading.

On file upload we will not save file url to content state, but some other metadata which will be used by url resolver to resolve it on demand (while also checking if user has necessary permissions)

### DIFF
--- a/packages/plugin-file-upload/src/file-upload-viewer.jsx
+++ b/packages/plugin-file-upload/src/file-upload-viewer.jsx
@@ -15,6 +15,7 @@ class FileUploadViewer extends PureComponent {
     super(props);
     const { componentData } = props;
     validate(componentData, schema);
+    this.fileDownloadIframeId = `${Date.now()}`;
   }
 
   componentWillReceiveProps(nextProps) {
@@ -88,7 +89,7 @@ class FileUploadViewer extends PureComponent {
 
     const resolveFileUrl = () => {
       settings.resolveFileUrl(componentData).then(resolveFileUrl => {
-        document.getElementById('fileDownloadIframe').src = resolveFileUrl;
+        document.getElementById(this.fileDownloadIframeId).src = resolveFileUrl;
         this.setState({ resolveFileUrl });
       });
     };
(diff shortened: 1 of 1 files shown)


packages/plugin-file-upload/src/file-upload-viewer.jsx
```diff
@@ -44,37 +49,87 @@ class FileUploadViewer extends PureComponent {
...
 
+  renderFileUrlResolver() {
+    const { error, componentData, settings } = this.props;
+
+    if (error) {
+      return null;
+    }
+
+    const resolveFileUrl = () => {
+      settings.resolveFileUrl(componentData).then(resolveFileUrl => {
```
lxgreen: requires 'settings' and 'resolveFileUrl' validation before call

packages/plugin-file-upload/src/file-upload-viewer.jsx
```diff
@@ -84,10 +139,12 @@ FileUploadViewer.propTypes = {
   isLoading: PropTypes.bool.isRequired,
   componentData: PropTypes.object.isRequired,
   error: PropTypes.string,
+  settings: PropTypes.object.isRequired,
```
lxgreen: optional

packages/plugin-file-upload/src/file-upload-viewer.jsx
```diff
@@ -44,37 +49,87 @@ class FileUploadViewer extends PureComponent {
...
+  }
+
+  renderAutoDownloadIframe() {
+    const withFileUrlResolver = this.props.settings.resolveFileUrl;
+
+    if (!withFileUrlResolver) {
+      return null;
+    }
+
+    return <iframe id={this.fileDownloadIframeId} style={{ display: 'none' }} title="file" />;
```
lxgreen: please consider using ref instead id

vybu: I tried but for some reason download wouldn't trigger :thinking:

PR #335, opened 2019-07-23
File upload design update

vybu:
Update styling of upload plugin.

upload-restyled

### DIFF
--- a/packages/plugin-file-upload/src/typeMapper.js
+++ b/packages/plugin-file-upload/src/typeMapper.js
@@ -1,8 +1,10 @@
 import FileUploadViewer from './file-upload-viewer';
 import { FILE_UPLOAD_TYPE } from './types';
+import { containerClassName } from './classNameStrategies';
 
 export const typeMapper = () => ({
   [FILE_UPLOAD_TYPE]: {
     component: FileUploadViewer,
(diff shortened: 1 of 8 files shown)


packages/plugin-file-upload/src/typeMapper.js
```diff
@@ -1,10 +1,8 @@
 import FileUploadViewer from './file-upload-viewer';
 import { FILE_UPLOAD_TYPE } from './types';
-import { containerClassName } from './classNameStrategies';
 
 export const typeMapper = () => ({
   [FILE_UPLOAD_TYPE]: {
     component: FileUploadViewer,
-    classNameStrategies: { container: containerClassName },
```
vybu: don't know what's the purpose of this, but it was causing double container class to be added.

lxgreen: yep, this caused the double class adding. i've fixed it in `develop` too

in general, the `classNameStrategies` come to solve special cases where you have some plugin specific logic that decides how the size, alignment and text-wrap classes should be applied. e.g. the image on mobile should always have `fullWidth` size class, and it should ignore the alignment if the size is `original` and image actual width greater than certain value.

in this case, there's no need in container class strategy, i guess it's copy/paste error.

PR #349, opened 2019-07-30
Add rtlcss ignore directives

Lukermelesh:
Yoshi automatically runs all CSS through rtlcss and adjusts it for RTL layout. Since you guys already did a great job of supporting RTL in the RCE, we do not need to modify this CSS, so I added ignore directives.


lxgreen:
@Lukermelesh don't we also need such fix in the Viewer CSS?

PR #351, opened 2019-08-01
Fix video audio plays on tab switch

mharsat:


packages/plugin-video/src/reactPlayerWrapper.jsx
```diff
@@ -26,16 +26,31 @@ export default class ReactPlayerWrapper extends Component {
...
+  }
+
   render() {
     if (!this.state.vimeoLoaded && this.isVimeoAndRequireJS()) {
       return null;
     }
-    return <ReactPlayer {...this.props} />;
+    return (
+      <ReactPlayer
+        {...this.props}
```
bchmn: it's best practice to place the rest of the props last

packages/viewer/src/RichContentViewer.jsx
```diff
@@ -59,6 +59,8 @@ export default class RichContentViewer extends Component {
       [styles.rtl]: textDirection === 'rtl',
     });
 
+    this.contextualData = { ...this.contextualData, disabled };
```
bchmn: currently the entire `contextualData` object is created in the `initContent` method of this class, this way it is clear to other developers what data the `contextualData` object contains.  
you should add the `disabled` field to the `contextualData` in `initContext` as well in order to retain the behavior

PR #368, opened 2019-08-25
e2e/test rtl support

mharsat:


e2e/cypress/support/commands.js
```diff
@@ -6,8 +6,17 @@ import { INLINE_TOOLBAR_BUTTONS } from '../dataHooks';
 const resizeForDesktop = () => cy.viewport('macbook-15');
 const resizeForMobile = () => cy.viewport('iphone-5');
 
+const buildQuery = params => {
```
shaulgo: very nice

e2e/test-env/src/server/renderer.jsx
```diff
@@ -18,8 +18,9 @@ export default function renderer() {
   return (req, res) => {
     const [componentId, fixtureName = 'empty'] = req.path.replace(/^\/|\/$/g, '').split('/');
     const isMobile = req.query.mobile === '';
+    const locale = req.query.he === '' ? 'he' : 'en';
```
shaulgo: what do you think of `req.query.hebrew`?

e2e/test-env/src/shared/components/Editor/Editor.jsx
```diff
@@ -78,7 +102,23 @@ class Editor extends Component {
...
+          locale={this.props.locale}
         />
+        <ReactModal
+          isOpen={this.state.showModal}
+          contentLabel="External Modal Example"
+          style={modalStyles}
+          role="dialog"
+          onRequestClose={onRequestClose || this.helpers.closeModal}
+        >
+          {this.state.showModal && (
```
shaulgo: Is this needed?

mharsat: copied as is from the example, removed it now from both

e2e/cypress/support/commands.js
```diff
@@ -61,7 +79,10 @@ Cypress.Commands.add('newLine', () => {
 });
 
 Cypress.Commands.add('blurEditor', () => {
-  getEditor().blur();
+  getEditor()
+    .blur()
+    .get('[data-hook=inlineToolbar]')
```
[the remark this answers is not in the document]
mharsat: this is the recommended cypress pattern for making sure an operation completed before proceeding. particularly when the editor is blurred, the inline toolbar may not disappear instantly leading to a flakey behavior, and so it waits for the operation to finish

PR #378, opened 2019-09-03
[editor] ui (default-toolbar-settings): Plus icon is displayed on the…

talevy17:
… text line too


packages/editor/web/src/RichContentEditor/Toolbars/default-toolbar-settings.js
```diff
@@ -12,7 +12,7 @@ const defaultSideToolbarVisibilityFn = editorState => {
   const selection = editorState.getSelection();
   const currentContent = editorState.getCurrentContent();
   const currentBlock = currentContent.getBlockForKey(selection.getStartKey());
-  return currentBlock.getLength() === 0;
+  return currentBlock.getType() !== 'atomic' && currentBlock.getType() !== 'code-block';
```
bchmn: it's wasteful to call `currentBlock.getType()` twice ,store it in a variable and check the variable instead of invoking it twice

PR #382, opened 2019-09-08
+ 'React based' to Readme top

tomByrer:
I incorrectly assumed that you guys were using jQuery ;)

Though you mention importing React later in the docs, folks might not get that far when scanning.   Plus, viewing a project on mobile, the first page doesn't display the full READMEl you click through [View README.md] to see more than the first few lines..  So a "built with React' would be handy in the mobile's 'above the fold', it seems where I placed it is the most terse.


README.md
```diff
@@ -2,7 +2,7 @@
 
 [![lerna](https://img.shields.io/badge/maintained%20with-lerna-cc00ff.svg)](https://lernajs.io/) [![Build Status](https://travis-ci.org/wix-incubator/rich-content.svg?branch=master)](https://travis-ci.org/wix-incubator/rich-content)
 
-## A super charged rich text editor with an extensible plugin system
+## Ract based, super charged rich text editor with an extensible plugin system
```
shaulgo:
```suggestion
## React based, super charged rich text editor with an extensible plugin system
```

tomByrer: derp...
That's what I get for PRs with little sleep.
cheers

PR #402, opened 2019-09-27
Mentions improvement - add ability to limit height of suggestion box

vybu:


lxgreen:
@vybu thanks for the PR, just one note -- plese update the Mentions API doc: https://github.com/wix-incubator/rich-content/blob/master/docs/PluginCustomization.md#mentions-plugin-settings

Issue #399, opened 2019-09-24
Bug: Small divider is not aligning to center


tomas-ki:
Small divider with center align is aligned to the left side of the page. Divider container is not full width, so aligning it to center aligns the divider only inside it's container, not the whole page. 
Reproducible in example project.

lxgreen:
<img width="833" alt="image" src="https://user-images.githubusercontent.com/4170342/65858274-36fd7a00-e36e-11e9-9aea-1794911eb028.png">

tomas-ki:
4.0.18:
![image](https://user-images.githubusercontent.com/22006186/65858646-2bf71980-e36f-11e9-85d2-024dcd40391c.png)

PR #406, opened 2019-10-03
small-sized divider container is centered when center alignment is applied

barboaron:
…entered when center alignment is applied


bchmn:
fixes #399

PR #420, opened 2019-10-17
fix editor focus

maimonav:


packages/editor/web/src/RichContentEditor/RichContentEditor.jsx
```diff
@@ -225,7 +225,7 @@ class RichContentEditor extends Component {
     },
   });
 
-  focus = () => this.editor.focus();
+  focus = () => setTimeout(this.editor.focus());
```
lxgreen: please see the `setTimeout` docs https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/setTimeout

currently, this code does not solve the bug 😉

PR #433, opened 2019-10-24
Side toolbar and add button are positioned to the right in RTL

mharsat:


packages/editor/web/statics/styles/side-toolbar.scss
```diff
@@ -35,4 +35,5 @@
   padding: 9px 19px;
   box-sizing: border-box;
   white-space: nowrap;
+  width: fit-content;
```
lxgreen: not supported in IE + Edge

mharsat: changed to a fixed value

PR #434, opened 2019-10-24
text alignment dropdown tooltips 

barboaron:


packages/common/web/src/Components/ToolbarButton.jsx
```diff
@@ -12,13 +12,26 @@ class ToolbarButton extends Component {
...
 
   render() {
-    const { theme, showTooltip, tooltipText, button, tooltipOffset } = this.props;
+    const {
+      theme,
+      showTooltip,
+      tooltipText,
+      button,
+      tooltipOffset,
+      shouldRebuildOnUpdate,
```
lxgreen: The name `shouldRebuildOnUpdate` is not clear in the context of the toolbar button and dropdown. It should say something like `shouldRefreshTooltips`

Issue #425, opened 2019-10-22
`normalizeInitialState` throw exception when content entityMap empty


EdgBuc:
In the database on some posts we have such content saved:
```
content:{
      blocks: [
        {
          key: 'f5fca',
          text: ' ',
          type: 'atomic',
          depth: 0,
          inlineStyleRanges: [],
          entityRanges: [
            {
              offset: 0,
              length: 1,
              key: 0,
            },
          ],
          data: {},
        },
      ],
      entityMap: {},
    }
```
in our code we need to compare content so before comparison we are using
`normalizeInitialState` which throws `TypeError: Cannot read property 'type' of undefined`
on line ` var entityType = entityMap[range.key + ''].type;`

Expected to not get an exception in this case.

lxgreen:
thanks for the report, we'll fix this one in the next release

lxgreen:
https://github.com/wix-incubator/rich-content/pull/436

bchmn:
@EdgBuc do you know how this contentstate was created?  it's obviously invalid and we'd like to isolate the issue and prevent it from happening again.

PR #436, opened 2019-10-27
Fix/data normalization protection

lxgreen:
### :bug: Bug Fix
- `common`
  - data-normalization: missing entity protection added


bchmn:
fixes #425 425

PR #438, opened 2019-10-27
align divider container correctly

barboaron:


packages/common/web/statics/styles/general.rtlignore.scss
```diff
@@ -72,3 +72,7 @@
     margin-right: auto;
   }
 }
+
+.alignCenterDivider{
```
bchmn: this is specific to the divider - it should not be in the common package

packages/plugin-divider/web/src/classNameStrategies.js
```diff
@@ -0,0 +1,11 @@
+import classNames from 'classnames';
+import { upperFirst } from 'lodash';
+
+export const customClassName = (componentData, theme, styles) => {
+  const { alignment } = componentData.config || {};
+  if (!alignment) {
+    return '';
+  }
+  const key = `align${upperFirst(alignment)}Divider`;
```
bchmn: what happens when it is aligned to the left or right?
there are no `.alignLeftDivider` or `.alignRightDivider` classes declared

PR #444, opened 2019-10-31
Adds file upload support

sarunas:


packages/plugin-video/web/statics/styles/video-selection-input-modal.scss
```diff
@@ -169,3 +169,23 @@
...
+  z-index: -1;
+}
+
+.fileInputLabel {
+  display: inline-block;
+  cursor: pointer;
+}
+
+.fileInput:focus + .fileInputLabel,
+.fileInputLabel:focus {
```
bchmn: why are all these additional styles needed?

PR #445, opened 2019-11-03
[editor] feat(initial intent): works for all plugins except file upload plugins

maimonav:


packages/common/web/src/Base/createBaseInsertPluginButton.jsx
```diff
@@ -18,8 +19,21 @@ export default ({ blockType, button, helpers, pubsub, settings, t, isMobile }) =
       this.state = {};
       const { buttonStyles } = props.theme || {};
       this.styles = mergeStyles({ styles, theme: buttonStyles });
+      this.myRef = React.createRef();
```
bchmn:
```suggestion
      this.buttonRef = React.createRef();
```

packages/editor/web/src/RichContentEditor/RichContentEditor.jsx
```diff
@@ -47,6 +47,15 @@ class RichContentEditor extends Component {
     this.initContext();
     this.initPlugins();
   }
+  componentDidMount() {
+    this.resetInitialIntent();
+  }
+
+  resetInitialIntent = () => {
```
bchmn: why is this needed?

maimonav: It is needed because initial intent as i understand, needs to happen only one time.

PR #453, opened 2019-11-07
Feat/drag n drop

mharsat:


bchmn:
please add support for disabling d&d through `disableDragAndDrop: true` config - default is false AKA d&d is enabled

mharsat:
> please add support for disabling d&d using `disableDragAndDrop: true` from config.
> default is false AKA d&d is enabled

drag & drop is now enabled by passing `draft-js-drag-n-drop-plugin` to the editor. since it's now opt-in,  the variable created in context indicating the plugin is in use was named  `enableDragAndDrop`.

PR #466, opened 2019-11-13
Update package.json

alonbt:
For some reason when installing the project the draft-js version is 10.5 - we suspect this is because the draft-js was in the devDependency and not the dependency section - so we're moving it to the correct path. version 10.5 does not work with this library and 11.0+ is required


bchmn:
draft-js is a `devDependency` since we need it installed for testing and a `peerDependency` in order to let consumers know that they need to install it.
closing

PR #486, opened 2019-11-24
the suggestions can be navigated by the arrows on the keyboard

barboaron:


shaulgo:
looks good, just resolve the conflict

lxgreen:
how difficult is it to add the scrolling to the keyboard navigation? currently the selected item remains overflow when navigating down

bchmn:
it's very difficult since the div that needs to be scrolled belongs to the underlying mentions plugin

PR #497, opened 2019-12-02
Content state utils

talevy17:


packages/common/web/statics/schemas/content-state.schema.json
```diff
@@ -144,7 +144,7 @@
             "title": "text",
             "default": "",
             "examples": ["test"],
-            "pattern": "^(.*)$"
+            "pattern": "^([^]*)$"
```
lxgreen: WTF?

packages/common/web/src/Utils/data-schema-validator.js
```diff
@@ -1,11 +1,11 @@
-import contentStateSchema from '../../statics/content-state.schema.json';
+import contentStateSchema from '../../statics/schemas/content-state.schema.json';
 
-export const validate = (data, schema) => {
+export const validate = (data, schema, alert = true) => {
```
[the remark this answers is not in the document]
talevy17: I thought that returning true or false was the desired behaviour, I'll delete it.

packages/common/web/statics/schemas/content-state.schema.json
talevy17: The previous pattern didn't accept line breaks in code block for some reason (e.g \n)
I tried a few, this was the only one that passed my tests

lxgreen: still this is weird way to accept any symbol. the `[^]` means "everything except ...nothing". please use `^(.|\n)*$` instead

talevy17: Done.

PR #507, opened 2019-12-05
Support all media urls in video and SoundCloud

mharsat:


packages/plugin-video/web/src/toolbar/videoSelectionInputModal.jsx
```diff
@@ -131,7 +132,7 @@ export default class VideoSelectionInputModal extends Component {
           <label
             htmlFor={this.id}
             className={styles.fileInputLabel}
-            role="button"
+            role="button" // eslint-disable-line jsx-a11y/no-noninteractive-element-to-interactive-role
```
shaulgo: @lxgreen do we need this role?

PR #514, opened 2019-12-08
[bi-events] feat: infrastructure for events of plugins usage

Barackos:


examples/main/shared/editor/Editor.jsx
```diff
@@ -160,25 +162,28 @@ export default class Editor extends PureComponent {
...
-          textToolbarType={textToolbarType}
-          theme={theme}
-          editorKey="random-editorKey-ssr"
-          anchorTarget={anchorTarget}
-          relValue={relValue}
-          locale={this.props.locale}
-          localeResource={this.props.localeResource}
+        <WrapWithCallbacks
+          {...debugBiLoggers()}>
+          <RichContentEditor
```
shaulgo: I think it is more straightforward to have the callbacks (onPluginAdd, onPl...) as props of the editor.

examples/main/config/WrapWithCallbacks.js
```diff
@@ -0,0 +1,42 @@
...
+			onPluginAdd: async (...args) => onPluginAdd(...args),
+			onPluginChange: async (...args) => onPluginChange(...args),
+			onPluginDelete: async (...args) => onPluginDelete(...args),
+		}
+		return Children.only(React.cloneElement(children, { helpers }));
+	}
+}
+
+WrapWithCallbacks.publish = async (post_id, editorState = {}, callBack = data => true) => {
+	if (Object.entries(editorState).length === 0)
```
shaulgo: this logic should go into draftUtils.js. 
Maybe getPostContentSummery(editorState) {...}

examples/main/config/WrapWithCallbacks.js
```diff
@@ -0,0 +1,42 @@
+import React, { Children } from 'react';
+import PropTypes from 'prop-types';
+import { getEntities } from '../../../packages/editor/web/src/lib/editorStateConversion';
+
+export default class WrapWithCallbacks extends React.Component {
+
+	render() {
+		const { onPluginAdd, onPluginChange, onPluginDelete, children } = this.props;
```
shaulgo: onPluginAdd, onPluginChange, onPluginDelete should be in the props of the editor. And then there is no need for this file

examples/main/config/WrapWithCallbacks.js
```diff
@@ -0,0 +1,42 @@
...
+		helpers.activityCallbacks = {
+			onPluginAdd: async (...args) => onPluginAdd(...args),
+			onPluginChange: async (...args) => onPluginChange(...args),
+			onPluginDelete: async (...args) => onPluginDelete(...args),
+		}
+		return Children.only(React.cloneElement(children, { helpers }));
+	}
+}
+
+WrapWithCallbacks.publish = async (post_id, editorState = {}, callBack = data => true) => {
```
shaulgo: onPublish should be in the wixWrapper. We should actually already ask our consumers about this tomorrow, even before the release.

packages/editor-common/web/src/Base/createBaseInsertPluginButton.jsx
```diff
@@ -15,12 +15,18 @@ export default ({ blockType, button, helpers, pubsub, settings, t, isMobile }) =
   class InsertPluginButton extends React.PureComponent {
     constructor(props) {
       super(props);
-      this.state = {};
+      this.state = { toolbarName: '' };
```
shaulgo: This need to be removed? 
you are getting it from the props

packages/editor-common/web/src/Base/createBaseInsertPluginButton.jsx
```diff
@@ -75,8 +82,17 @@ export default ({ blockType, button, helpers, pubsub, settings, t, isMobile }) =
       return { newBlock, newSelection, newEditorState };
     };
 
+    find(element) {
```
shaulgo: This need to be removed? 
you are getting it from the props

packages/editor-common/web/src/Base/createBaseInsertPluginButton.jsx
```diff
@@ -75,8 +82,17 @@ export default ({ blockType, button, helpers, pubsub, settings, t, isMobile }) =
...
+      const name = element.parentElement.parentElement.id;
+      const options = ['side_bar', 'footer'];
+      if (options.find(name) !== undefined) return name;
+    }
+
     onClick = event => {
       event.preventDefault();
+      const toolbarName = event.currentTarget.parentElement.parentElement.id;
```
shaulgo: This need to be removed? 
you are getting it from the props

packages/editor-common/web/src/Base/createBaseInsertPluginButton.jsx
```diff
@@ -15,12 +15,18 @@ export default ({ blockType, button, helpers, pubsub, settings, t, isMobile }) =
...
-      this.state = {};
+      this.state = { toolbarName: '' };
       const { buttonStyles } = props.theme || {};
+      const { toolbarName } = props;
       this.styles = mergeStyles({ styles, theme: buttonStyles });
       this.buttonRef = React.createRef();
+      this.onPluginAdd = this.getOnPluginAdd(helpers, toolbarName);
     }
 
+    getOnPluginAdd = ({ activityCallbacks: { onPluginAdd = () => false } }, toolbarName) => {
```
shaulgo: Maybe something like this instead 
onPluginAdd = pluginType => helpers?.onPluginAdd?.(pluginType, toolbarName);

packages/editor/web/src/RichContentEditor/RichContentEditor.jsx
```diff
@@ -231,7 +232,51 @@ class RichContentEditor extends Component {
     return element && element.querySelector('*[tabindex="0"]');
   }
 
+  calculateDiff = async (prevState, newState) => {
+    const blocks = state => state.getCurrentContent().getBlocksAsArray();
```
shaulgo: let's sit on this tomorrow. I want it to stay as immutablejs. And should also move to draftUtils

packages/editor/web/src/lib/editorStateConversion.js
```diff
@@ -13,4 +13,33 @@ const createEmpty = () => addVersion(EditorState.createEmpty(), version);
 const createWithContent = contentState =>
   addVersion(EditorState.createWithContent(contentState), contentState.VERSION);
 
-export { EditorState, createEmpty, createWithContent, convertToRaw, convertFromRaw };
+const getEntities = (editorState, entityType = null) => {
```
shaulgo: also will sit on it together

examples/main/config/WrapWithCallbacks.js
Barackos: I was thinking of making this specific wrapper as the open source wrapper, and move this file to a more appropriate place (and perhaps with a better name). What are your thoughts? (next phase? keep it?)

examples/main/config/WrapWithCallbacks.js
Barackos: It is there as well - wondered what would be the best practice. In my mind this is the open source wrapper, so it's here too (unless we remove it completely).

examples/main/config/WrapWithCallbacks.js
shaulgo: Will talk tomorrow, maybe you will convince me otherwise. But, from what I see, once `onPluginAdd, onPluginChange, onPluginDelete` move into the editor's props (instead of helpers). Then this wrapper doesn't do anything anymore.

packages/editor/web/src/RichContentEditor/Toolbars/SideToolbar/AddPluginFloatingToolbar.jsx
```diff
@@ -171,6 +171,7 @@ export default class AddPluginFloatingToolbar extends Component {
               setEditorState={setEditorState}
               theme={theme}
               hidePopup={this.hidePopup}
+              toolbarName={'Sidebar'}
```
lxgreen: we have TOOLBARS enum

packages/editor-common/web/src/Base/createBaseInsertPluginButton.jsx
```diff
@@ -56,6 +58,7 @@ export default ({
     };
 
     createBlock = (editorState, data, type) => {
+      this.onPluginAdd(blockType, this.state.toolbarName);
```
oded-soffrin: this.onPluginAdd function shouldn't get any variables according to constructor definition

packages/editor-common/web/src/Utils/draftUtils.js
```diff
@@ -300,6 +300,72 @@ function getSelection(editorState) {
...
+
+function quantitiesOf(entities) {
+  return entities.reduce((countArray, curr) => {
+    return {
+      ...countArray,
+      [curr.type]: !countArray[curr.type] ? 1 : countArray[curr.type] + 1,
+    };
+  }, {});
+}
+
```
oded-soffrin: Reduce function is still confusing for me so I had to stop and think what those few lines mean, so I also:
1. Made it more general #abstract

```
function countByType(arr) {
    return arr.reduce((countArray, curr) => {
      const {type} = curr;
      return {
        ...countArray,
        [type]: (countArray[type] || 0) + 1
      };
    }, {});
}
```js

2. Thought about where to move from draftUtil (it's not about draft) #scoutrule

3. Replace it w/ lodash #reuse

```
let  myArr = [{type: 'a'}, {type: 'b'}, {type: 'a'}];
_.countBy(myArr, x => x.type); // Returns: {a: 2, b: 1}
```

File: packages/editor/web/src/RichContentEditor/RichContentEditor.jsx [hunk 14]
```diff
@@ -435,6 +448,11 @@ class RichContentEditor extends Component {
   }
 }
 
+RichContentEditor.publish = async (postId, editorState = {}, callBack = () => true) => {
```
oded-soffrin: how is this function being called?

packages/editor/web/src/RichContentEditor/RichContentEditor.jsx
```diff
@@ -477,6 +495,9 @@ RichContentEditor.propTypes = {
   onAtomicBlockFocus: PropTypes.func,
   initialIntent: PropTypes.string,
   siteDomain: PropTypes.string,
+  onPluginAdd: PropTypes.func,
```
oded-soffrin: The propTypes of rich content editor is *super* important.
this single place of code (propTypes) actually defines the API between us and our cosumers
every addition / change in the API should be thought over carefully

in this case - we add 3 props where we can add 1 (they're all plugin changes "hooks")

packages/editor-common/web/src/Utils/draftUtils.js
```diff
@@ -300,6 +300,72 @@ function getSelection(editorState) {
...
+//ATM, looks for deleted plugins.
+//onChanges - for phase 2?
+//Added Plugins - checked elsewhere via toolbar clicks
+export const calculateDiff = async (prevState, newState, onPluginDelete) => {
+  const prevEntities = quantitiesOf(getEntities(prevState));
+  const currEntities = quantitiesOf(getEntities(newState));
+  const getBlockPlugins = blocks =>
+    blocks.filter(block => block.type !== 'unstyled' && block.type !== 'atomic');
```
oded-soffrin: duplication in getPostContentSummary, deserve its own func

packages/editor-common/web/src/Utils/draftUtils.js
```diff
@@ -300,6 +300,72 @@ function getSelection(editorState) {
...
+  const currBlockPlugins = quantitiesOf(getBlockPlugins(currBlocks));
+
+  const prevPluginsTotal = Object.assign(prevEntities, prevBlockPlugins);
+  const currPluginsTotal = Object.assign(currEntities, currBlockPlugins);
+
+  Object.keys(prevPluginsTotal).forEach(type => {
+    if (!currPluginsTotal[type] || prevPluginsTotal[type] > currPluginsTotal[type]) {
+      onPluginDelete(type);
+    }
+  });
```
oded-soffrin: aren't we afraid of performance degradation if it's called for each editor state change?

packages/editor/web/src/RichContentEditor/RichContentEditor.jsx
Barackos: done. called it "pluginHooks"

PR #515, opened 2019-12-09
Dec9 full screen tests

noamtwix:

### DIFF
--- a/e2e/cypress/support/index.js
+++ b/e2e/cypress/support/index.js
@@ -1,3 +1,9 @@
 import './commands';
 
 import '@applitools/eyes-cypress/commands';
+
+Cypress.on('uncaught:exception', (err, runnable) => {
+  // returning false here prevents Cypress from
+  // failing the test
+  return false;
+});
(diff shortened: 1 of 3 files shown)


e2e/cypress/support/index.js
```diff
@@ -1,3 +1,9 @@
 import './commands';
 
 import '@applitools/eyes-cypress/commands';
+
+Cypress.on('uncaught:exception', (err, runnable) => {
```
shaulgo: what is this for?

e2e/tests/plugins.e2e.js
```diff
@@ -76,6 +76,42 @@ describe('plugins', () => {
...
   });
 
+  context('full screen', () => {
+    before(function() {
+      eyesOpen(this);
+    });
+
+    after(() => cy.eyesClose());
+
+    context('image full screen', () => {
```
shaulgo: let's remove this context and the next one. Put both tests in the 'full screen' context.

PR #521, opened 2019-12-10
remove p element from blockRenderMap in getBlockRenderMap.js

noamtwix:


noamtwix:
After checking, I think that we can delete all the 'getBlockRenderMap.js' file and just use the DefaultDraftBlockRenderMap from draft.js.
The only thing that we need to check (and I'm waiting to answers from the slack channel), is if someone of the consumers passing theme for ol/ul.
Because right now this is the only thing that will be miss if we delete the file..

please let me know what do you think.

lxgreen:
> After checking, I think that we can delete all the 'getBlockRenderMap.js' file and just use the DefaultDraftBlockRenderMap from draft.js.
> The only thing that we need to check (and I'm waiting to answers from the slack channel), is if someone of the consumers passing theme for ol/ul.
> Because right now this is the only thing that will be miss if we delete the file..
> 
> please let me know what do you think.

i've seen you asked consumers to check this on `#rich-content-dev`, good idea.

PR #522, opened 2019-12-10
[editor] fix(draft.scss): support legacy margin scss attributes

maimonav:


packages/editor/web/statics/styles/draft.rtlignore.scss
```diff
@@ -75,6 +75,9 @@
       /*!rtl:begin:ignore*/
       direction: rtl;
       text-align: right;
+      @include phone {
+        margin-right: 10px;
```
shaulgo: this makes rtl have 10px margin. It's different from ltr. Why do we need this?

packages/editor/web/statics/styles/draft.rtlignore.scss
```diff
@@ -87,16 +90,17 @@
...
 
     .public-DraftStyleDefault-ol,
     .public-DraftStyleDefault-ul {
       margin: 0;
       padding: 0;
-
+      direction: ltr;
       & > * {
-        margin-inline-start: 40px;
+        margin-left: 40px;
```
shaulgo: What happens in rtl?

packages/editor/web/statics/styles/draft.rtlignore.scss
maimonav: Because of the plus icon.
In mobile, the plus icon is always in the right side so when you start writing in rtl the text is on top of the plus icon.

shaulgo: Did you check how it looks in the blog?

packages/editor/web/statics/styles/draft.rtlignore.scss
```diff
@@ -87,16 +90,17 @@
...
+      margin-right: 40px;
       /*!rtl:end:ignore*/
     }
 
     .public-DraftStyleDefault-ol,
     .public-DraftStyleDefault-ul {
       margin: 0;
       padding: 0;
-
+      direction: ltr;
```
shaulgo: this sets the direction of all lists to ltr?

PR #538, opened 2019-12-18
Remove docker, use Github actions

mharsat:


packages/editor/web/test/rich-content-editor.spec.jsx
```diff
@@ -2,17 +2,22 @@ import React from 'react';
...
 const { shallow } = Enzyme;
 
 // the mock is needed to overcome the 'Context is undefined' issue
 const getRichContentEditor = () => {
-  jest.doMock('wix-rich-content-editor-common', () => ({
+  jest.doMock('wix-rich-content-common', () => ({
     ...Common,
   }));
 
+  jest.doMock('wix-rich-content-editor-common', () => ({
```
lxgreen: do we need this mock? the reason for mock that I'm aware of, is the React Context.

PR #539, opened 2019-12-18
[emojiIcon]: fix: fix emoji icon

danarish4:


packages/plugin-emoji/web/src/icons/EmojiPluginIcon.jsx
```diff
@@ -2,11 +2,20 @@
...
 
 const EmojiPluginIcon = props => (
-  <svg viewBox="0 0 15 15" {...props}>
-    <path
-      d="M7.5 15a7.5 7.5 0 1 1 0-15 7.5 7.5 0 0 1 0 15zm0-1a6.5 6.5 0 1 0 0-13 6.5 6.5 0 0 0 0 13zM5 7a1 1 0 1 1 0-2 1 1 0 0 1 0 2zm4.53 4.575c-.639.444-1.03.708-2.03.708-1 0-1.564-.194-2.175-.666-.611-.473-1-.867-1.25-1.617-.056-.167-.083-.536 0-.675a.677.677 0 0 1 .333-.292.603.603 0 0 1 .492.042c.139.083.203.194.258.333.167.5.459.903.875 1.209.417.305.967.458 1.467.458.5 0 .88-.153 1.325-.458A2.12 2.12 0 0 0 9.7 9.408a.677.677 0 0 1 .292-.333.527.527 0 0 1 .458-.042.506.506 0 0 1 .333.292.84.84 0 0 1 .042.458 3.25 3.25 0 0 1-1.295 1.792zM10 7a1 1 0 1 1 0-2 1 1 0 0 1 0 2z"
-      id="path-1"
-    />
+  <svg width="19px" height="19px" viewBox="0 0 19 19" {...props}>
+    <g
+      id="01-Foundation-/-Icon-Small-18-/-02-General-/-17-Smile-Copy-2"
```
lxgreen: unless you see that the `id` is used somewhere else in the SVG markup, you should delete it. if it is used, you should validate that it is unique across all the SVG icons. the best way to make it unique for sure is to prepend the icon name to each `id`. 
in this specific icon, no `id`s are required, so please remove them.

packages/plugin-emoji/web/src/icons/EmojiPluginIcon.jsx
```diff
@@ -3,19 +3,11 @@ import React from 'react';
...
-        id="Color"
-        fill="currentColor"
-      />
-    </g>
+    <path
+      d="M9.5,16.5 C5.63400675,16.5 2.5,13.3659932 2.5,9.5 C2.5,5.63400675 5.63400675,2.5 9.5,2.5 C13.3659932,2.5 16.5,5.63400675 16.5,9.5 C16.5,13.3659932 13.3659932,16.5 9.5,16.5 Z M9.5,15.5 C12.8137085,15.5 15.5,12.8137085 15.5,9.5 C15.5,6.1862915 12.8137085,3.5 9.5,3.5 C6.1862915,3.5 3.5,6.1862915 3.5,9.5 C3.5,12.8137085 6.1862915,15.5 9.5,15.5 Z M12.3848284,11.3260771 C12.0260871,12.5812211 10.8703397,13.5 9.5,13.5 C8.12595472,13.5 6.9676639,12.5762454 6.61227891,11.3158876 C7.09025574,11.1805267 6.61227891,11.3158876 7.57588228,11.0475635 C7.81400628,11.8859408 8.58529254,12.5 9.5,12.5 C10.4151019,12.5 11.1866589,11.8854111 11.4244255,11.0464788 C11.7613068,11.1460724 11.4244255,11.0475635 12.3848284,11.3260771 Z M7.5,7.5 C8.05228475,7.5 8.5,7.94771525 8.5,8.5 C8.5,9.05228475 8.05228475,9.5 7.5,9.5 C6.94771525,9.5 6.5,9.05228475 6.5,8.5 C6.5,7.94771525 6.94771525,7.5 7.5,7.5 Z M11.5,7.5 C12.0522847,7.5 12.5,7.94771525 12.5,8.5 C12.5,9.05228475 12.0522847,9.5 11.5,9.5 C10.9477153,9.5 10.5,9.05228475 10.5,8.5 C10.5,7.94771525 10.9477153,7.5 11.5,7.5 Z"
+      id="Color"
```
lxgreen: how about this id?

shaulgo: I bet you totally had enough with this pr. But this svg is way too big. Gmail broke trying to load the whole string ;)
Also, if we use svg files they get optimized even further. (we have rollup plugin that does this for us)
Finally, just changing the viewbox a little bit in the current svg will do the trick. Remind me next time you see me. We'll sit on it together.

PR #546, opened 2019-12-23
Link plugin - saves the last data and the initial state of the checkboxes("Open in a new tab", "Add a nofollow tag") is according to the defaults

barboaron:


packages/editor/web/src/RichContentEditor/Toolbars/MobileBlockLinkModal.jsx
```diff
@@ -24,7 +26,7 @@ export default class MobileBlockLinkModal extends Component {
     const { pubsub, theme, isMobile, anchorTarget, relValue, t, uiSettings } = this.props;
     const componentLink = pubsub.getBlockData({ key: 'componentLink' });
     const { url, target, rel } = componentLink || {};
-    const targetBlank = target === '_blank';
+    const targetBlank = target === undefined ? anchorTarget === '_blank' : target === '_blank';
```
barboaron:
```suggestion
    const targetBlank = target === '_blank' || anchorTarget === '_blank';
```

packages/plugin-button/web/src/constants.js
```diff
@@ -25,29 +25,31 @@ export const COLORS = {
...
-    size: 'small',
-    width: 'fit-content',
-  },
-  button: {
-    settings: {
-      buttonText: 'Click Me',
-      url: '',
-      rel: false,
-      target: false,
+export const defaultContentState = (rel, target) => {
```
lxgreen: function name should be a verb, e.g. `getDefaultContentState`. Also, the name is wrong -- this is just a component data, not the whole content state.

packages/plugin-image/web/src/toolbar/image-settings.js
```diff
@@ -89,10 +89,14 @@ class ImageSettings extends Component {
 
   saveLink = () => {
     const { linkPanelValues } = this.state;
-    if (linkPanelValues.url === '') {
+    const { anchorTarget, relValue } = this.props;
+    const { url, targetBlank, nofollow, isValid } = linkPanelValues;
+    const target = targetBlank ? '_blank' : anchorTarget !== '_blank' ? anchorTarget : '_self';
```
lxgreen: same refactor here

PR #549, opened 2019-12-24
fix: [htmlPlugin] edit option is open when opening html plugin

danarish4:


oded-soffrin:
what's the status of this PR?

PR #552, opened 2019-12-24
[editor] fix: Plugin toolbar shouldCreate config crush

noamtwix:


CHANGELOG.md
```diff
@@ -32,6 +32,8 @@
     - [#543](https://github.com/wix-incubator/rich-content/pull/543) fix undo-redo icons on mobile
   - `editor`
     - [#521](https://github.com/wix-incubator/rich-content/pull/521) replace p element with div element in lists
+  - `editor`
```
lxgreen: please do not create duplicate `editor` section. there's already existing one at the line 29, move your PR line there. also, your previous comment in line 34 should go there

packages/editor-common/web/src/Base/createBaseToolbar.jsx
```diff
@@ -146,7 +146,7 @@ export default function createToolbar({
     };
 
     onVisibilityChanged = focusedBlock => {
-      if (focusedBlock) {
+      if (this.shouldCreate && focusedBlock) {
```
lxgreen: this will cause `hideToolbar` to be called if `shouldCreate === false`  - this is conceptually wrong to hide a toolbar if there's no toolbar at all. 
so , please change to `if (!shouldCreate) return;` at the first place

PR #557, opened 2019-12-29
Update messages_en.json

michael-wix:


shaulgo:
My guess is that most people don't care about this. And the ones that do might be looking for 'nofollow'. What do you think?

PR #561, opened 2019-12-30
Dec30 cleanup read only prop

noamtwix:


shaulgo:
after fixing, make it pass the tests  and merge

PR #563, opened 2020-01-01
In-Plugin Editing - Image Caption

mharsat:


packages/plugin-image/web/src/InPluginInput.jsx
```diff
@@ -0,0 +1,54 @@
...
+  );
+
+  render() {
+    const { onChange, value } = this.props;
+    return (
+      <input
+        className={this.className}
+        value={value}
+        readOnly={!onChange}
+        onChange={onChange}
```
shaulgo: extract the value from the event here

packages/plugin-image/web/statics/styles/in-plugin-input.scss
```diff
@@ -0,0 +1,14 @@
+.inPluginInput {
+    border: none;
```
shaulgo: add dir: auto

packages/plugin-image/web/statics/styles/in-plugin-input.scss
```diff
@@ -0,0 +1,14 @@
...
+    line-height: 21px;
+    width: 100%;
+    position: inherit;
+    z-index: 1;
+    font: inherit;
+    font-size: 14px;
+
+    &Viewer {
+        @extend .inPluginInput;
```
shaulgo: do you need @extend?

packages/plugin-image/web/src/InPluginInput.jsx
mharsat: shouldn't the value be taken from the `componentData`?

packages/plugin-image/web/statics/styles/in-plugin-input.scss
mharsat: I want the input area to have the normal clicking cursor in viewer so it won't look editable. I need the extend in order to have both `inPluginInput` and `inPluginInputViewer` classes without repeating code. An alternative is to extract `cursor: pointer` into a different class

packages/plugin-image/web/statics/styles/in-plugin-input.scss
mharsat: nice!

PR #574, opened 2020-01-06
[*] improve custom icon API

barboaron:


oded-soffrin:
I see toolbar.icons.undo/redo/replace weren't change - is it on purpose?
also - is that a breaking change?

barboaron:
The toolbar.icons.undo/redo weren't change because it's in the same plugin, so we need to use different names.
As far as I know, toolbar.icons.replace is not an insert button, so I didn't change it :)

PR #576, opened 2020-01-07
[colorPicker] new colors are added to the end of the list + 'reset to default' removes from the 'inlineStyleRanges' in the content state

barboaron:

### DIFF
--- a/packages/editor-common/web/src/Components/ColorPicker/ColorPicker.jsx
+++ b/packages/editor-common/web/src/Components/ColorPicker/ColorPicker.jsx
@@ -60,10 +60,7 @@ class ColorPicker extends PureComponent {
   }
 
   resetColor = () => {
-    const { defaultColor } = this.props;
-    const shouldRemoveInlineStyle = true;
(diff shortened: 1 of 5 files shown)


packages/editor-common/web/src/Components/ColorPicker/ColorPicker.jsx
```diff
@@ -62,7 +62,7 @@ class ColorPicker extends PureComponent {
   resetColor = () => {
     const { defaultColor } = this.props;
     this.setState({ color: defaultColor });
-    this.props.onChange(defaultColor);
+    this.props.onChange(defaultColor, true);
```
shaulgo: instead of using `onChange`. Add `props.onResetColor`

barboaron: I passed the “setColor” function twice- one by “props.onChange” and the second by “props.onResetColor”, I think it’s better than duplicating code. If you think differently,I will change it.

PR #582, opened 2020-01-09
image-viewer: rerender on image load

lxgreen:
rerender on image load


packages/plugin-image/web/src/image-viewer.jsx
```diff
@@ -88,6 +88,7 @@ class ImageViewer extends React.Component {
   onHighResLoad = e => {
     e.target.style.opacity = 1;
     this.preloadImage && (this.preloadImage.style.opacity = 0);
+    this.forceUpdate();
```
shaulgo: why is this needed?

PR #600, opened 2020-01-14
[anchor] feat: new anchor plugin

noamtwix:


oded-soffrin:
@noamtwix we can probably close the PR right? (keep the branch for reference)

PR #609, opened 2020-01-19
fix (lerna): storybook auto-versioning

Barackos:


examples/storybook/package.json
```diff
@@ -21,26 +21,29 @@
     "release": "yoshi release",
     "storybook": "start-storybook -p 6006",
     "clean": "rm -rf dist storybook-static",
-    "build": "npx yoshi build && npx build-storybook"
+    "build": "yarn build-storybook"
```
shaulgo: we have npm across all the configes. Is there a reason to use yarn here?

scripts/deployExamples.js
```diff
@@ -12,7 +12,7 @@ const EXAMPLES_TO_DEPLOY = [
   {
     name: 'rich-content-storybook',
     path: 'examples/storybook',
-    buildCmd: 'npm i && yarn build-storybook',
+    buildCmd: 'yarn build-storybook -s public',
```
shaulgo: Same as previous comment. Why yarn instead of npm?

PR #612, opened 2020-01-19
[plugin-map] fix: warnings issues of google-maps-loader props

maimonav:


packages/plugin-map/web/src/MapViewer.jsx
```diff
@@ -129,7 +129,7 @@ MapViewer.propTypes = {
   settings: PropTypes.shape({
     width: PropTypes.number,
     height: PropTypes.number,
-    googleMapApiKey: PropTypes.string.isRequired,
+    googleMapApiKey: PropTypes.string,
```
lxgreen: should be required as the plugin is useless without it.

packages/plugin-map/web/src/MapViewer.jsx
```diff
@@ -84,7 +84,7 @@ export class MapViewer extends Component {
         <div ref={this.setRootElementRef} style={style} data-hook="mapViewer">
           <ReactGoogleMapLoader
             params={{
-              key: googleMapApiKey,
+              key: googleMapApiKey || '',
```
lxgreen: could be defined in defaultProps

PR #619, opened 2020-01-20
[Link plugin] fix:when updating the link the content state data is changed to the correct fields

barboaron:


packages/editor-common/web/src/Utils/draftUtils.js
```diff
@@ -164,8 +165,9 @@ export const getAnchorBlockData = editorState => {
 
 export const setEntityData = (editorState, entityKey, data) => {
   if (entityKey) {
+    const newData = createEntityDataForLink(data);
```
shaulgo: this will break every change in entityData

shaulgo:
Also fix content state tests

packages/editor-common/web/src/Utils/draftUtils.js
barboaron: I fixed it:)

PR #620, opened 2020-01-22
[common] bugfix(normalizeInitialState): image normalizer - move…

oded-soffrin:
[common] bugfix(normalizeInitialState): image normalizer - move 'width' from compData root to config

(image sizes with resize created before version 6 were ignored, creating smaller images in version update)


oded-soffrin:
@lxgreen hotfix, please check and merge

PR #628, opened 2020-01-26
Fix weird space bug

oded-soffrin:
To check out the change effect, run storybook and see test case example (further instruction there)
Caused by - https://jira.wixpress.com/browse/NWB-2849


shaulgo:
unit tests are failing. Also, deployd example app doesn't work https://rich-content-fix-weird-spaces.surge.sh/#!

packages/viewer/web/src/utils/convertContentState.jsx
```diff
@@ -53,13 +53,16 @@ const getBlocks = (mergedStyles, textDirection, { config }) => {
...
         const { interactions } = blockProps.data[i];
-        const BlockWrapper = isArray(interactions)
+        const BlockWrapper = Array.isArray(interactions)
           ? getInteractionWrapper({ interactions, config, mergedStyles })
           : DefaultInteractionWrapper;
 
-        return (
-          <BlockWrapper key={`${blockProps.keys[i]}_wrap`}>
+        const inner =
+          Type === 'br' ? (
```
shaulgo: br doesn't get the class names. Line-height might break. Wouldn't it just work without the trinary?

PR #631, opened 2020-01-27
Html height

oded-soffrin:
This is a fix to an issue introduced 3 months ago, fixing one problem and creating another

this fix https://github.com/wix-incubator/rich-content/commit/8b7ac51f5118c90b116d1509f5609f0a4ac0cdba added width&height to html plugin in viewer - I think the goal was trying to fix left/right alignments

the problem is the height isn't updated in viewer (no pubsub) so i removed the height part, what do you think?

(most of the changes is storybook boiler plate, the only production change is on file `getPluginViewers.jsx`)
to see the fix run storybook and go to **TestCases** story
To see the problem copy the _content state_ and past in exampleApp Playground


packages/viewer/web/src/getPluginViewers.jsx
```diff
@@ -71,8 +71,10 @@ class PluginViewer extends PureComponent {
             containerProps.style = { width: componentData.config.width };
           }
           if (type === 'wix-draft-plugin-html') {
-            const { width: currentWidth, height: currentHeight } = componentData.config;
-            containerProps.style = { width: currentWidth, height: currentHeight };
+            // const { width: currentWidth, height: currentHeight } = componentData.config;
```
lxgreen:
```suggestion
```

packages/viewer/web/src/getPluginViewers.jsx
```diff
@@ -71,8 +71,10 @@ class PluginViewer extends PureComponent {
...
+            // containerProps.style = { width: currentWidth, height: currentHeight };
```
lxgreen:
```suggestion
```

PR #642, opened 2020-02-02
 block selection isn't removed when losing focus

barboaron:


packages/editor-common/web/src/Utils/draftUtils.js
```diff
@@ -1,5 +1,6 @@
 import { EditorState, Modifier, RichUtils, SelectionState, AtomicBlockUtils } from 'draft-js';
 import { cloneDeep, flatMap, findIndex, findLastIndex } from 'lodash';
+// import DraftOffsetKey from 'draft-js/lib/DraftOffsetKey';
```
shaulgo: delete comment

PR #647, opened 2020-02-04
block selection isn't removed when losing focus

shaulgo:


barboaron:
I opened a new PR : [editor] fix(*): fixCreateBlockChangeNativeSelection - fix 'block selection isn't removed when losing focus'

PR #648, opened 2020-02-04
Fix/video loader

talevy17:


packages/plugin-video/web/src/video-viewer.jsx
```diff
@@ -32,6 +32,9 @@ class VideoViewer extends Component {
           url.then(url => this.setState({ url: this.normalizeUrl(url) }));
         }
       }
+      if (nextProps.componentData.tempData !== this.props.componentData.tempData) {
+        this.onReactPlayerReady();
```
shaulgo: actually I don't see why this is needed at all. onReactPlayerReady will be called twice. Once for the local video and once when the remote video is loaded. It should work fine without this?

packages/plugin-video/web/src/video-viewer.jsx
```diff
@@ -32,6 +32,9 @@ class VideoViewer extends Component {
           url.then(url => this.setState({ url: this.normalizeUrl(url) }));
         }
       }
+      if (nextProps.componentData.tempData !== this.props.componentData.tempData) {
```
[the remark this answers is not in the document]
talevy17: since it should happen only once the video is loaded then that should be enough, I will change it

packages/plugin-video/web/src/video-viewer.jsx
talevy17: The thing is that the video can be ready to play before it's completely uploaded (which caused the issue to be raised in the first place)
I will optimise the execution though as it's not entirely necessary
perhaps using two flags will solve it (one for video ready to play, and one for video finished uploading to the remote server)

PR #649, opened 2020-02-04
[e2e] test(plugins.e2e.js): add gallery-out-of-view fixutre & test

oded-soffrin:


e2e/tests/plugins.e2e.js
```diff
@@ -107,6 +107,12 @@ describe('plugins', () => {
       cy.eyesCheckWindow(this.test.title + ' settings');
     });
 
+    it('render gallery out of view', function() {
+      cy.loadEditorAndViewer('gallery-out-of-view');
```
shaulgo: should we test that it's not loading pictures in the gallery when it's out of view?
We could add another eyes before the scroll.

PR #651, opened 2020-02-05
[test-env] fix(examples/shared): gallery scrolling element

oded-soffrin:


e2e/tests/__snapshots__/plugins.e2e.js.snap
```diff
@@ -2647,26 +2647,24 @@ exports[`plugins > gallery > render gallery out of view #0`] =
...
           "enableScroll": true,
           "fixedColumns": 0,
-          "galleryLayout": 7,
+          "galleryLayout": 0,
+          "gallerySizePx": "300",
           "galleryThumbnailsAlignment": "none",
-          "galleryType": "Strips",
-          "groupSize": 1,
-          "groupTypes": "1",
+          "galleryType": "Columns",
```
shaulgo: I'm guessing you approve these changes?

examples/main/shared/editor/Editor.jsx
```diff
@@ -170,18 +191,10 @@ export default class Editor extends PureComponent {
           onChange={this.handleChange}
           helpers={this.helpers}
           plugins={Plugins.editorPlugins}
-          config={Plugins.config}
-          editorState={this.props.editorState}
-          initialState={this.props.initialState}
-          isMobile={this.props.isMobile}
-          textToolbarType={textToolbarType}
-          theme={theme}
+          config={Plugins.getConfig({ scrollingElement: scrollingElementFn })}
```
shaulgo: maybe add some way to know this is for the gallery? 
It could be galleryScrollingElment. Or GALLERY_TYPE: scrollingElement and then there's no need for a function. 
I wonder what you think? ;)

shaulgo: also, I think if it's a function it should stay with the `Fn`

oded-soffrin: agree with the GALLERY_TYPE
but can't change `scrollingElement` to `scrollingElementFn` (that will be a breaking change)

PR #654, opened 2020-02-06
Fix/context issues

lxgreen:
### :bug: Bug Fix
  - `viewer`
    - plugin settings retrieved correctly


packages/viewer/web/src/getPluginViewers.jsx
```diff
@@ -52,8 +52,8 @@ class PluginViewer extends PureComponent {
     } = this.props;
     const { component: Component, elementType } = pluginComponent;
     const { container } = pluginComponent.classNameStrategies || {};
-    const settings = (config && config[type]) || {};
     const { anchorTarget, relValue, config, theme } = context;
+    const settings = (config && config[type]) || {};
```
mharsat:
```suggestion
    const settings = config?.[type] || {};
```

PR #660, opened 2020-02-09
[editor] fix focus/unselection of plugins when they are added

shaulgo:


.github/workflows/actions.yml
```diff
@@ -3,7 +3,7 @@ on: push
 
 jobs:
   visual-tests:
-    runs-on: ubuntu-18.04
```
lxgreen: AFAIK, this caused the test run to fail. that was the reason for a fixed Ubuntu version.

packages/editor/web/package.json
```diff
@@ -37,7 +37,7 @@
     }
   },
   "dependencies": {
-    "draft-js-focus-plugin": "~3.0.1",
+    "draft-js-focus-plugin": "github:shaulgo/draft-js-plugins#focusPlugin",
```
lxgreen: wtf?

PR #665, opened 2020-02-10
wiring move up/down inline buttons

noamtwix:


oded-soffrin:
@noamtwix still relevant?

PR #666, opened 2020-02-10
Feat/es check

lxgreen:
### :house: Internal
  - `general`
    - ES5 syntax validation in `dist/statics` on postbuild


scripts/esCheck.js
```diff
@@ -0,0 +1,37 @@
...
+      process.stderr.write(e.stderr);
+      process.exit(1);
+    }
+  }
+}
+
+const ES_CHECK_DEFAULTS = {
+  ecmaVersion: 'es5',
+  module: false,
+  files: 'dist/statics/**/*.js',
```
oded-soffrin: why only in dist/statics and not in all of /dist/**/*.js ? no `const` should be there as well, right ?

lxgreen: the `dist/` contains both `esm` and `cjs` files. The `esm` files contain the `es6` at the top level, and that's fine. The `cjs` do not contain the `es6`, as these are transpiled to `es5`. The only place that is not processed is `dist/statics` -- the items are just copied from statics as they are.

lxgreen: clarification: although the `esm` does not contain `const`, it does contain `import` which also fails the es5 test

PR #683, opened 2020-02-13
[plugins] New Plugin Generator script

danarish4:


scripts/templates/atomic-plugins/web/src/icons/InsertPluginIcon.jsx
```diff
@@ -0,0 +1,32 @@
...
+                id="🔹Icon-Color"
```
oded-soffrin: looks like a funky character

danarish4: changed :)

PR #686, opened 2020-02-14
[gallery] fix(*): add disableHover config for blog backward support

oded-soffrin:
Didn't end up using versioning, because in the example they reported it already updated to version 6 so won't fix.
Used additional gallery config, that we can later on use to add the needed checkbox to display alt-text as title. for now it just resolve the urgency for blog and restore the behvaior so they continue with upgrading

### DIFF
--- a/packages/common/web/src/Utils/normalization/normalizeInitialState.js
+++ b/packages/common/web/src/Utils/normalization/normalizeInitialState.js
@@ -6,6 +6,7 @@ import {
   LINK_TYPE,
   VIDEO_TYPE_LEGACY,
   IMAGE_TYPE_LEGACY,
+  GALLERY_TYPE,
 } from '../../consts';
 import { linkDataNormalizer, imageDataNormalizer } from './dataNormalizers';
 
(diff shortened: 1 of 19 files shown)


packages/common/web/src/Utils/normalization/normalizeInitialState.js
```diff
@@ -20,49 +21,66 @@ const normalizeComponentData = (type, componentData, config) =>
...
-    delete componentData.alignment;
-    config.alignment = alignment;
-    config.size = 'small';
+const normalizeComponentConfig = (entityType, componentData, config) => {
+  if (
+    entityType === GALLERY_TYPE &&
+    config?.config?.[entityType]?.displayTitleDefault !== undefined
+  ) {
+    const { items } = componentData;
+    const displayTitleDefault = config?.config?.[entityType]?.displayTitleDefault;
```
lxgreen: suggest to call the new field `showTitle` like in image entity `data.config`. This allows to unify config schemas

packages/common/web/src/Utils/normalization/normalizeInitialState.js
```diff
@@ -20,49 +21,66 @@ const normalizeComponentData = (type, componentData, config) =>
...
+    return {
+      ...componentData,
+      items: items.map(item => ({ displayTitle: displayTitleDefault, ...item })),
```
lxgreen: do we want to toggle the title display at a single image level?

oded-soffrin: I add this because I assumed we'll add a checkbox to display/hide the title from exapnd mode

packages/common/web/src/Utils/normalization/normalizeInitialState.js
oded-soffrin: 🆗🆗🆗 will change

packages/common/web/src/Utils/normalization/normalizeInitialState.js
lxgreen: 👍

oded-soffrin:
Tests are passing (video & map are still flaky)

PR #691, opened 2020-02-16
Resize image border

ghost:


oded-soffrin:
@shaulgo do you remember what this was about?

shaulgo:
> @shaulgo do you remember what this was about?

I think it makes the resize handles in all corners. Instead of the two we have in the middle of the left and right sides. 
But there's a product issue here. Because we compute the height according to the width. 
It is possible to solve with crop or some other solution. But I don't believe we will do this anytime soon. And when/if we will then it will probably be easier to do from scratch. 
I'll close it.

PR #698, opened 2020-02-17
New tests for v6 8 0

mharsat:


oded-soffrin:
Matan: 
I added a test for custom video upload and different gallery layouts. There is some issue where some gallery layouts require some scrolling before they render and some don’t, and the new baselines capture this pretty well. Maybe we should check this. Anyway, the PR is available for merge 

Specifically the mobile baselines should be checked to see the galleries are rendered as expected, I was not actually sure how they should look like in mobile

PR #707, opened 2020-02-19
[hashtag] fix allow hashtag and link in the same block

Sapir-Shamay:


packages/common/web/src/Utils/draftUtils.js
```diff
@@ -1,6 +1,16 @@
 export const hasLinksInBlock = (block, contentState) => {
-  return block.entityRanges.some(entityRange => {
+  return !!getLinkRangesInBlock(block, contentState);
```
Sapir-Shamay: always true

packages/plugin-hashtag/web/src/HashtagDecorator.jsx
```diff
@@ -2,24 +2,23 @@ import { range } from 'lodash';
...
 import hashtagRegexes from './hashtagRegexes';
 
-export default (hasLinksInBlock, immutableList) =>
+export default (getLinkRangesInBlock, immutableList) =>
   class HashtagDecorator {
     constructor(componentProps) {
       this.componentProps = componentProps;
     }
 
+    checkOverlap = (range, start, end) => {
```
Sapir-Shamay:
```suggestion
    isOverlapping = (range, start, end) => {
```

PR #752, opened 2020-03-09
Move draft-js  to be dependency (instead of peerDep) of editor-common only

oded-soffrin:


oded-soffrin:
@shaulgo @lxgreen all tests, exampleApp and storybook works
would love to bounce ideas how to check if it effects bundle size of plugins / editor before merging as breaking change

lxgreen:
i remember an issue -- the consumer SSR applications required draft-js dependency in order to build the server. i wander is it still relevant

packages/editor-common/web/src/index.js
```diff
@@ -96,3 +96,21 @@ export {
   PLUGIN_DECORATION_PROPS,
   PLUGIN_DECORATIONS,
 } from './consts';
+
+export {
+  convertToRaw,
```
shaulgo: I can see pro and cons for "wrapping" draft-js like so. I would like to better understand your perspective.

rollup.externals.js
```diff
@@ -12,45 +10,19 @@ export const externals = [
...
 
-export const isExternal = id =>
-  !id.startsWith('\0') &&
-  !id.startsWith('.') &&
-  !id.startsWith('/') &&
-  !excludedExternals.find(regex => (typeof regex === 'string' ? regex === id : regex.test(id))) &&
-  !!externals.find(externalName => new RegExp(externalName).test(id));
+export const isExternal = id => {
+  return (
+    !localPrefixes.find(prefix => id.startsWith(prefix)) &&
```
shaulgo: I'm not sure how the externals work. Let's talk about it. I want to know :)

shaulgo: fun fact - `some` is like `find` but returns boolean

packages/editor-common/web/src/index.js
oded-soffrin: Currently consumer use draft functionality, and I want them to get it via RCE, so we can change to @wix/draft-js seamlessly

rollup.externals.js
oded-soffrin: rollup external explains. 
https://rollupjs.org/guide/en/#peer-dependencies

(changed find -> some)

packages/editor-common/web/src/index.js
shaulgo: Our consumers are using these functions? I wonder why.
What is the advantage to using it like this inside our code? instead of importing from draft-js.

lxgreen: what cons do you mind?

oded-soffrin: Not following, let's discuss f2f tomorrow ?

PR #754, opened 2020-03-09
add Normalize to convertFromRaw

ShamaySapir:


ShamaySapir:
@shaulgo can be closed?

PR #756, opened 2020-03-10
fix: when canceling link panel the selection jump

ShamaySapir:

### DIFF
--- a/packages/plugin-link/web/src/toolbar/TextLinkPanel.jsx
+++ b/packages/plugin-link/web/src/toolbar/TextLinkPanel.jsx
@@ -27,8 +27,8 @@ export default class TextLinkPanel extends Component {
       t,
       isActive: !isEmpty(linkData),
       onDone: this.createLinkEntity,
-      onCancel: this.cancelLink,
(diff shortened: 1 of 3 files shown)


packages/plugin-link/web/src/toolbar/TextLinkPanel.jsx
```diff
@@ -26,7 +27,8 @@ export default class TextLinkPanel extends Component {
       t,
       isActive: !isEmpty(linkData),
       onDone: this.createLinkEntity,
-      onCancel: this.hideLinkPanel,
+      onCancel: this.cancelLink,
+      onHide: this.hideLinkPanel,
```
danarish4: hide panel

packages/plugin-link/web/src/toolbar/TextLinkPanel.jsx
```diff
@@ -62,6 +64,18 @@ export default class TextLinkPanel extends Component {
     setEditorState(newEditorState);
   };
 
+  cancelLink = () => {
```
danarish4: change func name to on cancel

PR #772, opened 2020-03-15
fix(wrapper): onChange not called

Barackos:


rollup.config.common.js
```diff
@@ -14,7 +14,7 @@ export default (output, shouldExtractCss) => {
   const plugins = createPlugins(shouldExtractCss);
   output = output.map(o => ({ ...o, sourcemap: true }));
   if (process.env.MODULE_WATCH) {
```
oded-soffrin: Suggestion:  if (process.env.MODULE_WATCH && ! process.env.BUILD_CJS) {
}

Barackos: [https://github.com/wix-incubator/rich-content/pull/773](https://github.com/wix-incubator/rich-content/pull/773)

PR #778, opened 2020-03-15
[plugins] fix fullwidth

ShamaySapir:


shaulgo:
content-state-tests are not passing

PR #782, opened 2020-03-15
social polls plugin

mbutsykin:


CHANGELOG.md
```diff
@@ -13,6 +13,10 @@
     Click to see more.
   </summary>
 
+### :rocket: New Feature
+- `poll-plugin`
+  - base ui version of poll plugin
```
oded-soffrin: we add the PR like this:
- [#782](https://github.com/wix-incubator/rich-content/pull/782) base ui version of poll plugin

oded-soffrin:
@shaulgo groups would like to try and consume it to check their side, any way we can release an additional alpha (draft fork is 'catching' the current alpha)

PR #805, opened 2020-03-19
Fix/text color mobile

talevy17:


examples/main/shared/editor/Editor.jsx
```diff
@@ -55,10 +55,14 @@ export default class Editor extends PureComponent {
     };
     this.helpers = {
       //these are for testing purposes only
-      onPluginAdd: async (plugin_id, entry_point, version) => console.log('biPluginAdd', plugin_id, entry_point, version),
-      onPluginDelete: async (plugin_id, version) => console.log('biPluginDelete', plugin_id, version),
-      onPluginChange: async (plugin_id, changeObj, version) => console.log('biPluginChange', plugin_id, changeObj, version),
-      onPublish: async (postid, callback, version) => console.log('biOnPublish', ({ data }) => data, version),
+      onPluginAdd: async (plugin_id, entry_point, version) =>
```
shaulgo: is this only whitespace changes?

packages/plugin-text-color/web/src/toolbar/BaseTextColor.jsx
```diff
@@ -38,9 +38,8 @@ export default class BaseTextColor extends Component {
   };
 
   closePanel = editorState => {
-    this.setState({ isPanelOpen: false });
+    this.setState({ isPanelOpen: false }, () => this.preserveSelectionState(editorState));
```
shaulgo: do you know what this does?

examples/main/shared/editor/Editor.jsx
talevy17: Yes

packages/plugin-text-color/web/src/toolbar/BaseTextColor.jsx
talevy17: adds the next function call (that sets the editor state) as a callback to the state change that closes the modal
I'm not sure if the forceselection is needed though it's probably there for a reason

PR #808, opened 2020-03-20
[editor] remove custom list support (getBlockRenderMap)

oded-soffrin:


oded-soffrin:
breaks theme unfortunately

PR #823, opened 2020-03-23
move to @wix/draft-js

oded-soffrin:


rollup.plugins.js
```diff
@@ -1,16 +1,26 @@
...
   });
 };
 
+const resolveAlias = () => {
+  const alias = require('@rollup/plugin-alias');
+  return alias({
+    entries: {
+      'draft-js': '@wix/draft-js',
+    },
+    resolve,
```
bchmn: something seems off here, I might be missing something but according to the [plugin docs](https://github.com/rollup/plugins/tree/master/packages/alias#custom-resolvers):
- there is no need to pass the `resolve` plugin to the `alias` plugin, it just needs to be in the array of plugins passed to rollup.
- if you would like to pass a `customResolver` you need to use `customResolver` as the key in the options object.

oded-soffrin:
Fixed and move to this PR due to deploy scripts - https://github.com/wix-incubator/rich-content/pull/829

PR #825, opened 2020-03-23
Prop types fix

oded-soffrin:
@talevy17 is that what you meant?


oded-soffrin:
Sounds good. Disregard this PR than

PR #832, opened 2020-03-24
[viewer] fix: emoji changes from left side to right side of text

barboaron:


packages/viewer/web/src/utils/textUtils.js
```diff
@@ -1,7 +1,26 @@
-import direction from 'direction';
+const RTL = '\u0591-\u07FF\uFB1D-\uFDFD\uFE70-\uFEFC';
```
shaulgo: move to new file (direction.js)

PR #844, opened 2020-03-26
[link-preview] fix: convert pasted link to link preview

danarish4:


packages/plugin-link/web/src/createLinkPlugin.js
```diff
@@ -80,7 +80,7 @@ const createLinkPlugin = (config = {}) => {
   const shouldLinkify = consecutiveString =>
     consecutiveString.string.length >= settings.minLinkifyLength &&
     isValidUrl(consecutiveString.string) &&
-    !rangeContainsEntity(consecutiveString);
+    !(rangeContainsEntity(consecutiveString) && rangeContainsPlainText(consecutiveString));
```
shaulgo: how can it be a validUrl and not contain text?

danarish4: the mean is not contain any plain text rather  then the url itself (for case of copy url from the browser and then paste it in the editor - it paste it as a link - in that case I want to convert it to link preview)

PR #845, opened 2020-03-26
Fix/fullscreen design

talevy17:
Click below to open examples:
rich-content: https://rich-content-fullscreen-design.surge.sh/
rich-content-storybook: https://rich-content-storybook-fullscreen-design.surge.sh/


packages/fullScreen/web/src/FullScreen.jsx
```diff
@@ -1,6 +1,6 @@
 import React, { Component } from 'react';
 import ReactDOM from 'react-dom';
-import closeIcon from './icons/close.svg';
+import closeIcon from './icons/xl.svg';
```
shaulgo, line 3: change name

PR #870, opened 2020-03-31
[plugin-emoji] reduce the bundle size

barboaron:


packages/plugin-emoji/web/src/utils/getGroupEmojis.js
```diff
@@ -1,21 +1,17 @@
-import { toArray } from 'lodash';
 import convertShortNameToUnicode from './convertShortNameToUnicode';
-import createEmojisFromStrategy from './createEmojisFromStrategy';
 import emojiList from './emojiList';
-import strategy from 'emojione/emoji.json';
 
 const getGroupEmojis = category => {
-  const emojis = createEmojisFromStrategy(strategy)[category];
+  const emojis = emojiList[category];
```
shaulgo: is this var needed?

packages/plugin-emoji/web/src/utils/getGroupEmojis.js
```diff
@@ -1,21 +1,17 @@
...
   const emojisDisplay = [];
-  toArray(emojis).map(shortName => {
+  emojis.map(shortName => {
```
shaulgo: you used map as if it's foreach. map is the corect style here. But look how to use map

packages/plugin-emoji/web/src/utils/getGroupEmojis.js
```diff
@@ -1,21 +1,17 @@
...
     try {
```
shaulgo: try to remove the `try-catch` why do we need it here

packages/plugin-emoji/web/src/utils/getGroupEmojis.js
```diff
@@ -1,21 +1,17 @@
...
-      emojisDisplay.push(convertShortNameToUnicode(emojiList.list[shortName][0]));
+      emojisDisplay.push(convertShortNameToUnicode(shortName));
```
shaulgo: rename function to what it does

PR #878, opened 2020-04-02
feat/gitComment

maimonav:
Click below to open examples:
rich-content: https://rich-content-gitComment.surge.sh/
rich-content-storybook: https://rich-content-storybook-gitComment.surge.sh/


.github/workflows/pr_actions.yml
```diff
@@ -0,0 +1,18 @@
+# name: action
```
shaulgo, line 1: delete?

.github/workflows/pr_actions.yml
```diff
@@ -0,0 +1,17 @@
...
+jobs:
+  deploy-pr-comment:
+    runs-on: ubuntu-18.04
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v1.1.0
+        with:
+          node-version: 12
+      - run: npm install
+      - run: npm run build
```
shaulgo, line 12: I don't think we need to build for this.

.github/workflows/pr_actions.yml
maimonav: Since we decided to separate it to another action,
I needed to do other .yml file for pr actions.

.github/workflows/pr_actions.yml
maimonav: Ok, i'll check it and then i'll merge. Thanks.

PR #899, opened 2020-04-06
Link preview fixes

danarish4:
Click below to open examples:
rich-content: https://rich-content-linkPreviewFixes.surge.sh/
rich-content-storybook: https://rich-content-storybook-linkPreviewFixes.surge.sh/

### DIFF
--- a/packages/plugin-link-preview/web/src/consts.js
+++ b/packages/plugin-link-preview/web/src/consts.js
@@ -1,6 +1,10 @@
 export const DEFAULTS = {
   alignment: 'center',
   size: 'content',
+  link: {
+    target: '_blank',
+    rel: 'noopener',
+  },
 };
 
 export const REMOVE_LINK_PREVIEW = 'remove-link-preview';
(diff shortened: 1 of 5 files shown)


oded-soffrin:
does the link should open in new window ?

packages/plugin-link-preview/web/src/consts.js
```diff
@@ -1,6 +1,12 @@
...
   size: 'content',
+  link: {
+    target: '_blank',
+    rel: 'noopener',
+  },
 };
 
 export const REMOVE_LINK_PREVIEW = 'remove-link-preview';
+
+export const AUTO_GENERATED_LINK_PREVIEW_PROVIDER = 'wix';
```
shaulgo, line 12: we try to keep wix out of our project. What is this for?

oded-soffrin: That's validation on the service response data, meant to filter out auto-generated HTML by the service. 
but this is a valid point, the filter logic should be moved here
 https://github.com/wix-private/rich-content-services/blob/master/wix-rich-content/src/linkPreviewStrategy/linkPreviewStrategy.ts

oded-soffrin:
merging and add to tasks to remove the wix-referer in regard to  data validation

PR #907, opened 2020-04-07
[wrapper] infra(wrapper): convert to typescript

mharsat:
Click below to open examples:
rich-content: https://rich-content-rich-content-wrapper-typescript.surge.sh/
rich-content-storybook: https://rich-content-storybook-rich-content-wrapper-typescript.surge.sh/


packages/wrapper/web/src/RichContentWrapper.tsx
```diff
@@ -1,23 +1,31 @@
...
   }
 
+  static propTypes = { children: PropTypes.element.isRequired };
+
+  static defaultProps = { locale: 'en' };
+
   updateLocale = async () => {
     const { locale, children } = this.props;
-    await localeStrategy(children.props.locale || locale).then(localeData => {
+    await localeStrategy(children?.props.locale || locale).then(localeData => {
```
Barackos, line 28: let's make `children` prop required, so we won't have to check it's existence (it is needed anyhow, and it will also provide an appropriate error in console.log)

packages/wrapper/web/src/themeStrategy/themeTypes.d.ts
```diff
@@ -0,0 +1,21 @@
+interface ThemeGeneratorFunction {
+  (colors: any, utils: any): any;
```
Barackos, line 2: `colors` can be more specified: an object, defined in `ThemeGenerator.ts`.
`utils` can be specified as an object that contains utils functions (located in `themeStrategy/themes/utils.ts`)

packages/wrapper/web/src/themeStrategy/themes/editor.ts
```diff
@@ -1,5 +1,5 @@
 /* eslint-disable camelcase */
-export default function editor(colors, utils) {
+export default function editor(colors: any, utils: any) {
```
Barackos, line 2: same here: `colors` and `utils` type

packages/wrapper/web/src/themeStrategy/themes/viewer.ts
```diff
@@ -1,5 +1,5 @@
 /* eslint-disable camelcase */
-export default function viewer(colors) {
+export default function viewer(colors: any) {
```
Barackos, line 2: `colors` type

packages/wrapper/web/tests/palettesExample.ts
```diff
@@ -224,7 +224,7 @@ const examples = {
 const paletteToWixPalette = palette =>
   palette.map((color, i) => ({ ...TemplateExample[i], value: color }));
 
-const wixPalettes = {};
+const wixPalettes: any = {};
```
Barackos, line 227: Check `palettesExample.js` inside storybook - this is the type of object that is received.

packages/wrapper/web/src/RichContentWrapper.tsx
mharsat: Done

mharsat: I'm also want required props do be specifically added as proptypes

packages/wrapper/web/src/themeStrategy/themeTypes.d.ts
mharsat: I was really hoping you would take care of some of the `any` types 😃
I created many types for objects I understood/or investigated but some remained enigmatic to me

PR #913, opened 2020-04-12
[editor-common] calculateDiff(draftUtils.js): debounced

Barackos:
Click below to open examples:
rich-content: https://rich-content-calculate-diff-debounce.surge.sh/
rich-content-storybook: https://rich-content-storybook-calculate-diff-debounce.surge.sh/


packages/editor-common/web/src/Utils/draftUtils.js
```diff
@@ -444,18 +445,23 @@ export const calculateDiff = async (prevState, newState, onPluginDelete) => {
...
   // onPluginChange -> for Phase 2
   //else {
   // const before = beforePlugins[key];
   // const after = afterPlugins[key];
   // if (JSON.stringify(before) !== JSON.stringify(after))
   //   onPluginChange(type, { from: before, to: after });
   //}
+}, 300);
+
+export const calculateDiff = (previousState, newState, onPluginDelete) => {
```
shaulgo, line 462: What do you think of this?
`export const createCalcContentDiff = (state, onPluginDelete) => {
  let prevState;
  return debounce((state, onPluginDelete) => {
    ...
    prevState = state
  }, 300);
};`

Barackos: Adopted, looks better

PR #918, opened 2020-04-13
[wrapper + storybook] fix(viewer): re-render on new initialState

Barackos:
Click below to open examples:
rich-content: https://rich-content-wrapper-rerender-fix.surge.sh/
rich-content-storybook: https://rich-content-storybook-wrapper-rerender-fix.surge.sh/


CHANGELOG.md
```diff
@@ -22,6 +22,12 @@
 - `wrapper`
   - [#919](https://github.com/wix-incubator/rich-content/pull/919) added internal mobile support
   - [#920](https://github.com/wix-incubator/rich-content/pull/920) refactor: `FullscreenRenderer.tsx`, `ModalRenderer.tsx`
+- `wrapper + storybook`
+  - [#918](https://github.com/wix-incubator/rich-content/pull/918)
+    - wrapper: viewer is now re-rendered for new initialState prop (fix)
+    - storybook: hotfix (`editor` prop replaced with `isEditor`)
+    - storybook: live example app (viewer + editor side by side)
+    - storybook: live example app in palettes page
```
mharsat, line 30: Please split PR #918 into the lines needed, and place them in context of the package that changed (instead of using package1 + package2).
Like PR #624 was split into 3 here - [https://github.com/wix-incubator/rich-content/blob/master/CHANGELOG.md#670-jan-27-2020](url)

packages/wrapper/web/src/FullscreenRenderer.tsx
```diff
@@ -46,17 +38,20 @@ export default class FullscreenRenderer extends Component<Props, State> {
 
   setData = data => this.setState({ data });
 
+  createHelpers = (helpers: object) => ({ ...helpers, onExpand: this.onExpand });
```
mharsat, line 41: `createHelpers` sounds like a lot of logic, if this is a separate function maybe consider `addFullscreen` or `addExpand`?

packages/wrapper/web/src/FullscreenRenderer.tsx
```diff
@@ -1,8 +1,9 @@
 import React, { Component, Fragment, Suspense, Children, ReactElement } from 'react';
-import { RichContentProps } from './RichContentWrapperTypes';
 
 interface Props {
   children: ReactElement;
+  helpers?: object;
+  initialState: object;
```
mharsat, line 6: We have an `InitialState` type in `RichContentWrapperTypes` that you can use.
Also, I suggest extracting `helpers` from `RichContentProps` in the same way, and then reusing it when referring to a helpers instance.

packages/wrapper/web/src/RichContentWrapper.tsx
```diff
@@ -57,9 +57,14 @@ export default class RichContentWrapper extends Component<
...
+    const { theme } = themeStrategy(isEditor, {
+      theme: childTheme,
+      palette,
+      themeGenerators,
+    });
     const mergedRCProps = merge(
-      pluginsStrategy(isEditor, plugins, children.props),
-      themeStrategy(isEditor, { theme, palette, themeGenerators }),
+      { theme },
+      pluginsStrategy(isEditor, plugins, { ...children.props, theme }),
```
mharsat, line 67: Instead of overriding children props, how about changing the strategy's signature so we can be more explicit. Something like:
`pluginsStrategy((isEditor, plugins: PluginConfig[], theme: SomeNewThemeType, childProps: RichContentProps)`

packages/wrapper/web/src/themeStrategy/themeTypes.d.ts
```diff
@@ -1,5 +1,22 @@
...
+  fallbackColorBright: string;
+  isBright: (hexColor: string) => boolean;
+  adaptForeground: (actionColor: string) => string;
+  hexToRgbA: (hexColor: string, opacity: number) => string;
+}
+
 interface ThemeGeneratorFunction {
-  (colors: any, utils: any): any;
+  (colors: PaletteColors, utils: ThemeUtils): object;
 }
```
mharsat, line 20: Super awesome 🤟🏾

packages/wrapper/web/src/themeStrategy/themes/global.ts
```diff
@@ -0,0 +1,10 @@
+/* eslint-disable camelcase */
+export default function global(colors: PaletteColors) {
```
mharsat, line 2: What does global mean? Please use a more expressive name so even I can understand 😄

PR #926, opened 2020-04-16
[test-env] wrapper

Barackos:


oded-soffrin:
@lxgreen @shaulgo this is the wrapper test if you want to discuss.
I agree duplicate tests are a pain, but we should at least check themeing for each plugin
do you think anything here is redundant ?

shaulgo:
> @lxgreen @shaulgo this is the wrapper test if you want to discuss.
> I agree duplicate tests are a pain, but we should at least check themeing for each plugin
> do you think anything here is redundant ?

I'm talking with Barak about this. I think we should move most of the tests to wrappers. That way we test both wrapper and rce. Only tests that have a config that isn't supported in wrappers should be left without it.

packages/wrapper/web/src/pluginsStrategy/pluginsStrategy.ts
```diff
@@ -39,15 +39,16 @@ function viewerStrategy(
...
     typeMappers: (typeMapper && prev.typeMappers.concat([typeMapper])) || prev.typeMappers,
     decorators:
       (decorator && prev.decorators.concat([decorator(theme, config)])) || prev.decorators,
     inlineStyleMappers:
       (inlineStyleMapper &&
         contentState &&
-        prev.inlineStyleMappers.concat([inlineStyleMapper?.(config, contentState)])) ||
+        prev.inlineStyleMappers.concat([inlineStyleMapper?.(finalConfig, contentState)])) ||
       prev.inlineStyleMappers,
   };
```
Barackos, line 53: This change was needed to solve custom inlineStyleMappers for the viewer.
Identically solved in pr #983.

Barackos:
This branch has unexpected behaviour in Applitools, tests stopped passing.
Squashed to (#1045)

PR #929, opened 2020-04-16
[viewer] fix: empty lists viewer issues

maimonav:
Click below to open examples:
rich-content: https://rich-content-emptyListViewer.surge.sh/
rich-content-storybook: https://rich-content-storybook-emptyListViewer.surge.sh/


packages/viewer/web/src/utils/convertContentState.jsx
```diff
@@ -13,7 +13,9 @@ import { combineMappers } from './combineMappers';
 import { getInteractionWrapper, DefaultInteractionWrapper } from './getInteractionWrapper';
 
 const isEmptyContentState = raw =>
-  !raw || !raw.blocks || (raw.blocks.length === 1 && raw.blocks[0].text === '');
+  !raw ||
```
shaulgo, line 16: why is this change needed?

maimonav: It solves the edge case which is when the editor has empty list and the viewer doesn’t show anything, with this change, the viewer gets render with empty list.

PR #934, opened 2020-04-19
Notify build recovered

mharsat:
Click below to open examples:
rich-content: https://rich-content-notify-recovery-on-slack.surge.sh/
rich-content-storybook: https://rich-content-storybook-notify-recovery-on-slack.surge.sh/


mharsat:
> github.event.before will always be the previous commit?

Yes, the [push event](https://developer.github.com/v3/activity/events/types/#events-api-payload-34) holds the SHA of the most recent commit to `master`

PR #941, opened 2020-04-20
gallery settings panel tooltips

ShamaySapir:


packages/editor-common/web/src/Components/Checkbox.jsx
```diff
@@ -79,15 +78,7 @@ export default class Checkbox extends React.Component {
...
-        {contentForInfoIcon && (
-          <Tooltip
-            shouldRebuildOnUpdate={() => true}
-            content={contentForInfoIcon}
-            theme={styles.theme}
-          >
-            <InfoIcon className={styles.checkbox_infoIcon} />
-          </Tooltip>
-        )}
+        <InfoIcon tooltipTextKey={tooltipTextKey} />
```
shaulgo, line 81: does it need the translation func?

packages/editor-common/web/src/Components/InputWithLabel.jsx
```diff
@@ -34,14 +36,17 @@ class InputWithLabel extends Component {
...
     if (label) {
       return (
-        <label htmlFor={id}>
-          <span className={styles.inputWithLabel_label}>{label}</span>
+        <div>
+          <label htmlFor={id} className={generalstyles.infoContainer}>
+            <span className={styles.inputWithLabel_label}>{label}</span>
+            <InfoIcon iconStyles={styles.infoIcon} tooltipTextKey={tooltipTextKey} />
+          </label>
           {this.renderInput()}
```
shaulgo, line 47: why did you move this out from the label?

packages/editor-common/web/statics/styles/input-with-label.scss
```diff
@@ -35,3 +35,11 @@
 .inputWithLabel_fullHeight {
   height: 100%;
 }
+
+.infoIcon {
```
shaulgo, line 39: seems like almost duplicate from the general.scss

packages/editor-common/web/src/Components/Checkbox.jsx
ShamaySapir: no, the translation func is inside InfoIcon, so we only need to send the tooltip key

packages/editor-common/web/statics/styles/input-with-label.scss
ShamaySapir: you're right, but it doesn't look good with the padding-top: 2px that there is in the general.scss, I needed to avoid this style in the case of input with lable.

packages/editor-common/web/src/Components/InputWithLabel.jsx
ShamaySapir: I did it for style reasons, but I found a better way.

PR #942, opened 2020-04-20
[vertical-embed] fix: replace modal UI fix

danarish4:


oded-soffrin:
still not working on example app... (clicking product button, drop down of the modal is empty can't add products..)

PR #949, opened 2020-04-21
[link-preview] feat: Let the consumers decide which services they wan…

danarish4:


packages/plugin-link-preview/web/src/lib/utils.js
```diff
@@ -10,16 +10,20 @@ import {
 } from 'wix-rich-content-editor-common';
 
 export const addLinkPreview = async (editorState, config, blockKey, url) => {
+  const fixedUrl = url.split('\u21b5').join(''); //remove {enter} char
```
shaulgo, line 13: What's this char? how does it enter the middle of the url? Your comment makes me think it might have to do only with the tests?
(by the way, replace would be the more direct way to do this)

packages/plugin-link-preview/web/src/lib/utils.js
```diff
@@ -44,13 +48,23 @@ const isValidImgSrc = url => {
...
+  if (enableLinkPreview && title && thumbnail_url) {
     return isValidImgSrc(thumbnail_url);
   }
   return false;
 };
 
+const shouldAddEmbed = (html, enableEmbed, url) => {
+  if (Array.isArray(enableEmbed)) {
+    return (
+      enableEmbed.filter(whiteListType => url.toLowerCase().includes(whiteListType.toLowerCase()))
```
shaulgo, line 61: you could use `some` instead of filter

PR #955, opened 2020-04-23
code refactoring to github action-bundleSize

barboaron:


plugins-bundle-analyzer/bundleSizes.json
```diff
@@ -0,0 +1,28 @@
+{
```
shaulgo, line 1: why do we need this file?

scripts/gitPRComment.js
```diff
@@ -2,14 +2,16 @@ const github = require('@actions/github');
...
   const { REPO_TOKEN } = process.env;
-  const request = {
-    owner: github.context.repo.owner,
-    repo: github.context.repo.repo,
-    pull_number: github.context.payload.pull_request.number,
-  };
-  request.body = message;
-  const client = new github.GitHub(REPO_TOKEN);
-  await client.pulls.update(request);
+  if (REPO_TOKEN) {
```
shaulgo, line 5: when is the repo_token missing?

plugins-bundle-analyzer/compareBundles.js
```diff
@@ -4,15 +4,67 @@ const chalk = require('chalk');
...
+
+const generatePRComment = () => {
+  let message = newBundles
+    ? 'New packages found.\nPlease update the baseline file by running locally "npm run analyzeBundles" and push the changes.\n\n'
+    : '';
+  message = message.concat(grewDownMessage ? `Packages that shrank:\n${grewDownMessage}\n` : '');
+
+  !newBundles &&
+    grewDownMessage &&
+    (message = message.concat(
```
shaulgo, line 21: no need to change. Usually, we do `str1 + str2`

plugins-bundle-analyzer/bundleSizes.json
barboaron: I wanted the analysis and comparison to be done in separate stages. so it saves the baseline and the current bundles and then compares them

scripts/gitPRComment.js
barboaron: When we run the comparison script locally.
The comparison script calls to the gitPRComment function, but locally we don't want that this function will be execute (because it will fail the comparison test).

plugins-bundle-analyzer/bundleSizes.json
shaulgo: What about a function that analyzes without saving. Then it can be called by other functions either to print to screen or to compare?

barboaron: It's definitely possible, I will change it after I fix the bug in the gallery

plugins-bundle-analyzer/compareBundles.js
barboaron: I changed it to str1+str2

PR #966, opened 2020-04-26
[wrapper] infra(rcw): children prop no longer required

mharsat:
Packages that shrank:
viewer-with-wrapper: 524KB => 174KB

Please update the baseline file by running locally "npm run analyzeBundles" and push the changes.


packages/wrapper/web/src/RichContentWrapper.tsx
```diff
@@ -55,17 +53,39 @@ class RichContentWrapper extends Component<
...
       palette,
       plugins = [],
-      children,
       isEditor = false,
       rcProps,
       forwardedRef,
       ...rest
     } = this.props;
+
+    const child = this.getChildComponent();
```
Barackos, line 86: It can move to `componentDidMount`, so it'll be triggered only once and not in every render. Over calculations IMO. Take a look at `ModalDialogProvider` for example.
What do you think?

packages/wrapper/web/src/ViewerModal.tsx
```diff
@@ -1,12 +1,10 @@
 import React from 'react';
 import getImagesData from 'wix-rich-content-fullscreen/dist/lib/getImagesData.cjs.js';
 import Fullscreen from 'wix-rich-content-fullscreen';
-import { RawDraftContentState } from 'draft-js';
 
 interface Props {
-  initialState: RawDraftContentState;
+  initialState: ContentState;
```
Barackos, line 6: I like this change - no more RawDraft

packages/wrapper/web/src/RichContentTypes.d.ts
```diff
@@ -2,8 +2,13 @@ type Theme = {
...
   [propName: string]: any;
 };
+
 type TextToolbarType = 'inline' | 'static';
+
 type Helpers = { [propName: string]: (...args: any[]) => any };
 
-declare module 'wix-rich-content-editor';
-declare module 'wix-rich-content-viewer';
+type DraftContentState = import('draft-js').RawDraftContentState;
```
Barackos, line 10: wouldn't this work? :
`import { RawDraftContentState } from 'draft-js'`?

packages/wrapper/web/src/RichContentWrapper.tsx
mharsat: Interesting, but what if consumer implements something like this
```jsx
<RichContentWrapper>
  {isEditing ? <RichContentEditor /> : <RichContentViewer />}
</RichContentWrapper>
```
It means we need to keep listening to changes in `children`

packages/wrapper/web/src/RichContentTypes.d.ts
mharsat: that code doesn't work in declaration files, this is the workaround

PR #969, opened 2020-04-27
improve truncatContentState func

maimonav:


e2e/tests/fixtures/truncateContentState.json
```diff
@@ -0,0 +1,113 @@
+{
```
shaulgo, line 1: let's move this to the test folder. it is only for the contentState spec

packages/common/web/src/test/contentStateServices.spec.js
```diff
@@ -1,7 +1,7 @@
 /* eslint-disable max-len */
-import plain from '../../../../../e2e/tests/fixtures/plain.json';
+import raw from '../../../../../e2e/tests/fixtures/truncateContentState.json';
 import { truncateContentState } from '../lib/contentStateServices';
-const plainTop2Blocks = {
+const expectedRaw = {
```
shaulgo, line 4: let's move this to the same file with the state. 
So we have contentStateServices.specData.js
This file can have the data, truncated0, truncated1..., truncated100

e2e/tests/fixtures/truncateContentState.json
maimonav: fixed.

packages/common/web/src/test/contentStateServices.spec.js
maimonav: I did it in a different way before I saw your comment, have a look and tell me if the way i did it suits you.

PR #970, opened 2020-04-27
[vertical-embed] feat: search items

danarish4:


packages/plugin-vertical-embed/web/src/toolbar/postSelectionInputModal.jsx
```diff
@@ -1,55 +1,25 @@
...
+import { contentTypeMap } from '../constants';
 export default class PostSelectionInputModal extends Component {
+  constructor(props) {
+    super(props);
+    const {
+      fetchFunctions,
+      componentData: { type },
+    } = props;
+    this.fetcher = fetchFunctions[type];
+    this.fetcher.get().then(products => this.setState({ products }));
```
oded-soffrin, line 13: as discussed, we can remove the get() and just use search with empty searchTerm
also, fetchFunctions are now serachmodules or verticalAPI, so we can reneame
this.verticalApi.Search('something')

PR #975, opened 2020-04-28
[wrapper] fix(RichContentWrapper): Suspense not supported in SSR

Barackos:


packages/wrapper/web/src/RichContentWrapper.tsx
```diff
@@ -107,20 +111,22 @@ class RichContentWrapper extends Component<
...
+        isEditor={isEditor}
+        initialState={contentState}
+        {...rest}
+      >
+        {child}
+      </EngineWrapper>
     );
+
+    return this.shouldUseSuspense
+      ? isMounted && <Suspense fallback={<div />}>{engineWrapper}</Suspense>
```
shaulgo, line 128: Did you read about how to do this? This seems weird. Would it render engineWrapper in ssr and in the first render? And then add suspense for any later renders?

Barackos: True this is a quick & temporary solution for now because SSR crashed.
I noted to myself to create a different mechanism in which there'll be a "SmartSuspense" specifically for the child, that would forward props & ref for the child. At the moment SSR will not render the EngineWrapper, as you say

PR #982, opened 2020-04-30
[*] fix(package.json): concurrent watch + e2e wait-on serve ready

Barackos:


oded-soffrin:
`concurrently` and `wait-on` both rings a bell :)
is there a way we could define this scripts one time via lerna root?
it's super tedious  if everytime we want something changed in the plugin environment means we have to duplicate it across all plugins... 

maybe devDependency could resolved by lerna hosting, and using --prefix can move all the scripts into one line in main package.json

Barackos:
> `concurrently` and `wait-on` both rings a bell :)
> is there a way we could define this scripts one time via lerna root?
> it's super tedious if everytime we want something changed in the plugin environment means we have to duplicate it across all plugins...
> 
> maybe devDependency could resolved by lerna hosting, and using --prefix can move all the scripts into one line in main package.json

Barackos:
> `concurrently` and `wait-on` both rings a bell :)
> is there a way we could define this scripts one time via lerna root?
> it's super tedious if everytime we want something changed in the plugin environment means we have to duplicate it across all plugins...
> 
> maybe devDependency could resolved by lerna hosting, and using --prefix can move all the scripts into one line in main package.json

Actually we can, didn't think about it - We can create a `npm run watch` with an argument containing a postfix of the package name, and use concurrently only from the main package.json. I'll work it out & push a fix

PR #984, opened 2020-04-30
[common] fix: implemented func, no use of draft

maimonav:


packages/common/web/src/lib/contentStateServices.js
```diff
@@ -1,13 +1,14 @@
...
   }
-  const truncateBlocks = blocks.slice(0, index);
-  const newContentState = ContentState.createFromBlockArray(truncateBlocks);
-  const newRaw = convertToRaw(newContentState);
-  return newRaw;
+  const newEntityMap = {};
+  const newBlocks = [...contentState.blocks.slice(0, index)];
+  newBlocks.forEach(block => {
+    if (block.entityRanges?.length > 0) {
+      const key = block.entityRanges[0].key;
```
shaulgo, line 9: does this handle a case where there is more than one entity for a block? Also, make sure the tests have blocks with 0,1,2 entities. You can have more than one entity with links.

maimonav: no, I didn't know that it's possible.
I'm on it.
Thanks

PR #991, opened 2020-05-03
[storybook] fix(Preview): css console warning: "flex-end" instead of "end"

Barackos:


Barackos:
Solves this warning:
Screen Shot 2020-05-03 at 8 43 04

PR #1000, opened 2020-05-04
content state validator revise

oded-soffrin:


oded-soffrin:
@talevy17 continuing your fine work

packages/common/web/src/lib/contentStateValidator.spec.js
```diff
@@ -0,0 +1,47 @@
+import { isValidEditorData } from './contentStateValidator';
```
lxgreen, line 1: kudos!

PR #1031, opened 2020-05-07
[wrapper] fixed(build-warning): "punycode" external lib

Barackos:


Barackos:
Solves this warning when building the wrapper.
Those external libs warnings come up as error logs when working with Yoshi / CJS sometimes (happened to us with JSS).
Screen Shot 2020-05-07 at 10 09 14

PR #1045, opened 2020-05-07
[test-env] wrapper - new branch

Barackos:


Barackos:
squashed from (#926)

e2e/test-env/src/client/WrapperTestApp/WrapperTestApp.jsx
```diff
@@ -0,0 +1,70 @@
...
+      <RichContentWrapper plugins={editorPlugins} isEditor>
+        <RichContentEditor
+          placeholder={'Add some text!'}
+          onChange={onEditorChange}
+          editorState={editorState}
+          isMobile={isMobile}
+          locale={locale}
+          localeResource={localeResource}
+        />
+      </RichContentWrapper>
```
mharsat, line 24: Please remove the child so we actually test the new API

e2e/test-env/src/client/WrapperTestApp/WrapperTestApp.jsx
```diff
@@ -0,0 +1,70 @@
...
+
+    return (
+      <RichContentWrapper plugins={viewerPlugins}>
+        <RichContentViewer
+          initialState={contentState}
+          isMobile={isMobile}
+          locale={locale}
+          seoMode={seoMode}
+        />
+      </RichContentWrapper>
```
mharsat, line 39: Same goes here, remove the child

package.json
```diff
@@ -41,7 +41,7 @@
     "e2e:app": "cd e2e/test-env && npm run build && npm start",
     "e2e:app:run": "cd e2e/test-env && npm start",
     "e2e:app:dev": "cd e2e/test-env && npm run dev",
-    "e2e:app:wait": "wait-on http://localhost:3002/",
+    "e2e:app:wait": "wait-on http://localhost:3002/ --delay 10000 --timeout 180000 --interval 2000",
```
mharsat, line 44: Does `--delay` mean we always have to wait 10 sec even if everything is ready? Is it necessary when using the timeout as well? If there is no way to listen for something to happen instead of waiting, consider reducing the time 10 sec is a lot to wait

packages/wrapper/web/tsconfig.json
```diff
@@ -11,9 +11,9 @@
     "module": "ESNext",
     "moduleResolution": "node",
     "importHelpers": true,
-    "target": "es5",
+    "target": "ES5",
     "jsx": "react",
-    "lib": ["dom", "esnext"],
+    "lib": ["dom", "ESNext", "ES2015", "ES5", "ES2020"],
```
mharsat, line 16: Sure you need that? `esnext` already includes all previous APIs
https://www.typescriptlang.org/v2/tsconfig#lib

e2e/test-env/package.json
```diff
@@ -32,6 +32,9 @@
     "webpack-dev-middleware": "^3.7.0",
     "webpack-hot-server-middleware": "^0.6.0",
     "webpack-node-externals": "^1.7.2",
+    "wix-rich-content-plugin-button": "7.6.0",
```
lxgreen, line 35: why do we have a fixed versions?

e2e/test-env/src/client/WrapperTestApp/editorPlugins.js
```diff
@@ -0,0 +1,90 @@
+import { pluginLinkButton } from 'wix-rich-content-plugin-button';
```
lxgreen, line 1: vertical embed?

package.json
Barackos: True, will fix it - meant for it to exist on `e2e:debug` only.. will modify it (on debug it just prints lots of `waiting` logs so I wanted to delay it a little bit)

packages/wrapper/web/tsconfig.json
Barackos: Forgot about it, will remove

e2e/test-env/package.json
Barackos: idk actually, but it upgrades itself in each release automatically. Would it make a difference to have it with `^`?

lxgreen: oh, it's lerna.. nvm

Barackos:
> Finally testing is done on Ricos 🎉
> All tests will be converted to run loadWrapperEditorAndViewer right?

Indeed

PR #1047, opened 2020-05-07
Componet Testing

oded-soffrin:
faster than cypress, for cases we just want to make sure stuff work and not care of how they look
motivation was the map e2e test, which fails a lot and we ignore all the visual there because it changes all the time


examples/main/shared/theme/theme.js
```diff
@@ -1,4 +1,8 @@
 const isSSR = typeof window === 'undefined';
+
+if (typeof require.context === 'undefined') {
+  require.context = () => ({ keys: () => [] });
+}
```
oded-soffrin, line 5: hack to make it work in tests

packages/editor-common/web/src/Base/baseToolbarButton.jsx
```diff
@@ -285,6 +285,7 @@ class BaseToolbarButton extends React.Component {
           dataHook={this.getDataHook()}
           onChange={decoratedOnChange}
           getValue={decoratedGetValue}
+          theme={theme}
```
oded-soffrin, line 288: another fix to 'merge styles' errors, wasn't passed here

packages/editor/web/src/RichContentEditor/Toolbars/SideToolbar/SideToolbarPluginsSection.js
```diff
@@ -33,7 +33,10 @@ const SideToolbarPluginsSection = ({
       ? plugins.filter(({ section: pluginSection }) => pluginSection === section)
       : plugins;
     return (
-      <div className={classNames(Styles.section, horizontalMenu && Styles.horizontalMenu)}>
+      <div
+        key={section || 'section'}
```
oded-soffrin, line 37: fix to warning - Each child in an array or iterator should have a unique "key" prop... @danarish4

PR #1054, opened 2020-05-09
[editor-common] refactor(calculateDiff)

Barackos:


packages/editor-common/web/src/Utils/draftUtils.js
```diff
@@ -465,22 +466,30 @@ export const createCalcContentDiff = editorState => {
...
+  return (newState, { onPluginDelete } = {}) =>
+    calculate(newState, {
+      shouldCalculate: !!onPluginDelete,
+      onCallbacks: ({ pluginsDeleted }) => {
+        pluginsDeleted.forEach(type => {
+          onPluginDelete?.(type, version);
+        });
+      },
+    });
+}
```
Barackos, line 491: I think I over-complicated things here, but can't think of a way to simplify.
I want onPluginDelete (or any other future callback) to be able to change during the lifetime of the editor, so it can't be provided in `createPluginCallbacksHandler`, it's a constraint.

Barackos: - Move to editor, without callback
- createContentMutationEvents rename

lxgreen: > I want onPluginDelete (or any other future callback) to be able to change during the lifetime of the editor
didn't get it. how should it change?

Barackos: Via the props (not that I have an example, but theoretically, one can provide a different callback during the lifetime of the editor).
But after rethinking I'm not sure that's an issue for now...

PR #1073, opened 2020-05-12
[*] Reduce editor bundle size (internal dynamic loading)

barboaron:


oded-soffrin:
@shaulgo also can be probably closed? we took care of it some other way?

PR #1076, opened 2020-05-12
[editor] fix: link toolbar in storybook - missing uiSettings

noamtwix:


packages/editor/web/src/RichContentEditor/RichContentEditor.jsx
```diff
@@ -58,6 +58,9 @@ class RichContentEditor extends Component {
       uiSettings.blankTargetToggleVisibilityFn || (anchorTarget => anchorTarget !== '_blank');
     uiSettings.nofollowRelToggleVisibilityFn =
       uiSettings.nofollowRelToggleVisibilityFn || (relValue => relValue !== 'nofollow');
+    if (!props.config.uiSettings) {
+      props.config.uiSettings = uiSettings;
```
shaulgo, line 62: it's considered bad practice to change the props in this way. Where does it break?

noamtwix: Uncaught TypeError: Cannot read property 'linkPanel' of undefined
at LinkPanelContainer.render

the refactor doesn't do exactly the same as before.
if there is no uiSettings in the config, so nothing happens.

noamtwix: I'm guessing that all of our consumers (and in the ExampleApp) passing uiSettings in the config and this is the reason that nobody felt it.
In the StoryBook we don't pass uiSettings in the config and that is the reason it crashes.

PR #1078, opened 2020-05-13
[editor] internal: add new group to inline toolbar

maimonav:


shaulgo:
This should work, but I'm wondering what if we moved all the positions to a config for RCE. I believe that might be even better. But let's not put any more time to it right now. Maybe in the future. 
It will not need any logic. It would just get an array of arrays that would have the exact place of all the buttons.

PR #1084, opened 2020-05-13
Feat/external toolbar

lxgreen:
### :rocket: New Feature
- `editor`
  - external toolbar API added
### :bug: Bug Fix
- `plugins`
  - svg icon id conflicts fixed
### :house: Internal
- `exampleApp`
  - external plugin sidebar added
- `storybook`
  - initial intent example added


examples/main/src/Components/ExternalToolbar.jsx
```diff
@@ -0,0 +1,59 @@
...
+              return (
+                <Tooltip content={tooltip} place="right">
+                  <button onClick={onClick} disabled={isDisabled()}>
+                    <Icon />
+                  </button>
+                </Tooltip>
+              );
+            } else if (buttonType === 'file') {
+              return (
+                <FileInput {...fileInputProps} place="right">
```
oded-soffrin, line 44: This is how it works in RCE too? the fileinput is part of the button?

packages/editor/web/src/withPluginButtons.jsx
```diff
@@ -0,0 +1,26 @@
...
+        );
+        this.setState({ buttonsCollection });
+      });
+    }
+
+    render() {
+      return <WrappedComponent buttons={this.state.buttonsCollection || {}} {...this.props} />;
+    }
+  }
+  return PluginButtonProvider;
```
oded-soffrin, line 25: love hocs. too bad we can't use useEffect yet.

examples/main/src/Components/ExternalToolbar.jsx
lxgreen: yep, all the button-like looking file inputs

PR #1086, opened 2020-05-14
[editor] fix: danger declaration of uiSettings in config

noamtwix:


oded-soffrin:
React component shouldn't change its own props on constructor

PR #1091, opened 2020-05-16
iframes isolation 

oded-soffrin:
fix for security issue on ugc in HTML plugin code


oded-soffrin:
@lxgreen @shaulgo 
Added props 'iframeDomain' in viewer & editor, WDYT ?

PR #1133, opened 2020-05-21
[wrapper] improve(performance): strategies

Barackos:


packages/ricos-viewer/web/src/RicosEngine.tsx
```diff
@@ -48,12 +49,10 @@ export class RicosEngine extends Component<EngineProps, EngineState> {
...
-    const { theme: themeStrategyResult } = themeStrategy(
-      isViewer,
-      themeGeneratorFunctions,
-      theme?.palette,
-      cssOverride
-    );
+    const { theme: themeStrategyResult } =
+      theme || cssOverride
+        ? themeStrategy(isViewer, themeGeneratorFunctions, theme?.palette, cssOverride)
+        : { theme: defaultTheme };
```
mharsat, line 55: This moves some theme strategy logic from its proper modular home to the more general RicosEngine. Technically nothing changed, bc if no palette is provided the theme strategy already skips its internal palette logic and returns the default theme with cssOverride (if given).
Importing the default theme from themeStrategy in line 8 explicitly shows we are moving the theme strategy's logic up to the engine.

packages/ricos-viewer/web/src/pluginsStrategy/pluginsStrategy.ts
```diff
@@ -56,13 +57,16 @@ function viewerStrategy(
...
+  plugins: PluginConfig[],
   childProps: RichContentProps,
   cssOverride: CssOverride,
   content?: RicosContent
 ): PluginsStrategy {
   let strategy: EditorPluginsStrategy | ViewerPluginsStrategy;
 
+  if (plugins.length === 0) {
+    return (isViewer ? emptyViewerPlugins : emptyEditorPlugins) as PluginsStrategy;
+  }
```
mharsat, line 69: This also doesn't change the current behavior, bc the lines below already create a proper editor/viewer empty plugin strategy, and simply returns it if there are no plugins (reduce on an empty array with an initial value).
Moreover, this harms backwards compatibility bc if plugins were given to the child component in the legacy format we never reach line 86 that handles that.

PR #1138, opened 2020-05-21
Design/file upload plugin

talevy17:


packages/plugin-file-upload/web/src/consts.js
```diff
@@ -14,23 +14,47 @@ export const DEFAULTS = {
...
       color: textColor,
     },
+    file_upload_extension: {
+      color: textColor,
+    },
     file_upload_type: {
       color: textColor,
     },
     file_upload_icon: {
+      color: actionColor,
```
shaulgo, line 40: use currentColor in the svg. You can find it in other svg to see how it looks.
It then uses the color it inherits.

packages/plugin-file-upload/web/src/consts.js
```diff
@@ -14,23 +14,47 @@ export const DEFAULTS = {
   },
 };
 
+function addOpacityToRGB(rgb, alpha) {
+  const rgbaArr = [
```
shaulgo, line 18: 1. this is very strange. Seems like it is getting not rgb but only one color and alpha.
2. no need for array and join. Just + strings

packages/plugin-file-upload/web/src/consts.js
```diff
@@ -14,23 +14,47 @@ export const DEFAULTS = {
...
+    rgb.toString(16),
+    Math.round(alpha * 255)
+      .toString(16)
+      .substring(0, 2),
```
shaulgo, line 22: this will always be 1 or 2 chars?
the number is 0 to 255 => hex => 0 to FF

packages/plugin-file-upload/web/src/file-upload-viewer.jsx
```diff
@@ -30,34 +31,99 @@ class FileUploadViewer extends PureComponent {
...
+      this.switchReadyIcon();
+    }
   }
 
-  renderError = error => {
-    if (!error) {
-      return null;
-    }
+  switchReadyIcon = () => {
+    return this.setState({ showReadyIcon: true }, () =>
```
shaulgo, line 40: this is an icon that disappears after 2 seconds?

packages/plugin-file-upload/web/src/file-upload-viewer.jsx
```diff
@@ -30,34 +31,99 @@ class FileUploadViewer extends PureComponent {
...
+          ? size + 'B'
+          : size < 1000000
+          ? (size / 1000).toFixed(2) + 'KB'
+          : (size / 1000000).toFixed(2) + 'MB';
+      type_string = type_string + ' • ' + sizeString;
+    }
+    return { infoString: type_string, infoStyle: this.styles.file_upload_type };
+  }
+
   renderViewerBody({ type, name }) {
```
shaulgo, line 110: what is viewer?

packages/plugin-file-upload/web/src/file-upload-viewer.jsx
```diff
@@ -30,34 +31,99 @@ class FileUploadViewer extends PureComponent {
...
+    });
+    if (size) {
+      const sizeString =
+        size < 1000
+          ? size + 'B'
+          : size < 1000000
+          ? (size / 1000).toFixed(2) + 'KB'
+          : (size / 1000000).toFixed(2) + 'MB';
+      type_string = type_string + ' • ' + sizeString;
+    }
```
shaulgo, line 106: extract function
```suggestion
    const infoString = typeString + ' • ' + sizeToString(size);
```

shaulgo: for KB Math.round instead of toFixed
For MB you decide

packages/plugin-file-upload/web/src/consts.js
talevy17: the problem with it was that some elements in the svg needs actionColor and some needs bgColor
I do use currentColor in the svg but it didn't work without further manipulations

packages/plugin-file-upload/web/src/consts.js
talevy17: correct it's an hexadecimal color and not RGB, I'm renaming

packages/plugin-file-upload/web/src/consts.js
talevy17: it's a way of representing the "a" in "rgba" when converting it to hex color (here we already have the hex value of the color so I just convert the opacity factor to it's hex value and concatenate).
The jss library (used for the themeing) doesn't support `rgba($color: textColor, $alpha: 0.2)` so I used this workaround

packages/plugin-file-upload/web/src/file-upload-viewer.jsx
talevy17: Yup, by design

packages/plugin-file-upload/web/src/file-upload-viewer.jsx
talevy17: it's a bad name for the body of the component, was always like that couldn't think of something better

packages/plugin-file-upload/web/src/consts.js
shaulgo: you can either leave it like that. Or it might be nicer to remove the currentcollor completely.  And add classes. 
Maybe actionColor and bgColor. color1 and color2. or something else.
Try it out and see what you like more. Up to you.

talevy17: I don't really like the way it is, I tried classes but it doesn't work properly for some reason.
I'm gonna give it another shot but we shouldn't let this delay this release IMO

packages/plugin-file-upload/web/src/icons/LoaderIcon.jsx
```diff
@@ -10,7 +10,7 @@ const LoaderIcon = props => (
       />
     </defs>
     <g fill="none" fillRule="evenodd" transform="translate(4 4)">
-      <mask id="file-upload-loader-mask" fill="#fff">
+      <mask id="file-upload-loader-mask" fill="currentColor">
```
shaulgo, line 13: we don't like id. Find how to remove.

packages/plugin-file-upload/web/src/icons/ready.svg
```diff
@@ -0,0 +1,21 @@
+<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" width="24" height="24" viewBox="0 0 24 24">
+    <defs>
+        <filter id="prefix__a" width="271.4%" height="340%" x="-85.7%" y="-85.7%" filterUnits="objectBoundingBox">
```
shaulgo, line 3: it can make weird bugs when two svgs have the same id. So we try to remove them completely. If you don't manage to remove them, just make sure they have a unique name.

packages/plugin-file-upload/web/src/icons/LoaderIcon.jsx
shaulgo: it can make weird bugs when two svgs have the same id. So we try to remove them completely. If you don't manage to remove them, just make sure they have a unique name.

packages/plugin-file-upload/web/src/icons/getIcon.jsx
```diff
@@ -0,0 +1,51 @@
...
+    if (typePredicate(data, type)) {
+      retVal = icon;
+      return true;
+    }
+    return false;
+  });
+  return retVal;
+}
+
+function getIcon(type) {
```
shaulgo, line 44: unit test

PR #1139, opened 2020-05-21
Update gallery-component.jsx

zhukovka:
`event.timeStamp` has a floating point so id would look like '12345.54321' results in an invalid selector in `ProGallery`
example https://olenazh3.wixsite.com/social/group/gallery-video/discussion/b1a7df66-a4f5-41ed-9b22-662167f1a73f

Screen Shot 2020-05-21 at 9 55 54 PM
Screen Shot 2020-05-21 at 10 08 12 PM


shaulgo:
Thanks a bunch. Looks good but I rather it run with tests. Closing this. And trying to reopen and see if it runs the tests.

shaulgo:
fixed

PR #1140, opened 2020-05-24
Dynamic import to 'react-tooltip'

barboaron:


shaulgo:
Could you try to find out why some bundles are down by ~14kb and it goes
all the way up to 22kb.
We either have some problems. Or the baselines before are wrong.

PR #1159, opened 2020-05-26
remove editor from viewer

oded-soffrin:
Suggestion to remove editor parts from viewer


packages/ricos-editor/web/src/modals/RicosModal.tsx
```diff
@@ -0,0 +1,24 @@
+/* eslint-disable react/prop-types */
+import React, { Fragment, ComponentType, Children, FunctionComponent } from 'react';
+// import { EngineProps } from '../RicosEngine';
+import EditorModalProvider from './EditorModalProvider';
+
+// const RicosModal: FunctionComponent<EngineProps> = props => {
+const RicosModal: FunctionComponent<any> = props => {
```
oded-soffrin, line 7: @mharsat what's the best way to import types from a different package ?

mharsat: Depends on the use case, but if you only need the `EngineProps` I'd stop using it after these changes. All `EngineProps` does is make `children` a required prop and add the `isViewer` prop. After the changes `isViewer isn't needed any more. I'd go for example something like this for RicosModal editor version (no need to export)

```jsx
interface Props extends RicosEditorProps {
  children: RichContentChild;
  RicosModal: FunctionComponent;
}
```

PR #1186, opened 2020-06-01
Fix/indent with alignment

maimonav:


packages/common/web/statics/styles/draftDefault.rtlignore.scss
```diff
@@ -14,14 +14,14 @@
   }
   @mixin indent_blocks($depth, $indent) {
     .public-DraftStyleDefault-block-depth#{$depth} {
-    &.public-DraftStyleDefault-ltr, & .public-DraftStyleDefault-ltr{
+    &.public-DraftStyleDefault-text-ltr, & .public-DraftStyleDefault-text-ltr{
```
shaulgo, line 17: check if both of these cases are needed

packages/common/web/statics/styles/draftDefault.rtlignore.scss
```diff
@@ -14,14 +14,14 @@
...
        /*!rtl:begin:ignore*/
       margin-left: #{$indent}em !important;
       /*!rtl:end:ignore*/
     }}
   
     .public-DraftStyleDefault-block-depth#{$depth} {
-    &.public-DraftStyleDefault-rtl, & .public-DraftStyleDefault-rtl{
+    &.public-DraftStyleDefault-text-rtl, & .public-DraftStyleDefault-text-rtl{
```
shaulgo, line 24: check if both of these cases are needed

maimonav: fixed, nested `& .public-DraftStyleDefault-text-rtl` is not needed.

packages/common/web/statics/styles/draftDefault.rtlignore.scss
maimonav: fixed, nested & .public-DraftStyleDefault-text-rtl is not needed.

PR #1196, opened 2020-06-02
[e2e] improve(media): convert media tests to ricos

mharsat:


e2e/test-env/src/client/RicosTestApp/RicosTestApp.jsx
```diff
@@ -34,7 +38,11 @@ class RicosTestApp extends PureComponent {
         cssOverride={theme}
         //toolbarSettings={toolbarSettings}
       >
-        <RichContentEditor onChange={onEditorChange} config={testAppConfig.pluginsConfig} />
+        <RichContentEditor
+          onChange={onEditorChange}
+          config={testAppConfig.pluginsConfig}
+          helpers={{ onVideoSelected }}
```
Barackos, line 44: does it mean that `pluginVideo` is not completely supported in Ricos? (I ask because of the helpers)

e2e/tests/__snapshots__/media.e2e.js.snap
```diff
@@ -110,8 +110,6 @@ exports[`plugins > image > render image toolbar and settings #0`] =
         "config": {
           "alignment": "center",
           "link": {
-            "rel": "noopener",
```
Barackos, line 113: Sure about it? default should be `_blank` last time I checked with Dana

lxgreen: @Barackos, first of all, `_blank` is `target` attribute value, not `rel`. 

also, iirc in the project defaults are:

```
target = "_self"
rel = "noopener" | "noopener noreferrer" (in some places)

```

in what context did you discuss it with Dana?

e2e/test-env/src/client/RicosTestApp/RicosTestApp.jsx
lxgreen: +1

really, why not to pass via the `pluginVideo({ onVideoSelected })` ?

mharsat: `onVideoSelected` is part of helpers like the image upload functions, should be moved to the video plugin config. already sent @shaulgo a message on that.

I had to add it here for the same reason I had to do it in the old testing env (video metadata isn't written to content state) [https://app.asana.com/0/1110053346854169/1166561015756179](https://app.asana.com/0/1110053346854169/1166561015756179)

e2e/tests/__snapshots__/media.e2e.js.snap
mharsat: is this resolved then?

e2e/test-env/src/client/RicosTestApp/RicosTestApp.jsx
Barackos: Ok, so for now let's keep it that way.
I'm creating a new Asana ticket for moving `onVideoSelected` to plugin's config so we won't forget it.

e2e/tests/__snapshots__/media.e2e.js.snap
Barackos: I guess so. Looks like we leaned on an example test that didn't cover the default configuration of image (due to shared exampleApp config) so yeah, sounds like a fix actually

PR #1200, opened 2020-06-03
[test-env]social embed: fix test flakiness

danarish4:


e2e/tests/plugins.e2e.js
```diff
@@ -264,14 +264,13 @@ describe('plugins', () => {
 
     after(() => cy.eyesClose());
     const embedTypes = ['TWITTER', 'INSTAGRAM', 'YOUTUBE'];
-    it('render upload modals', function() {
-      embedTypes.forEach(embedType => {
-        const testKey = embedType + ': ' + this.test.title;
+    embedTypes.forEach(embedType => {
```
shaulgo, line 267: this only changes the name of the test?

e2e/tests/__snapshots__/plugins.e2e.js.snap
```diff
@@ -4877,3 +4878,196 @@ exports[`plugins > headings > Change headers - without dropDownOptions config #0
   ],
   "entityMap": {}
 };
+
+exports[`plugins > social embed > render twitter upload modals #0`] =
```
shaulgo, line 4882: what about the old content that isn't used anymore?

shaulgo:
all the visual tests failed. Is it only because the changes were not approved yet?

e2e/tests/plugins.e2e.js
danarish4: No - Instead of one test that tests all modals I split into 3 different tests that test each model individually and this is how it starts from a clean environment every time what solves the flakiness

PR #1214, opened 2020-06-07
[ricos] improve(API): modalSettings

Barackos:


packages/ricos-viewer/web/src/RicosTypes.ts
```diff
@@ -58,7 +65,7 @@ interface ToolbarSettings {
   useStaticTextToolbar?: boolean;
 }
 
-type Helpers = { [propName: string]: (...args: any[]) => any };
+type Helpers = { [propName: string]: ((...args: any[]) => any) | undefined };
```
mharsat, line 68: why is the undefined option needed here?

packages/ricos-viewer/web/src/modals/editorModal/EditorModal.tsx
```diff
@@ -10,13 +10,23 @@ interface Props {
...
+    this.updateModalContainer(this.props);
   }
+
+  componentWillReceiveProps(newProps: Props) {
+    if (newProps.modalContainer !== this.props.modalContainer) {
+      this.updateModalContainer(newProps);
+    }
+  }
+
+  updateModalContainer = (props: Props) => ReactModal.setAppElement(props.modalContainer || 'body');
```
mharsat, line 28: Can you create a story or test for this feature? Maybe something like [this](https://github.com/wix-incubator/rich-content/blob/b60ae75b00123b610a0093e12278b3fe560570b2/examples/storybook/stories/Wrapper/RicosStaticToolbar.js#L40)?

Also how about passing only the relevant prop to `updateModalContainer`, for a simpler API?

packages/ricos-viewer/web/src/RicosTypes.ts
```diff
@@ -50,6 +51,12 @@ type GetToolbarSettings = any; // Should be converted from flow types
 
 type RichContentChild = import('react').ReactElement<ExportedRichContentProps>;
 
+interface RicosModalSettings {
+  openModal?: (data: object) => void;
+  closeModal?: () => void;
+  container?: string;
```
mharsat, line 57: Let's follow a single name - `modalSettings`

packages/ricos-viewer/web/src/RicosTypes.ts
Barackos: Without it, compiler error in [those lines](https://github.com/wix-incubator/rich-content/pull/1214/files#diff-82fc6ead1f81cd7c05fd331ffb836cbaR98-R99)

packages/ricos-viewer/web/src/modals/editorModal/EditorModal.tsx
Barackos: Good idea. Will make those changes (I think test is enough, no need in story).

docs/docs/ricos/ricos-api.mdx
```diff
@@ -159,6 +159,34 @@ This text will appear in the editor until content is created
...
+Provides `aria-hidden` attribute to a given DOM Element ID while the modal is open, making it hidden for screenreaders of users.
+Passed directly into the `ReactModal`'s [AppElement](https://reactcommunity.org/react-modal/accessibility/#app-element).
+
+Default Value: `body`
+```jsx
+<RicosEditor modalSettings={{ ariaHiddenId: '#root' }} />
+```
+
+#### `openModal` & `closeModal`
+Legacy RCE's `openModal`, `closeModal`.
```
lxgreen, line 183:
```suggestion
Legacy RCE's `openModal`, `closeModal` (helpers)[rce_api/helpers].
```

lxgreen: should lead to helpers API doc (which is still empty)

PR #1216, opened 2020-06-07
[test-env] complete ricos coverage

Barackos:


e2e/test-env/src/client/RicosTestApp/RicosTestApp.jsx
```diff
@@ -36,11 +38,11 @@ class RicosTestApp extends PureComponent {
         isMobile={isMobile}
         locale={locale}
         cssOverride={theme}
-        //toolbarSettings={toolbarSettings}
+        toolbarSettings={createToolbarSettings(addPluginMenuConfig)}
       >
         <RichContentEditor
           onChange={onEditorChange}
-          config={testAppConfig.pluginsConfig}
+          config={{ ...uiSettings, ...testAppConfig.pluginsConfig }}
```
mharsat, line 45: are you sure `uiSettings` should be destructed? it's a key in the `config` object.
is `uiSettings` needed for testing?

e2e/test-env/src/client/RicosTestApp/RicosTestApp.jsx
```diff
@@ -56,6 +58,7 @@ class RicosTestApp extends PureComponent {
         isMobile={isMobile}
         locale={locale}
         cssOverride={theme}
+        _rcProps={{ config: { disableRightClick: true } }}
```
mharsat, line 61: please refrain from using `_rcProps` it's for wix ricos only, use child.
also, `disableRightClick` belongs to `uiSettings` not directly in `config`, why do we need it in the test env?

e2e/test-env/src/client/RicosTestApp/RicosTestApp.jsx
Barackos: `uiSettings` is needed for the [linkPanel functionality](https://github.com/wix-incubator/rich-content/pull/1216/files#diff-f96440c7c301701550cb9e005c71c88eR231-R237) on some of the tests, and should be destructed in case one day one would like to override it, as it is a part of the config that needs to support override (as a TestApp)

e2e/test-env/src/client/RicosTestApp/RicosTestApp.jsx
Barackos: Oops, wrongly placed indeed. It also has no effect ATM, I can remove it but shouldn't we want to match RicosTestApp configuration with its predecessor? or alternatively, why not?

PR #1227, opened 2020-06-09
[gallery] fix(gallery-viewer): cursor

talevy17:


packages/plugin-gallery/web/src/gallery-viewer.jsx
```diff
@@ -197,13 +197,26 @@ class GalleryViewer extends React.Component {
...
+    const {
+      settings: { onExpand },
+    } = this.props;
+    const isClickable = onExpand || itemProps.link;
+    const itemHoverElement = this.renderOnHover(itemProps, onExpand);
+    return isClickable ? (
+      <div className={this.styles.galleryItem}>{itemHoverElement}</div>
+    ) : (
+      itemHoverElement
+    );
```
shaulgo, line 217: should probably be something like this
`return (<div className={pointerStyle}>
	{onExpand && this.renderExpandIcon(itemProps)}
	{this.renderTitle(itemProps.title)}
</div>)`

talevy17: That was the previous case but since we need to add a div if an item is clickable I tried to refrain from the code duplication of the inner elements (title and expand icon)

shaulgo: totally agree about code duplication.
Let's fix it with styles like so:
const itemStyle = classNames(styles.galleryItem, {styles.isClickable: isClickable})

PR #1243, opened 2020-06-14
[editor] fix: handle pasted links

maimonav:


packages/editor/web/package.json
```diff
@@ -39,7 +39,7 @@
   },
   "dependencies": {
     "@wix/draft-js": "0.0.7",
-    "draft-convert": "2.1.8",
+    "draft-convert": "^2.1.10",
```
shaulgo, line 42:
```suggestion
    "draft-convert": "2.1.10",
```

shaulgo: we usually use a fixed version.
When it's on the latest, it sometimes fails because of an update and it takes us time to recognize why it failed.

e2e/tests/text.e2e.js
```diff
@@ -133,10 +133,20 @@ describe('text', () => {
...
...width: 0px; background-color: rgb(255, 255, 255); text-decoration-style: initial; text-decoration-color: initial; display: inline !important; float: none;">called<span> </span></span><a href="https://en.wikipedia.org/wiki/Anchor_text" title="Anchor text" style="text-decoration: none; color: rgb(11, 0, 128); background: none rgb(255, 255, 255); font-family: sans-serif; font-size: 14px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px;">anchor text</a><span style="color: rgb(32, 33, 34); font-family: sans-serif; font-size: 14px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(255, 255, 255); text-decoration-style: initial; text-decoration-color: initial; display: inline !important; float: none;">.<span> </span></span>`
```
shaulgo, line 147: what do you think about putting only text and link? Without all the extra styles.

packages/editor/web/src/RichContentEditor/handlePastedText.js
```diff
@@ -21,9 +21,14 @@ const clearAtomicBlockEntities = editorState => {
   return contentState;
 };
 
+//Fixes spaces deletion when pasting hyperlinks
+const replaceEmptySpansWithEmptySpaces = html =>
+  // eslint-disable-next-line no-irregular-whitespace
+  html.replace(/(<span> <\/span>)/g, ' ');
```
shaulgo, line 27: I think this only replaces the first occurrence? You can change the test and see if it fails.

packages/editor/web/src/RichContentEditor/handlePastedText.js
```diff
@@ -21,9 +21,14 @@ const clearAtomicBlockEntities = editorState => {
   return contentState;
 };
 
+//Fixes spaces deletion when pasting hyperlinks
+const replaceEmptySpansWithEmptySpaces = html =>
```
shaulgo, line 25: replaceSpansWithOnlySpaces

packages/editor/web/src/RichContentEditor/handlePastedText.js
shaulgo: let's change it to \s+ with a matching group. So it will work with multiple spaces as well

packages/editor/web/src/RichContentEditor/handlePastedText.js
```diff
@@ -21,9 +21,14 @@ const clearAtomicBlockEntities = editorState => {
   return contentState;
 };
 
+//Fixes spaces deletion when pasting hyperlinks
```
shaulgo, line 24:
```suggestion
//Fixes spaces deletion when pasting hyperlinks (draft-convert trimming)
```

packages/editor/web/package.json
maimonav: Done.

e2e/tests/text.e2e.js
maimonav: I think we should keep the html as is so if perhaps something changes, and it will break the style, we will catch it.

packages/editor/web/src/RichContentEditor/handlePastedText.js
maimonav: Isn't it correct to say empty spans? because we are not changing all spans.

packages/editor/web/src/RichContentEditor/handlePastedText.js
maimonav: Done.

e2e/tests/text.e2e.js
maimonav: And since draft has his way of passing the pasted html, I think it's best that we keep it the same as draft pasted html so it can more accurate and we might detect more problem in the future (I hope that there will be no problems of course, but we might have.. ).

packages/editor/web/src/RichContentEditor/handlePastedText.js
shaulgo: but they are not empty. They have a space character.

shaulgo: even better, replaceSpansContainingOnlySpaces

shaulgo: or, fixSpansContainingOnlySpaces

maimonav: Done.

PR #1244, opened 2020-06-15
Convert all packages to typescript

mharsat:


packages/plugin-image/web/src/typeMapper.js
```diff
@@ -11,6 +10,6 @@ const imageRenderDescriptor = {
 };
 
 export const typeMapper = () => ({
-  [IMAGE_TYPE_LEGACY]: imageRenderDescriptor,
-  [IMAGE_TYPE]: imageRenderDescriptor,
+  IMAGE: imageRenderDescriptor,
+  'wix-draft-plugin-image': imageRenderDescriptor,
```
Barackos, line 14: Why the change?

Barackos, line 14: Why the change?

packages/plugin-video/web/package.json
```diff
@@ -18,6 +18,8 @@
     "dist"
   ],
   "main": "dist/module.cjs.js",
+  "types": "dist/types/src/index.d.ts",
+  "types": "dist/types/src/index.d.ts",
```
Barackos, line 22:
```suggestion
  "types": "dist/types/src/index.d.ts",
```

packages/plugin-image/web/src/typeMapper.js
mharsat: Solved a weird issue "cannot use external module name as key", I followed the pattern done in other `typeMapper.js` files that were converted to flow like [video](https://github.com/wix-incubator/rich-content/blob/master/packages/plugin-video/web/src/typeMapper.js)

packages/plugin-video/web/package.json
mharsat: nice!

PR #1249, opened 2020-06-16
[ricos-viewer] feat: preview strategy

Barackos:


docs/docs/ricos/ricos-api.mdx
```diff
@@ -248,3 +249,34 @@ onChange?: (content: RicosContent) => void
...
+
+### `preview`
+Activates a preview decoration: a shortened version of a content, ready to be expanded.
+Preview props are aligned with the legacy [RichContentPreview](https://github.com/wix-incubator/rich-content/blob/master/docs/docs/rcp_api/rich-content-preview.md) props.
+```jsx
+preview?: {
+  transformation: ContentStateTransformation;
+  contentInteractionMappers: Record<string, unknown>;
+  onPreviewExpand?: () => void;
+}
```
mharsat, line 263: for consistency with the API documentation and for ease of reading, I think a short explanation for each key in the preview settings should be briefly explained

examples/storybook/stories/styles.global.scss
```diff
@@ -16,6 +16,6 @@
 @import '~wix-rich-content-plugin-map/dist/styles.min.css';
 @import '~wix-rich-content-plugin-file-upload/dist/styles.min.css';
 @import '~wix-rich-content-plugin-text-color/dist/styles.min.css';
+@import '~wix-rich-content-preview/dist/styles.min.css';
```
mharsat, line 19: since `wix-rich-content-preview` is not directly consumed anymore, I think it should be added to `RicosViewer`'s `styles.css` so consumers don't need to take care of its styles

docs/docs/ricos/ricos-api.mdx
```diff
@@ -248,3 +249,34 @@ onChange?: (content: RicosContent) => void
 ```
 
 A callback that runs whenever the editor's content is edited.
+
+## RicosViewer API
+
+### `preview`
+Activates a preview decoration: a shortened version of a content, ready to be expanded.
+Preview props are aligned with the legacy [RichContentPreview](https://github.com/wix-incubator/rich-content/blob/master/docs/docs/rcp_api/rich-content-preview.md) props.
```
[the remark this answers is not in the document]
Barackos: done, created a separated doc page

examples/storybook/stories/styles.global.scss
Barackos: followup after our discussion: preview now behaves as a plugin, therefore style will be imported by consumer as any other plugin, until we fix the css issue

Barackos:
> Does the current implementation bundle the preview code into ricos-viewer? Ideally, we should load it only if the prop exists

No bundle of preview code, as we discussed. The consumer will import the necessities similar to plugin consumption.

examples/storybook/stories/Components/ViewerWrapper.js
```diff
@@ -77,15 +77,16 @@ const ViewerWrapper = ({
...
 }) => {
   return (
     <RicosViewer
-      plugins={plugins}
+      plugins={plugins.concat(extraPlugins)}
       theme={{ palette }}
       content={content}
       isMobile={isMobile}
-      preview={preview}
+      previewSettings={previewSettings}
```
mharsat, line 89: looks like this was left over

Barackos: thanks

PR #1250, opened 2020-06-16
Update typeMapper.js

zhukovka:
do not bundle PollEditor for Viewer


oded-soffrin:
master is already updated w/ this change

PR #1257, opened 2020-06-17
[ricos] fix(editor): side toolbar '+' alignment

Barackos:


Barackos:
> looks good, timeout change due to flaky tests I assume?

Yes indeed

PR #1269, opened 2020-06-23
[plugin-gif] feat: add gif to plugin menu

danarish4:


packages/plugin-giphy/web/src/toolbar/giphyApiInputModal.jsx
```diff
@@ -22,68 +22,44 @@ export default class GiphyApiInputModal extends Component {
...
-        tabIndex="0"
-        onKeyPress={null}
-      />
-    );
     const mobileNavbar = (
       <div>
         <div className={styles.giphy_api_input_modal_navbar}>
-          {t('GiphyUploadModal_mobileNavbar_Title')} {backButton}
+          {t('GiphyUploadModal_mobileNavbar_Title')}
+          <div
```
shaulgo, line 37: 1. is there a reason to not leave it in a variable?
2. Why not use button?

packages/plugin-giphy/web/src/toolbar/giphyApiInputModal.jsx
```diff
@@ -22,68 +22,44 @@ export default class GiphyApiInputModal extends Component {
...
-                  onKeyPress={null}
-                  className={styles.giphy_api_input_modal_closeIcon}
-                >
-                  <CloseIcon />
-                </div>
-              )}
-            </div>
-          </div>
+          <TextSearchInput
+            inputRef={ref => {
```
shaulgo, line 54: we try to not create functions inside of render function. Why not use the createRef from react?

PR #1271, opened 2020-06-24
[ricos-common] fix: translation using remountKey

noamtwix:


mharsat:
@noamtwix Why actually do we need to move the whole localeStrategy to RicosViewer/Editor?
We could run the strategy from common and save the code duplication, and only have some `shouldRemount` function passed from editor/viewer to the engine, to be called once the localeResource was resolved

PR #1308, opened 2020-07-05
[ricos] fix(theme): default font - Helvetica

Barackos:


lxgreen:
So the font-family is going to be provided by the consumer/site theme?

Barackos:
> So the font-family is going to be provided by the consumer/site theme?

@lxgreen We actually have a `font-family` definition in [fonts.scss](https://github.com/wix-incubator/rich-content/blob/master/packages/common/web/statics/styles/fonts.scss) (`common` package) which takes place. If consumers will want to change it, they'll have to override it.

PR #1323, opened 2020-07-08
[editor] fix: support copy paste of atomic blocks

danarish4:


packages/editor/web/src/RichContentEditor/RichContentEditor.jsx
```diff
@@ -39,6 +39,7 @@ import styles from '../../statics/styles/rich-content-editor.scss';
 import draftStyles from '../../statics/styles/draft.rtlignore.scss';
 import 'wix-rich-content-common/dist/statics/styles/draftDefault.rtlignore.scss';
 import { deprecateHelpers } from 'wix-rich-content-common/dist/lib/deprecateHelpers.cjs.js';
+import { registerCopySource } from 'draftjs-conductor';
```
shaulgo, line 42: check about extracting this function as well. handlePastedTextFromEditor works together with it. If they change the way it works it will break the other part we took out.

PR #1356, opened 2020-07-15
[preview] feat(video): support videos in gallery

Barackos:


packages/preview/web/src/ContentStateBuilder/mergeEntityData.js
```diff
@@ -27,16 +27,16 @@ const galleryMerger = (items, entity) => ({
...
         ? {
-            itemId: item.url,
+            itemId: item.id || 'video-' + index,
             url: item.url,
             metadata: {
               type: 'video',
               videoUrl: item.mp4,
               videoId: item.url,
-              source: 'giphy',
+              source: item.type.includes('gif') && 'giphy',
```
oded-soffrin, line 39: I think we can split even more,
image/ gif/ video
(in current impl. gif has videoUrl & videoId)

packages/preview/web/src/ContentStateAnalyzer/ContentStateMetadata.js
```diff
@@ -127,6 +127,7 @@ const getContentStateMetadata = raw => {
   });
 
   const media = extractMedia(raw);
+  metadata.galleryItems = media.filter(({ type }) => type.includes('image') || type === 'video');
   metadata.images = media.filter(({ type }) => type.includes('image'));
```
oded-soffrin, line 131: please extract types.includes('image') to helper func isImage

oded-soffrin:
added some minor comments. 
bug with more then 3 gifs still going on, right ?

Barackos:
> added some minor comments.
> bug with more then 3 gifs still going on, right ?

@oded-soffrin fixed the issues, and yes issue still exists with multiple gifx / videos - will tackle it on Sunday with @talevy17

packages/preview/web/src/ContentStateBuilder/galleryEntityData.js
```diff
@@ -0,0 +1,45 @@
+const toGalleryItems = items =>
+  items.map((item, index) => ({
+    itemId: item.id || 'item-' + index,
+    url: item.url,
+    ...getItemByType(item),
+  }));
+
+const getItemByType = item =>
+  item.type === 'giphy' ? gifItem(item) : item.type === 'video' ? videoItem(item) : imageItem(item);
```
mharsat, line 9: point of personal preference, but I think they function would be clearer with a switch over `item.type` or using ifs instead of a one liner

Barackos: I was disappointed when it came out like that. Those ifs can be pretty with a different linter configuration, which I originally expected.

packages/preview/web/src/ContentStateAnalyzer/ContentStateMetadata.js
```diff
@@ -103,6 +103,8 @@ const extractSequentialBlockArrays = ({ blocks }, blockType) => {
 const extractMedia = ({ entityMap }) =>
   Object.values(entityMap).reduce((media, entity) => [...media, ...extractEntityData(entity)], []);
 
+const isGalleryItem = type => type === 'image' || type === 'video' || type === 'giphy';
```
oded-soffrin, line 106: style suggestion: ['image', 'video', 'giphy'].includes(type)

packages/preview/web/src/ContentStateBuilder/galleryEntityData.js
```diff
@@ -0,0 +1,53 @@
...
+      return videoItem(item);
+    default:
+      return imageItem(item);
+  }
+};
+
+const giphyItem = item => ({
+  metadata: {
+    type: 'video',
+    videoUrl: item.mp4,
```
oded-soffrin, line 22: giphy video url mp4? i'm a bit confused. is this a way to make it work in gallery?

packages/preview/web/src/ContentStateBuilder/galleryEntityData.js
```diff
@@ -0,0 +1,53 @@
+const toGalleryItems = items =>
+  items.map((item, index) => ({
+    itemId: item.id || 'item-' + index,
+    url: item.url,
+    ...getItemByType(item),
+  }));
+
+const getItemByType = item => {
```
oded-soffrin, line 8: I think we should revise naming a bit, it's a function that get item and return an item. 
what's the difference between them? what's the origin of the item data and to what it converts to?

packages/preview/web/src/ContentStateBuilder/galleryEntityData.js
Barackos: Yes, gallery does not seem to support a `.gif` file as `videoUrl`, only a video file (tried it and it crashed). But `giphy` has both `.gif` and `.mp4` fields so it helps here.

PR #1371, opened 2020-07-21
Add missing props to ricos

mharsat:


docs/docs/ricos/ricos-api.mdx
```diff
@@ -115,79 +145,134 @@ Defines the palette of colors that will be used. More info in [Theming](theming)
...
+  spellCheck?: boolean;
+  stripPastedStyles?: boolean;
+  handleBeforeInput?: ((chars: string, editorState: EditorState, eventTimeStamp: number) => 'handled' | 'not-handled');
+  handlePastedText?: ((text: string, html: string, editorState: EditorState) => 'handled' | 'not-handled');
+  handleReturn?: ((e: SyntheticKeyboardEvent, editorState: EditorState) => 'handled' | 'not-handled');
+}
 ```
 
-A react [Ref](https://reactjs.org/docs/refs-and-the-dom.html). Use this to access editor functions.
+draft-js props, as described in https://draftjs.org/docs/api-reference-editor
```
Barackos, line 163: current implementation works as a link? ignore this comment if it does
```suggestion
draft-js props, as described in [https://draftjs.org/docs/api-reference-editor](https://draftjs.org/docs/api-reference-editor)
```

docs/docs/ricos/ricos-api.mdx
```diff
@@ -115,79 +145,135 @@ Defines the palette of colors that will be used. More info in [Theming](theming)
 
 ## RicosEditor API
 
-### `ref`
+### `draftEditorSettings`
 
 ```jsx
-ref?: React.Ref
+draftEditorSettings?: {
```
oded-soffrin, line 151: why not `editorSettings` ? tomorrow it won't be draft anymore

docs/docs/ricos/ricos-api.mdx
```diff
@@ -115,79 +145,135 @@ Defines the palette of colors that will be used. More info in [Theming](theming)
...
+ref?: React.Ref
+```
+
+A react [Ref](https://reactjs.org/docs/refs-and-the-dom.html). Use this to access editor functions.
+
+| Function                                                  | Description                |
+| --------------------------------------------------------- | -------------------------- |
+| `focus(): void`                                           | Focuses the editor         |
+| `blur(): void`                                            | Blurs the editor           |
+| `getContent(postId?: string, forPublish?: boolean): void` | Retrieves the content data |
```
oded-soffrin, line 260: getContent is an editor function? i'm confused. and BI event is only in @wix/ricos right?

docs/docs/ricos/ricos-api.mdx
```diff
@@ -242,10 +327,31 @@ import content from '../../infra/editor/content/intro.json';
...
+
+### `preview`
 
 ```jsx
-onChange?: (content: RicosContent) => void
+preview?: {
+  transformation?: ContentStateTransformation;
+  contentInteractionMappers?: Funtion;
+  onPreviewExpand?: () => void;
+}
```
oded-soffrin, line 339: did we finalized already preview as prop and not as plugin config ?

docs/docs/ricos/ricos-api.mdx
```diff
@@ -242,10 +327,31 @@ import content from '../../infra/editor/content/intro.json';
...
+Read more in [Preview](preview)
+
+### `seoSettings`
+
+```jsx
+seoSettings?: boolean | {
+  paywall: {
+    className: string;
+    index: number;
+  };
```
oded-soffrin, line 352: className will be used to wrapped the paid-only content so it can be indexed and improve SEO in one hand, but won't be available in other hand.
index is the count of public blocks to display (we probably need to change naming here)

oded-soffrin:
Such a satisfying PR ! ! !

docs/docs/ricos/ricos-api.mdx
mharsat: when that happens these props will broken. these are totally draft-js specific and calling them that way is more indicative of their purpose. unless we want to commit to this interface even after leaving draft (handleBeforeInput, stripPastedStyles, etc.)

docs/docs/ricos/ricos-api.mdx
mharsat: yes getContent is an editor function. BI event is technically also in rich content's helpers I think, @Barackos can answer that.
anyway - these aren't my changes, you're getting them bc I sorted the API alphabetically

docs/docs/ricos/ricos-api.mdx
mharsat: It was already handled in the API in master, in this case I only added it to the docs

docs/docs/ricos/ricos-api.mdx
mharsat: if you mean a further explanation is required, @maimonav and I agreed that I'll add the `seoSettings` to the docs and then point him to it for explaining the paywall part

mharsat: and if renaming is needed let's discuss that tomorrow w/ aviv

maimonav: I choose to go with index since it's similar to truncateService, which works the same way as this one does, and it has an 'index' parameter, but sure, if you wish to rename it, let's do it for both.

maimonav: Matan is the RENAMER, take the lead @mharsat. :)

oded-soffrin: let's keep it this way for now, API is already in use and changing it is breaking, will have to communicate w/ blog

oded-soffrin: as far as docs:
className: Content blocked by a payall will be marked up using this string
(This class will be used by Googlebot to discern between paid and free content on a page level)

index: number of free content blocks

maimonav: Regarding docs, that's good.
Regarding keeping the exist naming, if we already changing the `seoMode` to `seoSettings`, isn't it a good opportunity to change it all at once?

docs/docs/ricos/ricos-api.mdx
Barackos: yes `getContent` is still untouched from last time we spoke about it - we need to change it into `getContent({ postId })`. It's an editor function only

docs/docs/ricos/ricos-api.mdx
oded-soffrin: Matan is defining a new API for ricos, so we can decide what ever we want
if we want to change seoMode in RCE it's a breaking change, so we need to checkup with consumers (or release a v8 version). same with the server service. that's more time consuming and not part of the scope of what Matan is currently doing

oded-soffrin: for example - [this](https://github.com/wix-private/app-market/blob/master/communities/communities-blog-client-common/src/rich-content/components/rich-content-viewer/rich-content-viewer.js#L317) will break

mharsat: that won't break that's a RCV

oded-soffrin: it will if we change it all at once

PR #1375, opened 2020-07-21
[ricos] fix(themeStrategy): external parent class

Barackos:
Theme fix for multiple instances of RicosEditor / RicosViewer
New field for `theme` prop: `parentClass`

### DIFF
--- a/docs/docs/ricos/ricos-api.mdx
+++ b/docs/docs/ricos/ricos-api.mdx
@@ -106,12 +106,26 @@ For rendering a viewer, import the plugin from the viewer entry point <br/> `wix
 ```jsx
 theme?: {
   palette?: Palette | PalettePreset
+  parentClass?: string;
 }
 ```
 
 #### `palette`
 
 Defines the palette of colors that will be used. More info in [Theming](theming)
+
+#### `parentClass`
+
+A parent css className that wraps a ricos instance.
+
+:::note
+This is a temporary solution for multiple instances of ricos inside TPA.
+You can ignore this field if you don't render multiple instances of ricos, or if you import package styles only once (which is the general case).
+:::
+
+Its purpose is to wrap palettes' classNames as if they were children of `parentClass`, in order to strenghten specifity of css & bypass duplicated style-imports.
+For example: Wix-TPA theme duplicates css files of `editor`/`editor-common`/`common` in each ricos component, and therefore overrides each one due to order of css.
+Defining a unique `parentClass` for each component overcomes this, as it creates unique style specifity for each instance.
 
 ## RicosEditor API
 
--- a/docs/docs/ricos/ricos-api.mdx
+++ b/docs/docs/ricos/ricos-api.mdx
@@ -117,15 +117,7 @@ Defines the palette of colors that will be used. More info in [Theming](theming)
 #### `parentClass`
 
 A parent css className that wraps a ricos instance.
-
-:::note
-This is a temporary solution for multiple instances of ricos inside TPA.
-You can ignore this field if you don't render multiple instances of ricos, or if you import package styles only once (which is the general case).
-:::
-
-Its purpose is to wrap palettes' classNames as if they were children of `parentClass`, in order to strenghten specifity of css & bypass duplicated style-imports.
-For example: Wix-TPA theme duplicates css files of `editor`/`editor-common`/`common` in each ricos component, and therefore overrides each one due to order of css.
-Defining a unique `parentClass` for each component overcomes this, as it creates unique style specifity for each instance.
+With this defined, dynamic styles will be wrapped by `parentClass` for CSS specifity.
 
 ## RicosEditor API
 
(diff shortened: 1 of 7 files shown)


packages/ricos-common/web/src/RicosTypes.ts
```diff
@@ -65,6 +65,7 @@ export interface RicosViewerProps extends RicosProps {
 
 export interface RicosTheme {
   palette?: Palette | PalettePreset;
+  parentClass?: string;
```
mharsat, line 68: Can this be part of a `Palette` object?
Asking bc the Theme should be preferably only things like - palette, typography, etc., and it seems like this is only relevant to the palette

packages/ricos-common/web/src/themeStrategy/themeStrategy.ts
```diff
@@ -15,8 +16,10 @@ interface ThemeState {
 interface ThemeStrategyArgs {
   isViewer: boolean;
   themeGeneratorFunctions?: ThemeGeneratorFunction[];
+  theme?: RicosTheme;
   palette?: Palette | PalettePreset;
   cssOverride?: RicosCssOverride;
+  parentClass?: string;
```
mharsat, line 22: palette & parentClass can be removed

packages/ricos-common/web/src/themeStrategy/themeStrategy.ts
```diff
@@ -26,25 +29,37 @@ interface ThemeStrategyResult {
 
 export type ThemeStrategyFunction = (args: ThemeStrategyArgs) => ThemeStrategyResult;
 
+const addParentClass = (rawCss: string, parentClass: string): string =>
+  rawCss
+    .split('\n')
+    .map(line => (line.startsWith('.') ? `.${parentClass} ${line}` : line))
+    .join('\n');
```
mharsat, line 36: This probably has you covered, but have in mind that a class selector can also be defined in the long version `[class='className']`

packages/ricos-common/web/src/RicosTypes.ts
Barackos: Technically it can, but palette could also be a string (`PalettePreset`) and it should work when this type is given as well

packages/ricos-common/web/src/themeStrategy/themeStrategy.ts
Barackos: Good one. Yea this function is safe because the format of JSS's `toString()` output (`rawCss` param) is never printed in such format

mharsat: cool

docs/docs/ricos/ricos-api.mdx
```diff
@@ -106,13 +106,27 @@ For rendering a viewer, import the plugin from the viewer entry point <br/> `wix
...
 #### `palette`
 
 Defines the palette of colors that will be used. More info in [Theming](theming)
 
+#### `parentClass`
+
+A parent css className that wraps a ricos instance.
+
+:::note
+This is a temporary solution for multiple instances of ricos inside TPA.
```
oded-soffrin, line 122: this is the OS documentation, TPA or WIX is irrelevant here.
the only thing we should mention here is that it give the option to wrap all dynamic styles in a parentClass for CSS specifity (the rest should go in @wix/ricos docs)

examples/storybook/stories/Components/EditorWrapper.js
```diff
@@ -163,7 +163,7 @@ class EditorWrapper extends React.Component {
       <RicosEditor
         ref={ref => (this.editor = ref)}
         plugins={this.editorPlugins}
-        theme={{ palette }}
+        theme={{ palette, parentClass: 'stories-Components-styles__rce-wrapper__1ks2F' }}
```
oded-soffrin, line 166: I think we can make a specific story to display and test it, we don't need all the stories to use it (since a lot of time it's where we go to checkout the usage and it might be confusing)

packages/ricos-common/web/src/themeStrategy/themeStrategy.spec.ts
```diff
@@ -32,9 +40,25 @@ describe('ThemeStrategy', () => {
 
   it('should set inner props to override the default theme', () => {
     const cssOverride: RicosCssOverride = { modalTheme: { content: { backgroundColor: 'white' } } };
-    const themeStrategyResult = driver.runStrategy(undefined, undefined, cssOverride);
+    const themeStrategyResult = driver.runStrategy(undefined, undefined, undefined, cssOverride);
```
oded-soffrin, line 43: a thought - how about to change runStrategy function to 
 runStrategy =  ({themeGeneratorFunctions, palette, parentClass, cssOverride}) => {...

so we could call driver.runStrategy({cssOverride}) without all the **undefined**s

packages/ricos-common/web/src/themeStrategy/themeStrategy.ts
```diff
@@ -26,25 +27,37 @@ interface ThemeStrategyResult {
 
 export type ThemeStrategyFunction = (args: ThemeStrategyArgs) => ThemeStrategyResult;
 
+const addParentClass = (rawCss: string, parentClass: string): string =>
+  rawCss
+    .split('\n')
+    .map(line => (line.startsWith('.') ? `.${parentClass} ${line}` : line))
```
oded-soffrin, line 33: nice!

PR #1380, opened 2020-07-22
[fullscreen] hotfix(images): onExpand url

Barackos:


packages/plugin-gallery/web/src/lib/baseUrlConverter.ts
```diff
@@ -8,5 +8,10 @@ export const isAbsoluteUrl = (url: string) =>
 
 export const getAbsoluteUrl = (url: string, type: 'image' | 'video') => {
   const baseUrl = type === 'image' ? WIX_MEDIA_BASE_URL : WIX_VIDEO_BASE_URL;
-  return !isAbsoluteUrl(url) && !url.startsWith(baseUrl) ? baseUrl + url : url;
+  const prefix = type === 'image' ? 'media/' : 'video/';
+  if (!isAbsoluteUrl(url) && !url.startsWith(baseUrl)) {
```
oded-soffrin, line 12: this is a really confusing conditions
if url doesn't start with http:// or https://, it's definitely doesn't start with https://*.wixstatic.com...
(if a word doesn't start with A, it doesn't start with ABCDE)

so (!isAbsoluteUrl(url)) is enough

packages/plugin-gallery/web/src/lib/baseUrlConverter.ts
```diff
@@ -8,5 +8,10 @@ export const isAbsoluteUrl = (url: string) =>
...
+    const postfix = url.startsWith(prefix) ? url : prefix + url;
```
oded-soffrin, line 13: I think you mean suffix ? :)

oded-soffrin: maybe it's better to split into video / image so it will be more clear, since the logic is a bit hard to follow

if (isAbsoluteUrl(url)) {
   return url;
}

let baseUrl, suffix;
if (type === 'image') {
    baseUrl = WIX_MEDIA_BASE_URL;
    baseFolder = 'media/'
} else { //video
    baseUrl = WIX_VIDEO_BASE_URL;
    baseFolder = 'video/'
}

const pathWithFolder = url.startsWith(baseFolder) ? url : baseFolder + url;
return baseUrl + pathWithFolder;

jurasm:
Maybe let's add test for this helper. And let's add e2e test for expand mode.

PR #1382, opened 2020-07-23
feat: dynamic position for plugin toolbar on mobile

maimonav:


maimonav:
@noamtwix Please take a look, I changed the behaviour of the plugin toolbar on mobile, I saw that you worked on it.

PR #1384, opened 2020-07-23
[ricos] fix(plugin-mentions): type difference

Barackos:


mharsat:
@Barackos is we need to choose only one, why go for `EXTERNAL_MENTIONS_TYPE` and not `MENTIONS_TYPE`?

Barackos:
> @Barackos if we need to choose only one, why go for `EXTERNAL_MENTIONS_TYPE` and not `MENTIONS_TYPE`?

@mharsat backwards-compatibility. I'd go for `MENTIONS_TYPE` if it could work with old content as is, but to support that we have to make conversions (most likely update the normalization util, which would also influence RCE consumers).
In this PR I matched Ricos to RCE/RCV behaviour, so I think it's healthy separate this into a different PR (there's also EXTERNAL_LINK_TYPE in RCE that can be removed).

mharsat:
> > @Barackos if we need to choose only one, why go for `EXTERNAL_MENTIONS_TYPE` and not `MENTIONS_TYPE`?
> 
> @mharsat backwards-compatibility. I'd go for `MENTIONS_TYPE` if it could work with old content as is, but to support that we have to make conversions (most likely update the normalization util, which would also influence RCE consumers).
> In this PR I matched Ricos to RCE/RCV behaviour, so I think it's healthy separate this into a different PR (there's also EXTERNAL_LINK_TYPE in RCE that can be removed).

Can't we just accept both options for backward compatibility [here](https://github.com/wix-incubator/rich-content/blob/9a1754a5655ec5e9ff53692fd9d8af8d95e57300/packages/plugin-mentions/web/src/createMentionsPlugin.js#L32)?

Barackos:
> > > @Barackos if we need to choose only one, why go for `EXTERNAL_MENTIONS_TYPE` and not `MENTIONS_TYPE`?
> > 
> > 
> > @mharsat backwards-compatibility. I'd go for `MENTIONS_TYPE` if it could work with old content as is, but to support that we have to make conversions (most likely update the normalization util, which would also influence RCE consumers).
> > In this PR I matched Ricos to RCE/RCV behaviour, so I think it's healthy separate this into a different PR (there's also EXTERNAL_LINK_TYPE in RCE that can be removed).
> 
> Can't we just accept both options for backward compatibility [here](https://github.com/wix-incubator/rich-content/blob/9a1754a5655ec5e9ff53692fd9d8af8d95e57300/packages/plugin-mentions/web/src/createMentionsPlugin.js#L32)?

@mharsat Actually you're right, those types exist only for the config extraction. Pushed a fix that is also backwards-compatible.

PR #1389, opened 2020-07-26
[ricos-common] fix: translation using remountKey

mharsat:


mharsat:
Solved in new branch #1391

PR #1393, opened 2020-07-27
[plugin-link] fix: position of anchor dropdown filter

noamtwix:


shaulgo:
fullscreen test: maybe try to open the first image in the viewer.

PR #1409, opened 2020-07-30
[plugin-link] fix: link viewer onClick function (from settings)

noamtwix:


shaulgo:
add test for clicking on scroll

PR #1414, opened 2020-08-02
[plugin-link] fix: anchor on small (nom-mobile) screens

noamtwix:


packages/editor-common/web/statics/styles/multi-select-link-panel.scss
```diff
@@ -104,90 +148,107 @@
...
+    border-bottom: 2px solid #0261ff;
+    color: #0261ff;
   }
+}
 
-  .multiSelectLinkPanel_tabsWrapper{
-    display: flex;
-    border-bottom: 1px solid gainsboro;
-    margin-bottom: 12px;
+// @media only screen and (max-width: 640px) {
```
shaulgo, line 174: do we need these comments?

noamtwix: oops, my bad

PR #1417, opened 2020-08-03
Ricos media events

mharsat:


packages/ricos-editor/web/src/utils/hasActiveUploads.spec.ts
```diff
@@ -0,0 +1,101 @@
...
+const createEmptyContentState = () => ContentState.createFromText('');
+
+const createContentState = (entityType: PluginType, entityData) => {
+  const content = createEmptyContentState().createEntity(entityType, 'IMMUTABLE', entityData);
+  return AtomicBlockUtils.insertAtomicBlock(
+    createWithContent(content),
+    content.getLastCreatedEntityKey(),
+    ' '
+  ).getCurrentContent();
+};
```
oded-soffrin, line 21: super useful - not only for ricos-editor unit tests
can we move it to editor-common test utils and use @wix/draft-js instead of direct import from draft-js ?

packages/ricos-editor/web/src/utils/hasActiveUploads.ts
```diff
@@ -0,0 +1,45 @@
+import { EntityInstance, ContentState, ContentBlock } from 'draft-js';
```
oded-soffrin, line 1: same, this whole file can sit in editor / editor-common and draft-js import should be @wix/draft

examples/storybook/stories/Ricos/RicosMediaStory.js
```diff
@@ -0,0 +1,53 @@
...
+    accept: '*',
+    onFileSelected: mockFileNativeUploadFunc,
+  }),
+  pluginGallery(),
+];
+
+export default () => {
+  const isMobile = mobileDetect.mobile() !== null;
+  const [isButtonDisabled, setButtonDisabled] = useState(false);
+  const onBusyChange = isUploading => setButtonDisabled(isUploading);
```
oded-soffrin, line 34: Suggestion - it took me a while to understand what this story is about when opening on surge (no text explanations, and busy indication is only on publish button hover)
I would add on-screen log for event display   and perhaps make a longer timeout for for image upload / video upload

packages/ricos-editor/web/src/utils/hasActiveUploads.spec.ts
mharsat: Yes we should definitely use @wix/draft-js here as we import values and not types.
But it should probably stay here next to the hasActiveUploads.ts file, and import whatever it needs from editor-common, right?

packages/ricos-editor/web/src/utils/hasActiveUploads.ts
mharsat: In this case we import types, so it doesn't really matter bc it will always import them from @types/draft-js. It can be moved to editor-common, though it is only used in ricos layer

examples/storybook/stories/Ricos/RicosMediaStory.js
mharsat: Sounds good, what do you mean by on-screen log?

packages/ricos-editor/web/src/utils/hasActiveUploads.spec.ts
mharsat: this relies on `createWithContent` from `wix-rich-content-editor` which can't be a dep of editor-common. leaving this here for now

oded-soffrin: ok

PR #1418, opened 2020-08-03
feat/error toasts

talevy17:


packages/editor-common/web/src/Components/Toast.jsx
```diff
@@ -0,0 +1,107 @@
...
+export default class Toast extends Component {
+  constructor(props) {
+    super(props);
+    this.state = { message: '', isOpen: false };
+  }
+
+  componentDidMount() {
+    // avoid rendering a portal when running headless
+    this.body = document.querySelector('body');
+    ReactModal.setAppElement('body');
```
shaulgo, line 18: try portal

shaulgo (changes requested):

packages/editor-common/web/src/Components/Toast.jsx
talevy17: works great (:

packages/editor/web/src/RichContentEditor/RichContentEditor.jsx
```diff
@@ -648,6 +653,7 @@ class RichContentEditor extends Component {
                   {this.renderEditor()}
                   {this.renderToolbars()}
                   {this.renderInlineModals()}
+                  {this.renderErrorToast(locale)}
```
shaulgo, line 656: 1. try to use position fixed instead of portal.
2. if works, it will also not need the locale.

shaulgo (changes requested):

packages/plugin-image/web/src/image-component.jsx
```diff
@@ -86,7 +90,7 @@ class ImageComponent extends React.Component {
   }
 
   handleFilesAdded = ({ data, error }) => {
-    const imageData = data.length ? data[0] : data;
+    const imageData = data?.length ? data[0] : data;
```
shaulgo, line 93: Is data sometimes undefined?

shaulgo (changes requested):

packages/plugin-image/web/src/image-component.jsx
talevy17: I don't think it supposed to, but it does depend on the consumer
in storybook one of the error examples didn't pass data on error and it crashed

packages/editor-common/web/src/Components/ErrorToast.jsx
```diff
@@ -0,0 +1,80 @@
...
+export default class ErrorToast extends Component {
+  constructor(props) {
+    super(props);
+    this.state = { error: {}, errorCount: 0 };
+  }
+
+  static contextType = GlobalContext;
+
+  componentDidMount() {
+    const { commonPubsub } = this.props;
```
shaulgo, line 25: window.__internaltRicosOnError__ = this.onError

shaulgo (approved):

PR #1419, opened 2020-08-03
[preview] fix(seeFullPost & imageCounter): display issues

Barackos:


packages/preview/web/src/Components/ImageCounter.jsx
```diff
@@ -49,7 +49,7 @@ class ImageCounter extends PureComponent {
 
   decorateImages = () => {
     if (this.wrapper) {
-      const images = this.wrapper.querySelectorAll('[role=img]');
+      const images = this.wrapper.querySelectorAll('[role=link]');
```
Barackos, line 52: @lxgreen For some reason, image counter stopped working. It looked for `role=img` but each image has changed into `link` someday, and I couldn't find when it happened & where. Have any idea?

lxgreen: You mean images rendered by the image viewer component?

lxgreen: The role selector was required as we also decorate the gallery. We can't modify its code, so we had to work externally. They render canvas elements with role=img

lxgreen: We may need to talk to the gallery devs so they allow us to decorate the gallery items in official way so they won't break us

e2e/test-env/src/server/renderer.jsx
```diff
@@ -9,7 +9,7 @@ import RicosTestApp from '../client/RicosTestApp';
 
 export default function renderer() {
   return (req, res) => {
-    const [componentId, fixtureName = 'empty'] = req.path.replace(/^\/|\/$/g, '').split('/');
+    const [componentId, fixtureName = 'empty'] = req.path.replace(/^\/|\/$/g, '').split(/\/(.+)/);
```
Barackos, line 12: Split only first occurance of '/' so fixtures could be stored in folders as well

packages/preview/web/src/Components/ImageCounter.jsx
Barackos: Felt like a good time to add e2e till I work this out with gallery devs, so at least we would know when it breaks

Barackos: Discovered [here](https://pro-gallery.surge.sh/?itemClick=link&groupTypes=1,2h,2v,3t,3b,3l,3r) that `role="link"` comes from the `itemClick="link"` style prop for pro-gallery that we provide [here](https://github.com/wix-incubator/rich-content/blob/master/packages/preview/web/src/Components/default-transformation.js#L39), so it feels safe for now

lxgreen: So we are safe

PR #1424, opened 2020-08-04
[video] revise validate input url

oded-soffrin:


packages/common/web/src/Utils/urlValidators.js
```diff
@@ -1,6 +1,18 @@
 import linkifyIt from 'linkify-it';
 const linkify = linkifyIt();
 
+const UrlPattern = new RegExp(
+  '^(https?:\\/\\/)?' + // protocol
```
shaulgo, line 5: you don't want to use linkify.test

oded-soffrin: Linkify By default understands:

http(s)://... , ftp://..., mailto:... & //... links
"fuzzy" links and emails (google.com, foo@bar.com).

we don't need all that here. we can use it and understand its API but not sure what's the benefit

shaulgo: ya I thought about it afterwards. Maybe we should remove it completely. I just like one implementation. Definitely not urgent :)

PR #1425, opened 2020-08-04
[preview] fix(onPreviewExpand)

Barackos:
Allows consumers to provide an `onPreviewExpand` callback to determine whether the preview was expanded into full-view mode.


Barackos:
> API change ===> update docs

@mharsat Yup documented already, fixed title - it's a fix and not a feature

PR #1443, opened 2020-08-12
Modal footer buttons

danarish4:


packages/editor-common/web/src/Modals/UrlInputModal.js
```diff
@@ -12,8 +12,24 @@ export default class UrlInputModal extends Component {
   constructor(props) {
     super(props);
     this.state = {};
-    const { theme = {} } = props;
+    const { theme = {}, isVerticalEmbed } = props;
```
lxgreen, line 15: `isVerticalEmbed` is meaningless in `UrlInputModal` context, also it couples the `UrlInputModal` and `vertical-embed`. Consider prop like `buttonAlignment` accepting `center` (default) or `end`

bundle-analyzer/bundlesSizesBaseline.json
```diff
@@ -1,39 +1,40 @@
 {
```
lxgreen, line 1: not sure this should be committed

packages/editor-common/web/src/Components/SettingsPanelFooter.jsx
```diff
@@ -12,30 +12,55 @@ class SettingsPanelFooter extends Component {
...
+    const {
+      save,
+      cancel,
+      theme,
+      cancelLabel,
+      saveLabel,
+      fixed,
+      className,
+      t,
+      isModal,
```
lxgreen, line 24: maybe it is worth to pack these two props in a container object like `layoutOptions`? this can grow

PR #1452, opened 2020-08-15
Feat/debugging info

lxgreen:
### :rocket: New Feature
- common
  - debugging info reported to console if `ricosDebug=true` query param found


packages/common/web/lib/debugging-info.js
```diff
@@ -0,0 +1,52 @@
+function formatPluginInfo(plugins) {
+  return plugins.reduce((result, plugin) => result + `\t${plugin}\n`, '');
+}
+
+export function reportDebuggingInfo({ version, plugins, getContent, getConfig, reporter }) {
+  try {
+    if (typeof window === 'undefined') {
+      return;
+    }
+    const ricosDebug = new URL(window.location.href).searchParams.get('ricosDebug');
```
shaulgo, line 10: I believe this will accept any capitalization and doesn't need the true.
`/ricosDebug/i.test(location.search)`
When I used to use the santa debug param. There was one place that needed 'all' another one needed 'true' and whatever else... I would always forget.

PR #1480, opened 2020-08-25
[inner-rce] fix: change editor state from plugin + fix link toolbar s…

noamtwix:
…tay open

### DIFF
--- a/packages/editor/web/src/RichContentEditor/InnerRCE.jsx
+++ b/packages/editor/web/src/RichContentEditor/InnerRCE.jsx
@@ -24,11 +24,19 @@ class InnerRCE extends Component {
     this.setState({ MobileToolbar, TextToolbar });
   }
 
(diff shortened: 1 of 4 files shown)


packages/editor/web/src/RichContentEditor/InnerRCE.jsx
```diff
@@ -24,11 +24,19 @@ class InnerRCE extends Component {
     this.setState({ MobileToolbar, TextToolbar });
   }
 
+  static getDerivedStateFromProps(props, state) {
+    return !state?.editorState?.getSelection?.().hasFocus ? { editorState: props.editorState } : {};
```
shaulgo, line 28: Are you sure they are needed. I think you always have editorState?
```suggestion
    return !state.editorState.getSelection.().hasFocus ? { editorState: props.editorState } : {};
```

shaulgo (approved):

PR #1498, opened 2020-09-01
Feat/plugin commons

lxgreen:


bundle-analyzer/analyzeBundles.ts
```diff
@@ -42,7 +42,11 @@ const getAllPluginsNames = ({
   return getPackages().then(allPackages => {
     return allPackages
       .filter(pkg => !pkg.private)
-      .filter(pkg => pkg.name.indexOf('wix-rich-content-plugin') === 0)
+      .filter(
+        pkg =>
+          pkg.name.indexOf('wix-rich-content-plugin') === 0 &&
+          pkg.name !== 'wix-rich-content-plugin-commons'
```
oded-soffrin, line 48: oh crap sorry about that

lxgreen: hmm i think we can add some metadata to package.json, e.g. `type: 'rich-content-plugin'` or `type: 'rich-content-infra'` for better handling stuff like this

oded-soffrin: I actually think it might be better to stop createing so many bundles for testing bundle size, but generate a few with subset of plugins, it take some time for our build process

lxgreen: gonna try perfer some time...

PR #1505, opened 2020-09-02
chore(*): make sure semicolon appears in rtl directives

oded-soffrin:


oded-soffrin:
fix to problem described here: https://wix.slack.com/archives/C8QHV6UM9/p1598970401008200

PR #1508, opened 2020-09-02
Fix child props never updated

mharsat:


Barackos (approved):

oded-soffrin:
This is solving native issue w/ editorState ?

mharsat:
> This is solving native issue w/ editorState ?

Yes, the more general issue was that child props were fixated in construction phase, so in particular, updating the `editorState` prop never reached the RCE component/draft

oded-soffrin:
Cool, please update noam after we release, their waiting for that

PR #1509, opened 2020-09-03
FIX: change url in html plugin test

ShamaySapir:


oded-soffrin (changes requested):
Changing to invalid URL is confusing, and we can't be sure HTML works if nothing is presented
You can use https://www.w3schools.com/

oded-soffrin (approved):

PR #1520, opened 2020-09-06
[editor] fix: inner modal theme

noamtwix:


oded-soffrin:
@lxgreen is this a case of what you've discussed where we use `theme` in our own styles ?

lxgreen:
@oded-soffrin nope, this one look fine. the bad case is to pass a package A styles via theme prop to a component imported from package B. in that case, we cannot guarantee that the theme will override the default style

oded-soffrin (approved):

oded-soffrin:
@shaulgo looks good to me, I'm not sure about innerModal structure, but I don't want to block it for no reason

shaulgo (approved):

PR #1525, opened 2020-09-06
[*] improve(theme): typed theme functions for all plugins

Barackos:


CHANGELOG.md
```diff
@@ -32,6 +32,8 @@
   - [#1509](https://github.com/wix-incubator/rich-content/pull/1509) test html plugin change url
 - `ricos-theme`
   - [#1515](https://github.com/wix-incubator/rich-content/pull/1515) bundle size improved by removing unused JSS plugins
+- `*`
```
mharsat, line 35: `general` can be used as a human language alternative

mharsat (approved):
Nice infra work!
I only think `THEME()` should be lowercase since it's a function and not a constant value. Especially since it is always used as `import { THEME as theme }`. Anyway, that's a matter of personal preference

PR #1527, opened 2020-09-07
ricos-content

oded-soffrin:
This package was create to seperate normalization logic.
OneApp have a few policy that prevent them from using our normalization:

- upon consuming 'ricos-common' they also get 'image-client-api' which is not related to normalization, and breaks in WebView environment (window dependency)
- They can't deep export 'common/dist/lib...' due to OneApp Police

So the thought is to create a library that is platform agnostic, that will contain ricos content manipulation, validation and creation.

Would love to discuss about naming, and how to build so when we move content state validator it won't bloat ricos-common who depend on it


packages/ricos-content/web/package.json
```diff
@@ -0,0 +1,79 @@
...
+    "cross-env": "^5.2.0",    
+    "jest": "^26.0.1",
+    "lodash": "^4.17.15",    
+    "rollup": "1.31.1",
+    "ts-is-present": "^1.1.1",
+    "ts-jest": "^25.5.0"
+  },
+  "dependencies": {    
+    "ricos-common": "7.18.3",
+    "wix-rich-content-common": "7.18.3"
```
mharsat, line 54: These create an interdependency and are not used

packages/common/web/src/types/index.ts
```diff
@@ -1,6 +1,7 @@
+export { ComponentData, RicosContentBlock, RicosContent, RicosEntityRange } from 'ricos-content';
```
mharsat, line 1: This excludes some types that were exported from common, and is not necessary bc of `export * from 'ricos-content'` in main index.ts file

packages/ricos-content/web/tests/palettesExample.ts
```diff
@@ -0,0 +1,234 @@
...
+  palette.map((color, i) => ({ ...TemplateExample[i], value: color }));
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+const wixPalettes: any = {};
+// eslint-disable-next-line array-callback-return
+Object.keys(examples).map(palette => {
+  wixPalettes[palette] = paletteToWixPalette(examples[palette]);
+});
+
+export { wixPalettes };
```
mharsat, line 234: Why is this file in ricos-content? Seems like a duplicate of both ricos-theme & storybook

mharsat (changes requested):
Looks like a good idea to separate some logic out of the common package.

This seems like the right place to also consolidate all draft-js manipulation utils we let consumers use, only blocker for that is that this is a viewer dependency.

As to `image-client-api`, this is a Wix dependency and should eventually be totally removed from this repo, once we decouple the image plugin from Wix.

Anything blocking us from moving the contentStateValidator and all schemas to here too?

packages/common/web/src/types/index.ts
oded-soffrin: these are all the types that are used outside of that package, I guess we can export all of them but they're just not needed

mharsat (approved):

shaulgo:
do we know why the bundles grew?

oded-soffrin:
Nope. I made this change before plugin-common merge, it might be related. it didn't grow before I've merged from master today...

oded-soffrin:
@shaulgo it was due to lib bundles this new package separately. 
was added to externals, @mharsat is looking this week on our road toward tree shaking to avoid these cases in the future

PR #1529, opened 2020-09-07
[workflows] improve(actions): ignores

Barackos:
Added [paths & paths-ignore](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#example-ignoring-paths) for more efficient CI process.

Example: no need to run tests if only `CHANGELOG.md` was changed.

@shaulgo @mharsat @lxgreen What do you think?

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -5,6 +5,11 @@ on:
       - '**'
     branches-ignore:
       - 'release*'
+    paths-ignore:
+      - '**/*.md'
+      - 'docs/**'
+      - 'examples/**'
+      - 'bundle-analyzer/**'
 env:
   captureVideos: 'true'
   captureImages: 'true'
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -5,6 +5,10 @@ on:
       - '**'
     branches:
       - '**'
+    paths-ignore:
+      - '**/*.md'
+      - 'docs/**'
+      - 'bundle-analyzer/**'
 jobs:
   deploy-examples:
     runs-on: ubuntu-18.04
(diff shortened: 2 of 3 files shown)


Barackos:
Never mind actually... It seems that a [required-if-run status checks](https://github.community/t/status-checks-required-if-run/117330)' mechanism isn't supported yet, and we need to leave e2e tests as required.

.github/workflows/tests.yml
```diff
@@ -5,6 +5,11 @@ on:
       - '**'
     branches-ignore:
       - 'release*'
+    paths-ignore:
+      - '**/*.md'
+      - 'docs/**'
+      - 'examples/**'
+      - 'bundle-analyzer/**'
```
lxgreen, line 12: Hmm.. yesterday I fixed my bug in the bundle analyzer code, and I definitely wanted to run the CI

lxgreen (commented):
Consider removing the bundle analyzer from ignore path. Besides that, great idea!

.github/workflows/tests.yml
Barackos: Indeed - notice that `tests.yml` is responsible for the `e2e` tests, while I assume you only wanted `analyzeBundles` to run, which is triggered in [bundleSize.yml](https://github.com/wix-incubator/rich-content/blob/master/.github/workflows/bundleSize.yml)

.github/workflows/tests.yml
```diff
@@ -5,6 +5,11 @@ on:
       - '**'
     branches-ignore:
       - 'release*'
+    paths-ignore:
+      - '**/*.md'
+      - 'docs/**'
+      - 'examples/**'
```
mharsat, line 11: test-env depends on examples, like [here](https://github.com/wix-incubator/rich-content/blob/f1e1a0cdb89ab30bb67fd33bef2c3849b37b0112/e2e/test-env/src/client/RicosTestApp/RicosTestApp.jsx#L11)

mharsat (commented):
Good idea 💡

.github/workflows/tests.yml
Barackos: Oops you're right

PR #1535, opened 2020-09-08
 [editor] feat(contentEvents): onPluginAddSuccess

Barackos:
A new callback, `onPluginAddSuccess`, is now added to helpers API, in addition to the already existing `onPluginAdd`.
This one is triggered only after a successful plugin addition, whereas `onPluginAdd` is triggered upon a click on `insertPluginButton` click (from any toolbar).

Moreover, fixed `onPluginAdd` to be triggered appropriately.


mharsat (commented):
> is now added to helpers API

I hear that and get shivers, helpers API should be on its way to full deprecation. How about we start passing our BI callbacks to child component via a dedicated entry point? The `BICallbacks` type can be moved from wix-ricos to rich-content, allowing consistency between this prop's type and the biCallbacks we have in wix-ricos

As to onPluginAddSuccess, how do we measure success of plugin addition? Is changing the editor state enough, or should this be confirmed by the plugin itself? (is an image plugin considered successfully added when the block appears, or when the upload succeeds? Until then the block is in a pending state)
What happens when some other developer adds a new util for adding blocks, and doesn't call the callback? Might need a solution that's more resistant to changes.

Barackos:
@mharsat 
1. I agree with you about the pain of growing the to-be-deprecated prop `helpers`. However, I think your suggestion should be implemented in 1 stage, meaning all `BICallbacks` should move to another entry at once, and not partially. Therefore I methodically prefer to split tasks & to do it in another PR, as such separation has its own capacity.

2. Regarding "Success" - you raised an interesting point: We currently define success to be the completion of the user's necessary steps to add a plugin to the content (a.k.a new block). So a successful upload was excluded from this definition, but I'll ask Emanuel for his thoughts - we can either redefine the purpose, or simply create a new event - `onUploadSuccess`).

3. Regarding updates via different utils - true. I'll make a workaround for external toolbars. For other case scenarios, I'll have to raise awareness among devs unfortunately. The simple solution it to add this event to the content change so we won't miss any plugin addition, but in this way we cannot tell from which toolbar a plugin was added, and it's a necessary parameter.

Anyhow, I'll share this dilemma with Emanuel and update accordingly.

mharsat:
@Barackos 

1. Definitely doesn't have to be in this PR
2. For media plugins success could be upload completed, but try to think of other plugins too. Are vertical embed or link preview only considered as valid in the content state once they fetch the data they need?
3. Maybe you can extract functionality such that we have a single `addBlock` utility (might be internally separated to cases) such that a block is added from a single function in which `onPluginAddSuccess` is called only once. Then anyone adding a new method will call that `addBlock` function.

PR #1547, opened 2020-09-13
Convert from HTML

oded-soffrin:
I tests the draft-html-import for FAQ to see whether it's basically working
in storybook example (Test Cases > HTML convert) it fails with: Invariant Violation: block is not a BlockNode

I believe we need to support it from within so it will use @wix/draft-js instead of draft-js to maker it work

on top of that we need to see what's the behavior with images/video etc

### DIFF
--- a/examples/storybook/package.json
+++ b/examples/storybook/package.json
@@ -30,6 +30,7 @@
     "@testing-library/react": "^9.4.0",
     "babel-jest": "^25.1.0",
     "classnames": "^2.0.0",
+    "draft-js-import-html": "^1.4.1",
     "jest": "^26.0.1",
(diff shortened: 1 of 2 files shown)


lxgreen:
> Invariant Violation: block is not a BlockNode

this usually means there are multiple versions of draft

lxgreen:
> rich-content-storybook: [rich-content-storybook-draft-html-import.surge.sh](https://rich-content-storybook-draft-html-import.surge.sh/)

failed to deploy

oded-soffrin:
yes, double instance, because html-import uses draft-js.
if we'll support it, rollup alias will use @wix/draft-js and only one instance will be there.

I don't know why it failed to deploy, but this is only a POC starter.

PR #1552, opened 2020-09-15
Vertical Embed send locale to verticalsApi

ShamaySapir:


oded-soffrin (commented):
Looks good to me
@lxgreen adding 'locale' to CreatePluginToolbar make sense, right ?

oded-soffrin (commented):
should be matched with changes in MockVerticalSearchModule

lxgreen:
@oded-soffrin up till now we didn't need it there, 'cause the `t` alone was sufficient. Now we need this, so it's ok

oded-soffrin (approved):

PR #1559, opened 2020-09-17
Workspaces

shaulgo:


shaulgo:
i18next should be peer dependency?

shaulgo:
seems like jsonschema adds 6kb to common

shaulgo:
ts-is-present should be dependency. Why did it not break?

PR #1561, opened 2020-09-20
[ricos-theme] JSS to css-variables migration

Barackos:
### :rocket: New Feature
- New design implementation for the plugins toolbar - active background & hover background
- 3-Color Palette API, in addition to Wix-Palette format
  
### 🏠 Internal
- JSS deprecation, improved bundle size. replaced with CSS-Vars.

-----

3-Color Palette API:

```jsx
import { createTheme } from 'ricos-theme';
// ...
const palette = {
  bgColor: '#FFFFFF',
  textColor: '#414141',
  actionColor: '#FA6400',
};
// ...
<RicosEditor theme={createTheme({ palette })} />;
```

- Unit-Test covered
- Implemented in Storybook


packages/common/web/statics/styles/palette.scss
```diff
@@ -1,15 +1,27 @@
...
 $secondary-text-color: #757575;
 
 $modal-input-color: #333333;
-$modal-warning-color: #ee5951;
-$modal-btn-confirm-primary-color: white;
-$modal-btn-confirm-secondary-color: #0261ff;
 
 $toolbar-font-color: #333333;
 $toolbar-border-color: #ededed;
+
```
Barackos, line 11: Cannot remove all remaining `${colorName}`s, as they are still in use.
Should consider deprecating those as well in the wiring process with Nitzan.

packages/ricos-theme/web/src/ThemeGenerator.ts
```diff
@@ -20,6 +15,24 @@ const getColorByCode = (palette: Palette, code: number): Color => {
...
+    --ricos-action-color: ${actionColor};
+    --ricos-action-color-tuple: ${toRgbTuple(actionColor)};
+    --ricos-action-color-fallback: ${adaptForeground(actionColor)};
+    --ricos-action-color-fallback-tuple: ${toRgbTuple(adaptForeground(actionColor))};
+    --ricos-action-color-fallback-bright: ${fallbackColorBright};
+    --ricos-action-color-fallback-bright-tuple: ${toRgbTuple(fallbackColorBright)};
+    --ricos-background-color: ${bgColor};
+    --ricos-background-color-tuple: ${toRgbTuple(bgColor)};
+  }\n`;
+};
```
Barackos, line 34: This is the interesting part

packages/ricos-theme/web/src/themes/utils.ts
```diff
@@ -22,16 +22,50 @@ export function adaptForeground(actionColor: string): string {
...
+}
+
+/**
+ * Converts `hexColor` from HEX format to a CSS RGBA string
+ * @example
+ * toCssRgbA('#FFFFFF', 0.5) => 'rgba(255, 255, 255, 0.5)'
+ *
+ * @param hexColor color in HEX format
+ * @returns RGB object
+ */
```
Barackos, line 65: Experimenting with TypeScript documentation.
@mharsat We might want to add this to other type definitions in the documentation effort, as it prompts descriptions while typing functions / fields.

mharsat: Don't think that's specifically TS related, but that's very good to have!

packages/ricos-theme/web/src/ThemeGenerator.ts
```diff
@@ -52,20 +65,15 @@ export default class ThemeGenerator {
 
   getStylesObject() {
```
mharsat, line 66: Looks like this doesn't return an object anymore, maybe rename?

mharsat (commented):
The ricos-theme part looks great 🚀 Make sure we can confidently make all the style changes, check that what we need is covered in e2e

oded-soffrin:
Looks good to me!
especially this part - ricos-viewer-with-theme: 227KB => 199KB

packages/ricos-theme/web/src/ThemeGenerator.ts
Barackos: sounds good

packages/common/web/statics/styles/palette.scss
```diff
@@ -1,15 +1,27 @@
-$primary-color: #607d8b;
```
lxgreen, line 1: maybe it's too late to ask, but couldn't you just
```
$primary-color: var(--primary-color);
```
and then you don't have to replace it everywhere?

Barackos: You're right, fixed! Never too late with `Find & Replace` 😄

Barackos:
merged into v8 updates

PR #1577, opened 2020-09-29
[RCE] Return the result of handleBeforeInput callback when present

muteor:
We use the prop callback to apply char limiting to the editor, for this to work we need the handler to return the result when we set it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,10 @@
     Click to see more.
   </summary>
 
-
+  ### :home: Internal
+- `rce`
+  - [#1577](https://github.com/wix-incubator/rich-content/pull/1577) Return the result of handleBeforeInput callback when present
+  
   </details>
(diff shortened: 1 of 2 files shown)


shaulgo:
Hey, I missed this PR. You are more than welcome to ping me in slack if I'm ever not responsive. 
I think we should implement char limit on our side. 
1. Other consumers might want it too
2. We are going to start a refactor to use prosemirrore instead of draft this Q. (will take time to complete). So anything that is using editor state will break.

You can either make a pr for charLimit prop. Or we can open a request and let the PMs decide on the priority.

shaulgo (approved):

shaulgo:
I'll fix the conflicts and merge. Sorry for taking so long. I hoped to give another solution.

shaulgo:
handleBeforeInput can return 'handled' now

PR #1578, opened 2020-09-29
Toast/ fix wix ad covering error toasts

talevy17:


Borealdesign:
ANDROID crash, how to fix?
Unbelievable, Tor almost dead, a suffer an attack, lost Gb's of work, 1 Google account , I use Arch reverse engineering to save my Adroid. Now I know that I didn't suffer a hacker attack, the cause was this shit from hell. 
Thats worst, now I not sure I fixed.
I fell Jonh Snow, know nothing

PR #1579, opened 2020-09-29
[inner-rce] fix: create inner-rce contentState without version

maimonav:


packages/editor/web/lib/editorStateConversion.ts
```diff
@@ -47,10 +47,20 @@ const convertToRaw = ContentState =>
     version
   );
 
+const convertToRawWithoutVersion = ContentState =>
```
shaulgo, line 50: 1. rename to __convertToRawWithoutVersion
2. convertToRaw should use this function

shaulgo (changes requested):

packages/editor/web/lib/editorStateConversion.ts
maimonav: done.

PR #1580, opened 2020-09-30
[plugin-link] fix: link panel width in safari

noamtwix:


shaulgo (approved):
I'm not completely sure what you did here. But looks legit enough.

PR #1581, opened 2020-09-30
Removed tsconfig files in packages

mharsat:
- Removed all package tsconfig files, substituted by rollup-plugin-typescript2's `tsconfigOverride` config
- Fixed type errors in draftUtils & draftUtils.spec
- Removed `sourceMap` compiler option as it caused errors in cypress. It was never used bc we create source maps with rollup, not tsc


Barackos (approved):

rollup.plugins.ts
```diff
@@ -83,9 +83,26 @@ const babel = (): Plugin => {
 };
 
 const typescript = (): Plugin => {
+  const absPath = (dir: string) => `${process.cwd()}/${dir}`;
   return typescriptPlugin({
     useTsconfigDeclarationDir: true,
     check: !!process.env.GITHUB_ACTIONS,
+    tsconfig: `${__dirname}/tsconfig.json`,
+    tsconfigOverride: {
```
lxgreen, line 91: According to plugin [docs](https://www.npmjs.com/package/rollup-plugin-typescript2#plugin-options):
> For simplicity and other tools' sake, try to minimize usage of defaults and overrides and keep everything in tsconfig.json file (tsconfigs can themselves be chained, so save some turtles).

This PR removes some flexibility we might need, e.g. now it's impossible to apply strict type-checking gradually (noImplicitAny) to a specific package.

rollup.plugins.ts
```diff
@@ -83,9 +83,26 @@ const babel = (): Plugin => {
...
+      },
+      include: [
+        'src',
+        'src/**/*.json',
+        'statics/**/*.json',
+        'statics/**/*.schema.json',
+        'package.json',
+        'lib',
+      ].map(path => absPath(path)),
+      exclude: ['node_modules', '**/*.spec.*'].map(path => absPath(path)),
```
lxgreen, line 104: what's the reason to exclude test files if they are .ts?

rollup.plugins.ts
mharsat: The original reason for multiple tsconfig files was to use options with paths like `declarationDir`, `include` & `exclude`. This creates maintenance complexity, as those config files were merely duplicates, and most changes required refactoring all those files together. This caused errors, as sometimes they were not aligned (e.g. new plugin merged from branch after the refactor happened).
After core3 guys showed us their usage of a single tsconfig for monorepos, I was looking for a way overcome the path issues and implement this in our setting.
This means we don't have gradual changes, but

- I don't believe gradual changes are ever completed (enforcing `noImplicitAny` in the project would be a nice and uncomplicated task for any dev to take)
- I think having a single config file in a monorepo has more pros than cons (for the same reasons we wouldn't have an eslint config file in every package, for gradually adding linting rules)

Let's discuss more scenarios we can think of

rollup.plugins.ts
mharsat: We don't treat .ts and .js files differently as we use `allowJs`. Why would we want test files in our dist?

lxgreen: a unit test file (.ts) could also test the unit-under-test type definition, especially for public APIs. 
Consider the following case:
```
	// someExpensiveDataRetrieve fetches data unless it's cached
	// cache is available within function closure

	function someExpensiveDataRetrieve(id: string) {
		if(cache[id]) {
			return cache[id]; // data retured
		}
		return fetch(getUrlById(id))
			.then(response => response.json())
			.then(data => {
			 	cache[id] = data;
				return data; // Promise returned
			});
	}
```

In this case, one code path returns `data`, another returns a `Promise<data>`. Of course, it's better to return `Promise.resolve(data)` so it's always a Promise, but let's consider it's just bad API. 

So we rely on TS to infer the return type. The interesting part here is that the `someExpensiveDataRetrieve` is absolutely correct from the TS perspective, and we won't get any warnings. The consumer will get the error if tries to use this API:
```
	someExpensiveDataRetrieve(id).then(data => {...}) // `then` is not defined for the `data`
```

So, consider we have a unit test for `someExpensiveDataRetrieve`, and we test both paths. If we'd enable the typechecking, we could catch it in the test.

mharsat: Both from TS perspective & runtime perspective - if `cache[id]` exists then the return type is `typeof data`, meaning that `then` is not defined.
AFAIU TS is correct in compilation time when it predicts that the return type is `typeof data | Promise<typeof data>`
We can have a `someExpensiveDataRetrieve.spec.ts` and it will fail on
```jsx
someExpensiveDataRetrieve(id).then(data => {...})
```
because ts-jest runs the test file. But if we return to the topic of excluding spec files, creating a `someExpensiveDataRetrieve.spec.d.ts` declaration file won't help us, ts-jest will

lxgreen: if we have sick test cases like `someExpensiveRetrieve(newId).then(...)` and `data = someExpensiveRetrieve(oldId)`, the runtime is ok. 
this example is too artificial, but my concern is -- we can have a case when the code is type-wise correct, but will fail (type-wise) when consumed. if you're sure that ts-jest covers us, then let's skip them.
in addition, when i worked on `ricos-content` tsconfig, the TS behavior surprized me -- the type-check was reporting only the files that run. so, during the build only the `rollup.plugins.ts` errors were reported, and the vast majority of type errors were reported during the `npm test`

mharsat: Yes that's expected, let's discuss that

PR #1596, opened 2020-10-04
[accordion] feat: reduce accordion bundle size

maimonav:


packages/plugin-accordion/web/src/defaults.ts
```diff
@@ -1,5 +1,7 @@
-import { ArrowIcon } from './icons';
-import { EditorState, __convertToRawWithoutVersion } from 'wix-rich-content-editor';
+import {
```
shaulgo, line 1: revert

packages/plugin-accordion/web/src/defaults.ts
```diff
@@ -1,5 +1,7 @@
...
+  EditorState,
+  __convertToRawWithoutVersion,
```
shaulgo, line 3: don't load defaults in viewer

packages/plugin-accordion/web/src/defaults.ts
maimonav: done.

packages/plugin-accordion/web/src/defaults.ts
maimonav: done.

maimonav: It appears that default is indeed needed in viewer, consider this:
https://github.com/wix-incubator/rich-content/blob/feat%2FreduceAccordionBundleSize/packages/plugin-accordion/web/src/viewer.ts#L14

So I think it's best to give default value without using `EditorState` & `__convertToRawWithoutVersion` at all.

PR #1601, opened 2020-10-05
[gallery] fix(gallery-viewer): horizontal layouts index 

talevy17:


e2e/tests/media.e2e.js
```diff
@@ -206,9 +212,13 @@ describe('plugins', () => {
         cy.waitForGalleryImagesToLoad();
         cy.eyesCheckWindow(this.test.parent.title + ' - select all items');
         cy.get(`[data-hook=${GALLERY_SETTINGS.DESELECT}]`).click();
-        cy.dragAndDrop(firstImage, anyImage, 1);
         cy.waitForGalleryImagesToLoad();
         cy.eyesCheckWindow(this.test.parent.title + ' - deselect items');
+        // TODO: stabalize reordering tests
```
shaulgo, line 217: is this left here by purpose?

e2e/tests/media.e2e.js
```diff
@@ -189,7 +195,7 @@ describe('plugins', () => {
     context('organize media', () => {
       it('allow to manipulate the media items', function() {
         const firstImage = `[data-hook=${GALLERY_SETTINGS.IMAGE}]:first`;
-        const anyImage = `[data-hook=${GALLERY_SETTINGS.IMAGE}]`;
+        // const anyImage = `[data-hook=${GALLERY_SETTINGS.IMAGE}]`;
```
shaulgo, line 198: is this left here by purpose?

shaulgo (approved):

e2e/tests/media.e2e.js
talevy17: tl;dr yes.
it appears there was some mixup when we started using applitools - the drag and drop test and the items selection tests were intertwined and it caused some flakiness issues.
ideally we would want to test the drag and drop (reording) capability but it appears unstable (the images in the gallery might get captured blurry and our waitForImageToLoad mechanism doesn't seem to solve it when the focus is on the settings modal.
I'm trying to work it out on a different branch

e2e/tests/media.e2e.js
talevy17: same as the previous comment

PR #1602, opened 2020-10-05
feat/Quote2.0 - improved quote behaviour

maimonav:


e2e/tests/fixtures/multiple-accordions-rich-text.json
```diff
@@ -413,7 +413,7 @@
         },
         "pairs": [
           {
-            "key": 1,
+            "key": "1",
```
shaulgo, line 416: where are this keys made? It looks like a counter. I want to make sure there is not problem with it.

packages/viewer/web/src/utils/convertContentState.jsx
```diff
@@ -106,7 +113,7 @@ const getBlocks = (mergedStyles, textDirection, context, addAnchorsPrefix) => {
             style={blockDataToStyle(blockProps.data[i])}
             key={blockProps.keys[i]}
           >
-            {_child}
+            {<div className={textClassName}>{_child}</div>}
```
shaulgo, line 116: adding textClassName on the childTag doesn't work?

CHANGELOG.md
```diff
@@ -17,6 +17,8 @@
   </summary>
 
 ### :rocket: New Feature
+- `quote`
+  - [#1602](https://github.com/wix-incubator/rich-content/pull/1602) improved quote behaviour
```
shaulgo, line 21: Let's give a better description of what it does.

e2e/tests/fixtures/empty-accordion.json
```diff
@@ -46,7 +46,7 @@
         },
         "pairs": [
           {
-            "key": 20195,
+            "key": "20195",
```
shaulgo, line 49: We should take another look at this keys. It could be not very fun if we have duplicate keys.

packages/plugin-accordion/web/src/accordion-viewer.jsx
```diff
@@ -23,15 +29,17 @@ class AccordionViewer extends Component {
...
 
   renderContent = idx => {
     const { innerRCV, componentData } = this.props;
     const { pairs } = componentData;
     const contentState = pairs[idx].content;
     const textAlignment = getTextAlignment(componentData);
-    return innerRCV({ contentState, textAlignment });
+    const direction = getDirection(componentData);
+    return innerRCV({ contentState, textAlignment, direction });
   };
```
shaulgo, line 44: let's add 
1. getPair(id)
2. renderInnerRCV(contentState)

shaulgo (changes requested):

e2e/tests/fixtures/multiple-accordions-rich-text.json
maimonav: I wrote this fixture before we used the key generator, new accordions use the generator.

e2e/tests/fixtures/empty-accordion.json
maimonav: Pretty sure it's random enough, don't mind this fixture, since I wrote them before we used the key generator.

shaulgo:
just saying I want to take another look with you next time we talk. It
might already be good enough. We need to remember that in its lifetime It
could be used over a million times.

CHANGELOG.md
maimonav: done.

packages/plugin-accordion/web/src/accordion-viewer.jsx
maimonav: done.

packages/viewer/web/src/utils/convertContentState.jsx
maimonav: It might, but now it's the same DOM structure as the editor, so I thought this would be better.

shaulgo (approved):

PR #1622, opened 2020-10-09
[ricos-content] isEmpty util

oded-soffrin:
Adding `isEmpty` & basic `createContent` functionality to `ricos-content` package


packages/ricos-content/web/src/contentStateUtils/contentStateUtils.ts
```diff
@@ -0,0 +1,11 @@
+import { RicosContent } from '../types';
+
+export const isContentStateEmpty = (contentState: RicosContent): boolean => {
+  const { blocks } = contentState;
+  if (blocks.length !== 1) {
+    return true;
+  }
+  const firstBlock = blocks[0];
+  const { text, type } = firstBlock;
+  return text === '' && type === 'unstyled';
```
maimonav, line 10: Do you consider depth here? For example, if block is empty but has indentation (depth=1), if so, you should add check for depth===0.

packages/ricos-content/web/src/contentStateUtils/createContent.ts
```diff
@@ -0,0 +1,27 @@
...
+  | 'unstyled'
+  | 'blockquote'
+  | 'header-one'
+  | 'header-two'
+  | 'header-three'
+  | 'header-four'
+  | 'header-five'
+  | 'header-six';
+
+export const createContent = (text: string, type: BlockType = 'unstyled') => {
```
maimonav, line 11: I think it's best to move use Editor.createEmpty().getCurrentContent() or Editor.createWithContent(), BTW, I would extract editorStateConversion.ts from 'editor' package to here and add tests as well.

maimonav (approved):

packages/ricos-content/web/src/contentStateUtils/createContent.ts
oded-soffrin: You're right - however `ricos-content` don't contain 'draft-js' yet due to size issue. 
There are few editorStateConversion function which is worth grabbing though, that's smart.

packages/ricos-content/web/src/contentStateUtils/contentStateUtils.ts
oded-soffrin: good point! I'll add it

packages/ricos-content/web/src/contentStateUtils/createContent.ts
maimonav: Yeah I agree, we might be able to ask Bachman or anyone who is maintaining `wix-draft` to this few functions as lib, but not sure it's worth the effort.

packages/ricos-content/web/src/contentStateUtils/contentStateUtils.ts
```diff
@@ -0,0 +1,11 @@
+import { RicosContent } from '../types';
+
+export const isContentStateEmpty = (contentState: RicosContent): boolean => {
+  const { blocks } = contentState;
+  if (blocks.length !== 1) {
+    return true;
```
shaulgo, line 6: is this supposed to be false?

shaulgo: I also like `> 1` more. As it conveys the meaning. Also, currently blocks.length can never be 0. But it would take care if it as well.

packages/ricos-content/web/src/contentStateUtils/createContent.ts
```diff
@@ -0,0 +1,27 @@
...
+        text,
+        type,
+        depth: 0,
+        inlineStyleRanges: [],
+        entityRanges: [],
+        data: {},
+      },
+    ],
+    entityMap: {},
+    VERSION: '7.5.0',
```
shaulgo, line 25: maybe delete this.

packages/ricos-content/web/src/contentStateUtils/contentStateUtils.ts
shaulgo: Where is this used? I'm not sure what is actually more useful for our consumers. 
I wouldn't be surprised if in most scenarios what is actually wanted is content.toString().trim().length === 0. 
It all depends on what consumers need.

packages/ricos-content/web/src/contentStateUtils/contentStateUtils.ts
oded-soffrin: whoops. Added a test for that, and 0 block return true. > 1 blocks return false.

packages/ricos-content/web/src/contentStateUtils/createContent.ts
oded-soffrin: why? it's part of the content state right? I'm just using it here for tests for now (and in docs soon)
following aviv suggesting i'm going to re-use editorstate conversions version behavior

shaulgo: reusing it is also good.

maimonav: When we create new content we add version into it, as you can see here:
https://github.com/wix-incubator/rich-content/blob/2b1d9df2b9c6e2e9719cf77adddcc9f37513cabb/packages/editor/web/lib/editorStateConversion.ts#L52

shaulgo: take a look at the blame :)
I just didn't like the fixed version. And it would be updated once used...
Although we did talk about adding created version. This one is actually edited version.

PR #1628, opened 2020-10-12
Improve bundle size - remove FileInput in social polls viewer

mbutsykin:
get rid of `wix-rich-content-editor-common` in social polls viewer


shaulgo (approved):
So nice to see such a drop. This is in viewer side, so super important.

PR #1629, opened 2020-10-12
[accordion] tpa bug fixes

maimonav:


packages/plugin-accordion/web/src/components/modal-components/AccordionSettings.jsx
```diff
@@ -32,7 +32,11 @@ class AccordionSettings extends Component {
     return (
       <>
         <RadioGroupVertical
-          label={t('Accordion_AccordionSettings_Tab_Settings_CollapseView_Title')}
+          label={
+            <p className={this.styles.title}>
```
shaulgo, line 36: move to radioGroupVertical

packages/plugin-accordion/web/statics/styles/accordion-settings.scss
```diff
@@ -20,7 +26,15 @@
 }
 
 .direction_selector_option{
-  justify-content: flex-end !important;
+  p{
+    margin-block-start: 1em;
```
shaulgo, line 30: fix this

shaulgo (changes requested):

packages/plugin-accordion/web/src/components/modal-components/AccordionSettings.jsx
maimonav: done.

packages/plugin-accordion/web/statics/styles/accordion-settings.scss
maimonav: I thought about it again, and I will keep it like that, since this margin is only needed here in the direction_selector_option. If I'll need it in more places I'll refactor.

PR #1643, opened 2020-10-15
Convert template to ts

mharsat:
- types - many new types in `packages/common/types` & more that moved down from RicosTypes. Now that the whole flow of code is typed from common to ricos, many other fixes were made.
- `CreateInlineButtons` / `CreateInsertButtons` / `CreatePluginToolbar` - simplified these types, removed the `CreateInsertButtons<'t' | 'settings' | 'isMobile'>` pattern
- `create***Plugin.js` files are now well defined with the `CreatePluginFunction` type, heavily depending on the successful conversion of `createBasePlugin`.
- `pubsub` - moved to common
- `RichContetntEditor.tsx` & `RichContentViewer.tsx` 🤯
- Plugin template - All files are TS. Moved to the `packages` folder so all paths would be correct, `generatePlugin` script updated.
- Easy wins in all packages - basic typing or simple file extension renames
- The (in)famous RichContentProps definition is no longer needed since RCE & RCV are typed
- Many unused variables/parameters were exposed in the light of typing, and were removed


packages/common/web/src/types/RicosTypes.ts
```diff
@@ -0,0 +1,78 @@
+import { GetToolbarSettings, AnchorTarget, RelValue } from '.';
```
oded-soffrin, line 1: naming - ricosType in common package? intended ?

package.json
```diff
@@ -4,12 +4,13 @@
...
 		"examples/storybook",
-		"packages/*/web",
+		"packages/!(template-*)/web",
 		"bundle-analyzer",
 		"docs"
 	],
 	"resolutions": {
-		"rollup": "1.31.1"
+		"rollup": "1.31.1",
+		"@types/react-i18next/@types/i18next": "11.9.0"
```
lxgreen, line 13: what's the reason?

packages/template-atomic-plugin/web/package.json
```diff
@@ -1,5 +1,6 @@
 {
   "name": "wix-rich-content-plugin-yourDpluginDname",
+  "private": true,
```
lxgreen, line 3: what's the point to set it private? we'll forget to remove this for sure, and we'll have to release once again after realizing the new plugin is not published. wdyt?

scripts/generatePlugin.js
```diff
@@ -101,11 +101,14 @@ function createDirectoryContents(templatePath, newProjectPath, pluginData) {
...
       console.log(chalk.cyan(`Creating ${fileName} file`));
       const contents = fs.readFileSync(origFilePath, 'utf8');
-      const result = contents.replace(
+      let result = contents.replace(
         // eslint-disable-next-line max-len
         /yourDpluginDname|yourPluginName|YOUR_PLUGIN_NAME|YourPluginName|yourPluginVersion|pluginAuthorName|pluginAuthorMailAddress/g,
         name => pluginNameMap[name]
       );
+      if (fileName === 'package.json') {
+        result = result.replace(/"private": true,\n {2}/g, '');
```
lxgreen, line 110: a-ha!

scripts/generatePlugin.js
```diff
@@ -126,13 +129,13 @@ function addPluginToProject(projectPath, pluginName) {
     if (err) {
       console.log(chalk.red('Fail to read package.json', projectPath, err));
     } else {
-      const pckageJsonObj = JSON.parse(data);
+      const packageJsonObj = JSON.parse(data);
```
lxgreen, line 132: if you parse it anyways, then just `package.private = false`

packages/viewer/web/src/components/Anchors.spec.tsx
```diff
@@ -9,7 +9,7 @@ describe('Anchors', () => {
 
     //do something with datahook
     const anchor = document.querySelector(`[data-hook=hello]`);
-    anchor.innerHTML = `<div data-hook="ad-id">My Ad!</div>`;
+    if (anchor) anchor.innerHTML = `<div data-hook="ad-id">My Ad!</div>`;
```
lxgreen, line 12:
```suggestion
    if (anchor) {
    	anchor.innerHTML = `<div data-hook="ad-id">My Ad!</div>`;
    }
```

lxgreen (changes requested):
please revert both `rich-content-viewer.scss.android.js` and `rich-content-viewer.scss.ios.js`. These two files are here for OneApp workaround, they have to remain `js`. They copied as is to the dist/statics, not transpiled.

scripts/generatePlugin.js
mharsat: That is the example's package.json, to which the new plugin is added as a dependency. Also I prefer not having that field at all

package.json
mharsat: `@types/react-i18next` imports latest `@types/i18next`:
```
    "dependencies": {
        "@types/i18next": "*",
        "@types/react": "*"
    },
```
This causes a mismatch with the one we use. We use a very old version, newer i18next/react-i18next versions have type definitions internally

packages/common/web/src/types/RicosTypes.ts
mharsat: I moved some of its types that were needed in common - that's a good thing.
I agree naming is bad, gave it another reorg now

package.json
lxgreen: 👍

scripts/generatePlugin.js
lxgreen: oh, got it

packages/common/web/src/Utils/simplePubsub.ts
```diff
@@ -1,10 +1,13 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
```
shaulgo, line 1: why was it moved? It's only used in editor flow.

shaulgo (approved):

packages/template-atomic-plugin/web/package.json
shaulgo: I think it's fine. No need to release completely half baked plugins.

packages/common/web/src/Utils/simplePubsub.ts
mharsat: True, but it's a negligible amount of code, and I needed its type definition availableto other interfaces in the common package

packages/common/web/src/Utils/simplePubsub.ts
```diff
@@ -1,10 +1,13 @@
...
-  let state = initialState || {};
-  const listeners = {};
+type Callback = (...args: any[]) => any;
 
-  const subscribe = (key, callback) => {
+export const simplePubsub = (initialState?: Record<string, any>) => {
+  let state: { focusedBlock?: string; [key: string]: any } = initialState || {};
+  const listeners: Record<string, Callback[]> = {};
+
+  const subscribe = (key: string, callback: (...args: any[]) => any) => {
```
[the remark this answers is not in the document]
mharsat: Thanks

packages/template-atomic-plugin/web/package.json
mharsat: I remove that field later in the [generatePlugin](https://github.com/wix-incubator/rich-content/pull/1643#discussion_r508826994) script in this PR.
@shaulgo I agree that it's better practice not to publish plugins in development. Please decide what flow makes more sense and I'll make the change

packages/common/web/src/Utils/simplePubsub.ts
shaulgo: what's using it? maybe it should also move from common?

mharsat: `InlineButton`, `InsertButton`, `TextButtonMapper`, `CreatePluginFunction`

We're basically still following the convenient rule of thumb saying all types are imported from `wix-rich-content-common`. I think it will be easier for all devs to follow this when they need to import types. Also this was the only thing that needed to move in the whole PR, if this sums up to a critical mass in the future we will definitely move things around to editor-common

shaulgo: Would it make sense to have a types package?

mharsat: I considered that actually but I don't see any advantage to that since common is already common to all. Also having a package with almost zero runtime code doesn't make sense

packages/template-atomic-plugin/web/package.json
shaulgo: I'm fine with it both ways. But I rather have private=true. 
why is it removed?

mharsat: The template wasn't private before this PR, I made it private bc I moved it to the packages folder.
np I will make it stay private from now on

packages/common/web/src/Utils/simplePubsub.ts
shaulgo: I don't love it. But we can live with it. Tree shaking will solve this sort of problems.

lxgreen (approved):
sorry for delay

PR #1666, opened 2020-10-19
[anchor] fix: anchor on sites with fixed headers

noamtwix:


packages/common/web/src/Utils/anchorScroll.ts
```diff
@@ -0,0 +1,11 @@
+export const anchorScroll = element => {
+  const stickyHeaderHeight = document.querySelector('[id="SITE_HEADER"]')?.clientHeight || 0;
+  const stickyAdd = document.querySelector('[id="WIX_ADS"]')?.clientHeight || 0;
+  element.style.marginTop = `-${stickyHeaderHeight + stickyAdd}px`;
+  element.style.paddingTop = `${stickyHeaderHeight + stickyAdd}px`;
+  setTimeout(() => {
```
shaulgo, line 6: why timeout?

shaulgo (approved):

packages/common/web/src/Utils/anchorScroll.ts
noamtwix: weird behavior of the `scrollIntoView` that solve by that

noamtwix: after a second check, it looks like there is no need for the timeout. removing it

PR #1684, opened 2020-10-20
Vertical Embed Plugin slimLayout setting

ShamaySapir:


oded-soffrin:
Nice, we're missing: 
- documentation (instad of the comment in editorPlugins)
- working example (you can use existing vertical embed story)

oded-soffrin (approved):

PR #1687, opened 2020-10-22
[accordion] fix: add stopPropagation, to prevent draft from throwing getIn error.

maimonav:
Added `onFocus` to AccordionComponent to fix a problem of draft, that throws "getIn" error.

### DIFF
--- a/packages/plugin-accordion/web/src/accordion-component.jsx
+++ b/packages/plugin-accordion/web/src/accordion-component.jsx
@@ -1,3 +1,5 @@
+/* eslint-disable jsx-a11y/no-noninteractive-tabindex */
+/* eslint-disable jsx-a11y/no-static-element-interactions */
 import React from 'react';
 import PropTypes from 'prop-types';
 import AccordionPairs from './components/AccordionPairs';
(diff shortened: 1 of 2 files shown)


packages/plugin-accordion/web/src/accordion-component.jsx
```diff
@@ -153,7 +157,12 @@ class AccordionComponent extends React.Component {
     const direction = this.getDataManager().getDirection();
 
     return (
-      <div className={this.styles[direction]} data-hook="accordionComponent">
+      <div
+        className={this.styles[direction]}
+        onFocus={this.onFocus}
+        tabIndex="0"
```
shaulgo, line 163: are you sure `tabIndex="0"` is needed?

shaulgo (approved):
just check if tabIndex="0" is needed or not and merge

packages/plugin-accordion/web/src/accordion-component.jsx
maimonav: Yes, it doesn't fix it without it.

PR #1688, opened 2020-10-22
[editor] disable link-preview in inner-rce (until we support atomic plugins in inner-rce)

maimonav:
Currently, we allow only text plugins in inner-rce, but the config of the inner-rce is given from the outer-rce, which contains atomic plugins. So in order to disable link-preview - we need to delete link preview from inner-rce config.
@noamtwix please take a look as well.


packages/editor/web/src/RichContentEditor/InnerRCE.jsx
```diff
@@ -6,19 +6,33 @@ import RichContentEditor from './RichContentEditor';
...
     this.state = {
       editorState,
     };
   }
 
+  cleanConfig = config => {
+    const configWithoutAnchor = this.removeAnchorFromLink(config);
+    const configWithoutLinkPreview = this.removeLinkPreview(configWithoutAnchor);
+    return configWithoutLinkPreview;
+  };
```
noamtwix, line 27: in my opinion it's a bit more clear
```suggestion
  cleanConfig = config => {
    let clearConfig = config;
    clearConfig = this.removeAnchorFromLink(clearConfig);
    clearConfig = this.removeLinkPreview(clearConfig);
    return clearConfig;
  };
```

noamtwix:
@maimonav did you check how it looks when adding the link preview to the inner-rce plugins?

maimonav:
> @maimonav did you check how it looks when adding the link preview to the inner-rce plugins?

Yes, it works ok. But after talking to @oded-soffrin regarding it, we decided that we don't want to expose atomic plugins yet, only in next phase, so for now, we want to disable it.

packages/editor/web/src/RichContentEditor/InnerRCE.jsx
maimonav: Sure.

shaulgo (approved):

PR #1737, opened 2020-11-09
add viewer toolbar into RicosViewer

ShamaySapir:


ShamaySapir:
A working solution, 
might be better to include in Ricos Viewer (and not impl. side-by-side with RicosEngine as in this PR) as discussed w/ @mharsat 
@oded-soffrin FYI

PR #1738, opened 2020-11-10
[viewer] split translations 

oded-soffrin:


oded-soffrin:
There's an issue here with the way we do formatting.
if we split the formatting, in exmaple app the viewer and the editor changes the same i18n, which break the tests
if we bundle the i18n individualy, the bundle size increases.

need to fixed in a way i18n will be bundled only once, but editor/viewer won't effect each other

PR #1740, opened 2020-11-10
[externals] image-client-api

oded-soffrin:
@justame @lxgreenFYI - first try

### DIFF
--- a/rollup.externals.ts
+++ b/rollup.externals.ts
@@ -23,6 +23,7 @@ const externals = [
   /^wix-rich-content-viewer$/,
   /^ricos-content$/,
   /^react-player$/,
+  /^image-client-api$/,
 ];
 
 const excludedExternalsRegexArr = [


shaulgo:
do you know why packages grew?

shaulgo:
It will break the demos?

oded-soffrin:
not sure what happen here

shaulgo:
Maybe the size is because it needs to be external in the webpack config of the bundlesSizes?
Also, should be added to the demos package.json.

oded-soffrin:
it's still in dependencies, so this change should have effect anywhere if I understand correctly. what do I miss?

shaulgo:
My guess is that because it's not bundled in rollup. It's later bundled in webpack. And there it might be bundled twice or something like that. But because we get it from the window. It should be external in webpack as well. We shouldn't count it at all.
As for the demoes/tests. I would try to add it in the demos package.json. 
It's all a bit but it's definitely weird. I'll aim to extract it completely from ricos into wixos this Q. (Although I'm not sure if PG depend on it or not)

oded-soffrin:
normally, adding to rollup externals prevent the code to be bundled twice, because it's can be bundled once due to webpack.
we get it from window only at wix viewer, not on other environments...

PR #1749, opened 2020-11-11
fix/error handilng ui for blocks

talevy17:


packages/plugin-image/web/src/image-component.jsx
```diff
@@ -168,9 +171,10 @@ class ImageComponent extends React.Component {
           onCaptionChange={this.handleCaptionChange}
           setFocusToBlock={blockProps.setFocusToBlock}
           setComponentUrl={setComponentUrl}
+          t={t}
```
shaulgo, line 174: revert

packages/plugin-image/web/src/image-component.jsx
```diff
@@ -168,9 +171,10 @@ class ImageComponent extends React.Component {
...
         />
         {(this.state.isLoading || componentData?.loading) && this.renderLoader()}
-        {error && <MediaItemErrorMsg error={error} t={t} />}
+        {error && this.renderErrorMessage(error)}
```
shaulgo, line 177: revert

packages/plugin-image/web/src/toolbar/inline-buttons.tsx
```diff
@@ -45,7 +45,7 @@ const createInlineButtons: CreateInlineButtons = ({
     mobile: false,
     tooltipTextKey: 'ImageEditorButton_Tooltip',
     mapComponentDataToButtonProps: componentData => ({
-      disabled: isEmpty(componentData.src),
+      disabled: isEmpty(componentData.src) || !!componentData.error,
```
shaulgo, line 48: check what happens during load.

shaulgo (changes requested):

packages/plugin-image/web/src/toolbar/inline-buttons.tsx
talevy17: it's disabled during load as is

shaulgo (approved):

PR #1768, opened 2020-11-16
[toolbars] add new toolbars package

noamtwix:


packages/toolbars/web/src/Toolbar/ButtonTypes/ColorPickerButton.jsx
```diff
@@ -0,0 +1,129 @@
...
+          isMobile={isMobile}
+          icon={getIcon()}
+          theme={theme}
+        />
+        {isModalOpen && (
+          <div className={classNames(styles.modal, styles.withoutTop)}>
+            <ColorPicker
+              color={currentColor}
+              palette={palette.slice(0, 6)}
+              userColors={userColors.slice(0, 12)}
```
shaulgo, line 80: what happens if more then 12 are added?

shaulgo (changes requested):

PR #1772, opened 2020-11-16
[viewer] create viewer only entry points

oded-soffrin:
@shaulgo as discussed


packages/common/web/Tooltip/package.json
```diff
@@ -0,0 +1,10 @@
+{
+    "name": "tooltip",
+    "version": "1.0.0",
```
shaulgo, line 3: do we want the version?

oded-soffrin: nope

PR #1773, opened 2020-11-17
Side toolbar accessibility

tamrat-bay:


packages/common/web/src/Components/AccessibilityListener.tsx
```diff
@@ -3,9 +3,14 @@ import noOutlineStyle from '../../statics/styles/no-outline.scss';
 
 export default class AccessibilityListener extends Component<{ isMobile?: boolean }> {
   handleTabKeyUp = (e: KeyboardEvent) => {
+    const firstPluginButton = document.querySelector('[data-hook="pluginMenuPluginSection"]')
```
shaulgo, line 6: it works :) 
But this file is unrelated. It's intended to remove the blue outline from focusable components when it's not wanted.

shaulgo: let's add something like this when enter is clicked on the input component.
`const elements = document.querySelectorAll('input, button');
const index = elements.indexOf(currentElement);
elements[index+1]?.focus()
`

shaulgo (changes requested):

PR #1775, opened 2020-11-18
[Vertical Embed] Fonts Wiring

oded-soffrin:


packages/plugin-vertical-embed/web/src/components/styles.scss
```diff
@@ -0,0 +1,30 @@
...
+                font-family: var(--ricos-custom-p-font-family);
+                color: var(--ricos-text-color);
+            }
+
+            .vertical-embed-buttonText {
+                color: var(--ricos-background-color);
+            }
+
+            .vertical-embed-button {
+                background-color: var(--ricos-action-color);
```
Barackos, line 25: Add this before merge (we now support custom button color - #1808 )
```suggestion
                background-color: var(--ricos-custom-button-color, var(--ricos-action-color));
```

oded-soffrin: 🚀

PR #1777, opened 2020-11-18
[*] feat: Ricos Hooks

Barackos:
Things to consider for review:
- New prop: `hooks`, containing all Hook callbacks. Naming is flexible before we merge.
- New Context: `HooksContext`
- New Docs in Open Source
- `withHooks` - HOC for Context. I didn't attach it to every plugin as most don't involve hook events at the moment, and it's easy to add in the future anyway.


packages/editor/web/src/RichContentEditor/RichContentEditor.tsx
```diff
@@ -983,39 +986,41 @@ class RichContentEditor extends Component<RichContentEditorProps, State> {
...
-                onFocus={this.onFocus}
-                onBlur={this.onBlur}
-                style={this.props.style}
-                ref={measureRef}
-                className={wrapperClassName}
-                dir={direction || getLangDir(this.props.locale)}
-                data-id={'rce'}
-              >
-                {this.renderStyleTag()}
+          <HooksContext.Provider value={hooks}>
```
oded-soffrin, line 989: I found out that since the context is defined on richContent component and not RicosEditor/Viewer, the provider doesn't cover all the fullscreen/editor modal. 
I was thinking of moving the GlobalContext one step up, it might be better to introduce the new context in RIcosEditor / Viewer to begin with

packages/common/web/src/types/hooksTypes.ts
```diff
@@ -11,7 +11,7 @@ export interface onPluginAddStepArgs extends biCallbackParams {
   entryType: EntryType;
   step: 'FileUploadDialog' | 'PluginModal';
 }
-export interface BICallbacks {
+export interface RicosHooks {
```
lxgreen, line 14: Didnt we discuss an option to replace the individually named callbacks with more generic `on(hookType, callback)` approach?

oded-soffrin: +1

mharsat: Nice, then you won't need to maintain an object and will only need a single `on` function on the context like the `t` function

packages/editor/web/src/RichContentEditor/RichContentEditor.tsx
mharsat: I think that should generally be true for any new feature. Hoist functionality to ricos editor instead of writing features to legacy RCE

packages/common/web/src/types/hooksTypes.ts
Barackos: No problem 👍🏻

PR #1784, opened 2020-11-19
[editor] fix: switching focus between inner-rce's in composition mode (Draft-js & Android with Google keyboard issues)

maimonav:
In Mobile, using **Android** device & **Google** keyboard, when moving between Inner-rce's, the focus bumps to the previous editor that was focused, there are two `onFocus` actions that are been triggered because of Draft-js & Android event handlers bugs.
In order to fix it, when `onBlur` triggers (of the first Editor that is been blurred), setting the SelectionState to have `hasFocus` = false, disables the second unwanted `onFocus` actions from triggering.
This change will only happen when blurring from Inner-RCE and EditorState is in CompositionMode.

### DIFF
--- a/packages/editor/web/src/RichContentEditor/RichContentEditor.tsx
+++ b/packages/editor/web/src/RichContentEditor/RichContentEditor.tsx
@@ -23,7 +23,6 @@ import {
   getBlockType,
   COMMANDS,
   MODIFIERS,
-  SelectionState,
 } from 'wix-rich-content-editor-common';
 import { convertFromRaw, convertToRaw } from '../../lib/editorStateConversion';
 import { EditorProps as DraftEditorProps } from 'draft-js';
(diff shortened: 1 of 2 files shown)


packages/editor/web/src/RichContentEditor/RichContentEditor.tsx
```diff
@@ -828,13 +829,33 @@ class RichContentEditor extends Component<RichContentEditorProps, State> {
...
+      anchorKey: selection.anchorKey,
+      anchorOffset: selection.anchorOffset,
+      focusKey: selection.focusKey,
+      focusOffset: selection.focusOffset,
+      isBackward: selection.isBackward,
+      hasFocus: false,
+    });
+    const newEditorState = EditorState.set(editorState, {
+      selection: newSelection,
+    });
```
shaulgo, line 844:
```suggestion
    const selection = editorState.getSelection().merge({ hasFocus: false });
    const newEditorState = EditorState.set(editorState, { selection });
```

shaulgo (changes requested):

packages/editor/web/src/RichContentEditor/RichContentEditor.tsx
maimonav: Done.

shaulgo (approved):

shaulgo (approved):

PR #1794, opened 2020-11-24
[ricos-content] schema migration from draft

bchmn:


packages/ricos-content/web/src/migrateSchema/consts.js
```diff
@@ -0,0 +1,29 @@
...
+  headerOne: 'header-one',
+  headerTwo: 'header-two',
+  headerThree: 'header-three',
+  headerFour: 'header-four',
+  headerFive: 'header-five',
+  headerSix: 'header-six',
+  unorderedListItem: 'unordered-list-item',
+  orderedListItem: 'ordered-list-item',
+  blockquote: 'blockquote',
+  codeeBlock: 'code-block',
```
mharsat, line 13: typo

packages/ricos-content/web/src/migrateSchema/fromDraft.js
```diff
@@ -0,0 +1,234 @@
...
+          process.exit(1);
+      }
+    }
+  };
+
+  const parseAtomicBlock = block => getEntity(block.entityRanges[0].key);
+
+  const parseBlockQuoteBlock = block => ({
+    type: 'blockquote',
+    nodes: parseTextBlock(block),
```
mharsat, line 55: this should be an array of nodes, `parseTextBlock` returns a single node

packages/ricos-content/web/src/migrateSchema/fromDraft.js
```diff
@@ -0,0 +1,234 @@
...
+        textNodes.push(
+          createTextNode({
+            text: text.substring(currentPos, end),
+            blockData,
+          })
+        );
+        currentPos = end;
+      }
+    }
+    return textNodes;
```
mharsat, line 169: If `currentRange === {}` (ranges is empty), are all these cases relevant?
For example, there will be no `offset` or `length`

mharsat (changes requested):

packages/ricos-content/web/src/migrateSchema/fromDraft.js
bchmn: that's correct but the else block needs to run, otherwise text blocks without ranges will be ignored. feel free to refactor however you see fit.

mharsat (approved):

packages/ricos-content/web/src/migrateSchema/fromDraft.js
mharsat: I [added types](https://github.com/wix-incubator/rich-content/pull/1794/commits/09c5790d5adcb29c17bac81cbd238963f491b6e8) to for clarification and refactored accordingly, please review logic doesn't change

bchmn: looks good

PR #1796, opened 2020-11-24
[editor] fix: convert <pre/> html tag to unstyled instead of converting to code-block type

maimonav:
Convert `<pre/>` html tag to `unstyled` type instead of `code-block` type.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,8 @@
 ### :bug: Bug Fix
 - `viewer`
   - [#1780](https://github.com/wix-incubator/rich-content/pull/1780) preview interaction block wrapping logic refactored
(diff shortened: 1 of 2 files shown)


shaulgo (approved):
add changelog

oded-soffrin:
@maimonav what's the destiny of this PR ?
I see it's ancient and approved, any reason why it hasn't been merged?

I checked the difference behavior in the exampleApp, and i'm not sure I understand the nature of this change

maimonav:
I don't remember who asked this, @shaulgo do you remember maybe?

shaulgo:
I'm guessing that pasting from somewhere got converted to code blocks.
I don't remember the exact instance.

PR #1799, opened 2020-11-25
[file-upload] fix: remove example resolveFileUrl

talevy17:


packages/plugin-file-upload/web/src/file-upload-viewer.jsx
```diff
@@ -151,17 +150,19 @@ class FileUploadViewer extends PureComponent {
...
   renderFileUrlResolver() {
-    const { componentData, settings } = this.props;
-    if (componentData.error) {
+    const { componentData, settings: resolveFileUrl } = this.props;
+    if (componentData.error || !resolveFileUrl) {
+      if (!resolveFileUrl) {
+        // eslint-disable-next-line no-console
+        console.warn('Missing resolveFileUrl function');
+      }
       return this.renderError();
```
oded-soffrin, line 159: rendering error was previously used to display upload error, 
if missing resolveFileUrl from setting that's a consumer error and not upload failure, 
I think we should seperate those too cases

shaulgo (approved):

PR #1807, opened 2020-11-29
[plugin-table] clean legacy theme code

Barackos:


Barackos:
Used in other table plugin PR

PR #1836, opened 2020-12-07
Welcome to midgard

lxgreen:


lxgreen:
Due to a bug in yarn, it might not alert on missing dependencies when running with `--frozen-lockfile`, and try to install them during the build, making the build non-reproducible.

The yarn bug is not going to be fixed, as the team is focused on yarn v2.
We tried to mitigate it in falcon's code, but it didn't go as expected.

Please start using it locally as well and make sure to always update your lock file when changing dependencies.
The fork is compatible with yarn, so it shouldn't be a breaking change, and you will also enjoy faster installation times.
You can install and use it like so:
```
yarn global add midgard-yarn@1.23.21
midgard-yarn install
```

PR #1841, opened 2020-12-08
[headers-markdown] fix wix-rich-content-plugin-commons imports

audriusk-wix:


shaulgo:
Nice, catch. We test for sizes of plugins. So I didn't understand why it didn't show it decrease. 
I see that the bundle size test for this plugin is broken (doesn't import the viewer). 
I'll fix it first. I hope to have everything released tomorrow.

shaulgo:
this is what happens when it's tested correctly
image

PR #1860, opened 2020-12-13
[button] fix: remove onBlur that changed target checked value

maimonav:
`onBlur` event changed target checked value, fix it by removing it.


packages/plugin-button/web/src/components/settings-component.jsx
```diff
@@ -41,10 +41,6 @@ class SettingsComponent extends PureComponent {
     this.setState({ rel: event.target.checked });
   };
 
-  onBlur = event => {
```
shaulgo, line 44: Very strange. Any ideas why we had this?

shaulgo (approved):
Make sure this gives the result we want. This is very strange.

shaulgo:
test is failing for link button

packages/plugin-button/web/src/components/settings-component.jsx
maimonav: Indeed strange, I have no idea why it was implemented like that in the first place.
I don't see any reason why we need to set the checked value of the radio button to false onBlur.

maimonav:
I'm updating the test snapshot now, the change is in the contentState, target changed from `false` to `true`.

PR #1872, opened 2020-12-16
[build-optimizer] add @angular-devkit/build-optimizer as rollup plugin

maimonav:
Adding @angular-devkit/build-optimizer as rollup plugin.


oded-soffrin:
@shaulgo still relevant for devEx ?

shaulgo:
> @shaulgo still relevant for devEx ?

I wasn't aware of this PR. But a short read on what the package is. I'm guessing this is related to tree shaking work. I believe this can be deleted. We are taking a different route to tree shaking.

PR #1882, opened 2020-12-21
[toolbar] fix: buttons bg color issue with mobile "false hover"

tamrat-bay:


packages/editor-common/web/statics/styles/inline-toolbar-button.scss
```diff
@@ -14,7 +14,7 @@
 
   &:hover {
     button {
-      background-color: rgba($fallbackColor, 0.06);
+      background-color: rgba($fallbackColor, 0.00);
```
shaulgo, line 17: What's this for? This makes the background transparent. If this is what we want. No need to use variable?

PR #1884, opened 2020-12-21
[link] fix: pass settings to createToolbar

maimonav:
`settings` was missing in createToolbar.


shaulgo (approved):
I'll take it from you that this is good.

PR #1886, opened 2020-12-21
ricos injectedContent prop

lxgreen:
allows to inject content to mounted editor via dedicated prop


Barackos (approved):
Cool 👍

packages/ricos-common/web/src/RicosTypes.ts
```diff
@@ -45,6 +45,7 @@ export interface RicosEditorProps extends RicosProps {
   placeholder?: string;
   toolbarSettings?: ToolbarSettings;
   onBusyChange?: OnBusyChangeFunction;
+  injectedContent?: RicosContent;
   /* Changes to this interface should also be reflected in the API docs */
```
mharsat, line 49: >   /* Changes to this interface should also be reflected in the API docs */

new prop needs to be added to docs

PR #1917, opened 2020-12-27
[editor] feat: improve pasted color detection & disable FG black color

maimonav:
- Remove FG black color
- Support inline styles of pasted text that contains color string css (yellow, cyan and so on..).


packages/editor/web/src/RichContentEditor/utils/pasting/pastedContentUtil.js
```diff
@@ -4,14 +4,39 @@ import { Modifier, SelectionState } from 'wix-rich-content-editor-common';
...
+  if (color && !isBlack(color)) {
+    return true;
+  }
+  return false;
+};
+
+const getInlineColors = style => {
+  const styles = [];
+  if (style.color) {
+    const FG = rgbToHex(style.color);
```
shaulgo, line 19:
```suggestion
    const FG = colorNameToHex(style.color) || rgbToHex(style.color);
```

shaulgo (approved):

packages/editor/web/src/RichContentEditor/utils/pasting/pastedContentUtil.js
maimonav: Done.

PR #1944, opened 2021-01-04
[editor] feat: improve inline color filtering from pasted text

maimonav:
Filter color by luma value.
Luma value range is 0..255, where 0 is the darkest and 255 is the lightest.
Accept only colors that are bigger than 40 & smaller than 220.


maimonav:
@shaulgo Maybe we should consider add this filter as content state normalizer, to fix all current issues with theming.

PR #1945, opened 2021-01-04
ricoseditor subscribed to editorEvents.publish

lxgreen:


packages/editor/web/src/RichContentEditor/RichContentEditor.tsx
```diff
@@ -635,12 +635,13 @@ class RichContentEditor extends Component<RichContentEditorProps, State> {
...
 
-  publish = async (postId: string) => {
+  // TODO: remove deprecated postId once getContent(postId) is removed (9.0.0)
+  publish = async (postId?: string) => {
     if (!this.props.helpers?.onPublish) {
       return;
     }
     const { pluginsCount, pluginsDetails } = getPostContentSummary(this.state.editorState) || {};
-    this.props.helpers.onPublish(postId, pluginsCount, pluginsDetails, Version.currentVersion);
+    this.props.helpers.onPublish(pluginsCount, pluginsDetails, Version.currentVersion, postId);
```
oded-soffrin, line 644: why the reorder of params ?

oded-soffrin: oh because it can be null?

lxgreen: basically we don't need it anymore since the postId is passed to the biSettings. it's null in this case, but until we don't release v9.0, we can't deprecate it completely, so it's here for the `getContent(postId)` compatibility

oded-soffrin: but we're changing the 'onPublish' hook here (not that I think it's used already)
why not - this.props.helpers.onPublish(postId = '', pluginsCount, pluginsDetails, Version.currentVersion);

packages/editor-common/web/src/EditorEventsContext.jsx
```diff
@@ -35,16 +37,22 @@ export class EditorEventsProvider extends React.Component {
...
-  dispatch(event, data) {
+  async dispatch(event, data) {
     const callbacks = this.events[event] || [];
-
     return Promise.all(callbacks.map(cb => cb(data)));
   }
 
+  async publish() {
+    const publishResponse = await this.dispatch(EditorEvents.PUBLISH);
+    const editorResponse = publishResponse.filter(({ type }) => type === 'EDITOR_PUBLISH')[0];
```
oded-soffrin, line 52: maybe we can add default data instead of updating all publish subscribes that doesn't send data
    const editorResponse = publishResponse.filter(({ type } = {}) => type === 'EDITOR_PUBLISH')[0];

packages/plugin-social-polls/web/src/components/poll-context/PollContext.jsx
```diff
@@ -110,6 +110,8 @@ export class PollContextProvider extends PureComponent {
     }
 
     this.props.setPoll(dto);
+    console.debug('polls publish callback'); // eslint-disable-line
+    return { type: 'POLLS_PUBLISH', data: {} };
```
oded-soffrin, line 114: not needed if we use default data (see above)

packages/ricos-common/web/src/RicosTypes.ts
```diff
@@ -46,6 +46,15 @@ export interface RicosEditorProps extends RicosProps {
   toolbarSettings?: ToolbarSettings;
   onBusyChange?: OnBusyChangeFunction;
   injectedContent?: RicosContent;
+  editorEvents?: {
+    subscribe: (event: string, callback: () => Record<string, unknown>) => Record<string, unknown>;
+    unsubscribe: (
+      event: string,
+      callback: () => Record<string, unknown>
+    ) => Record<string, unknown>;
+    dispatch: (event: string) => Promise<unknown>;
```
oded-soffrin, line 55: structure of {type: POLLS_PUBLISH | EDITOR_PUBLISH, data: unknown } can be used here in new `publish` event

packages/ricos-editor/web/src/RicosEditor.tsx
```diff
@@ -90,11 +105,26 @@ export class RicosEditor extends Component<RicosEditorProps, State> {
   getContent = (postId?: string, forPublish?: boolean, shouldRemoveErrorBlocks = true) => {
     const { getContentState } = this.dataInstance;
     if (postId && forPublish) {
+      /* eslint-disable */
+      console.warn(
+        'Please use biSettings.postId and ref.publish() API for publishing. The getContent(postId, isPublishing) API is deprecated and will be removed in ricos v9.0.0'
```
oded-soffrin, line 110: LIKE

packages/ricos-editor/web/src/RicosEditor.tsx
```diff
@@ -90,11 +105,26 @@ export class RicosEditor extends Component<RicosEditorProps, State> {
...
+      );
+      /* eslint-enable */
       this.editor.publish(postId); //async
     }
     return getContentState({ shouldRemoveErrorBlocks });
   };
 
+  pollContentStateForUploads = resolve => {
```
oded-soffrin, line 118: it's all pending uploads, right?

packages/editor/web/src/RichContentEditor/RichContentEditor.tsx
lxgreen: hmm the params with default value should be listed in the end, otherwise it's ambiguous. 
```
given f(a = 0,b,c), the call f(1,2) means either f(a = 0, b = 1, c = 2) or f(a = 1, b = 2, c = undefined). so, f(a = 0,b,c) is actually invalid definition, and f(1,2) fails

```

alternatively, we can define `onPublish({ postId, pluginsCount, pluginsDetails, version }) to prevent this

packages/editor-common/web/src/EditorEventsContext.jsx
lxgreen: :+1:

packages/ricos-common/web/src/RicosTypes.ts
lxgreen: yap, i still didn't handle types, wanted to get it working asap

packages/ricos-editor/web/src/RicosEditor.tsx
lxgreen: yup, the `hasActiveUploads` checks image, gallery and file uploads. here, we just poll it internally

packages/editor/web/src/RichContentEditor/RichContentEditor.tsx
oded-soffrin: I meant in the call to always pass a param so it won't be optional and the API won't change

packages/ricos-editor/web/src/RicosEditor.tsx
oded-soffrin: ahhh i thought poll as in polls plugin. get it now

lxgreen: removed this piece, wanna add the publish subscription to the upload-related plugins in this PR? cause now the publish won't wait for uploads

PR #1958, opened 2021-01-06
[link] feat: custom link-v2

maimonav:


docs/docs/plugins_api/LinkPlugin.mdx
```diff
@@ -92,6 +80,31 @@ Default value:
...
+```
+
+### Custom Link
+
+Allows users to render custom link panel, save their data, and get it by the `onClick` event.
+
+### `onLinkAdd`
+
+```jsx
+onLinkAdd?: (CustomLinkData: any, saveData: (CustomLinkData: any) => void ) => void
```
shaulgo, line 100: In general. If you have params that are mandatory and params that are not. You put the mandatory params first. 
One could use saveData without using the already saved data.

docs/docs/plugins_api/LinkPlugin.mdx
```diff
@@ -92,6 +80,31 @@ Default value:
...
+```
+
+`onLinkAdd` is a function provided to give you the option to customize our default link behaviour,
+you can render your own link panel, and call `saveData` function to save your changes. <br />
```
shaulgo, line 104:
```suggestion
you can render your own link panel, and call `saveData` function to save your changes. You will need to use onClick in the viewer.<br />
```

packages/plugin-link/web/src/types.ts
```diff
@@ -2,6 +2,7 @@ import { EditorPluginConfig, ViewerPluginConfig } from 'wix-rich-content-common'
 
 export const EXTERNAL_LINK_TYPE = 'wix-draft-plugin-external-link';
 export const LINK_TYPE = 'LINK';
+export const CUSTOM_LINK_TYPE = 'wix-rich-content-plugin-custom-link';
```
shaulgo, line 5:
```suggestion
export const CUSTOM_LINK_TYPE = 'ricos-plugin-custom-link';
```

shaulgo: this will be the name in the future. We can already use it for new type. Everything will be converted in the new scheme.

shaulgo (approved):

packages/plugin-link/web/src/types.ts
maimonav: Sure, done.

docs/docs/plugins_api/LinkPlugin.mdx
maimonav: Done.

PR #1959, opened 2021-01-06
[fullscreen]: fix mobile horizontal view

talevy17:


packages/fullScreen/web/src/InnerFullScreen.jsx
```diff
@@ -69,7 +69,7 @@ export default class InnerFullscreen extends Component {
     let arrowsPosition = 0;
     let slideshowInfoSize = 0;
```
talevy17, line 70: const isMobileHorizontal =

PR #1973, opened 2021-01-10
[editor] fix: convertToRaw when inner-rce contentState is undefined

maimonav:
This fix comes from Standalone, which uses RichContentEditor with onChange and update editorState with inner-rce contentState undefined (in initialization of table/accordion).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,9 @@
     Click to see more.
   </summary>
 
+### :bug: Bug Fix
+- `editor`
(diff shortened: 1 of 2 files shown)


shaulgo (approved):
when does this happen?

maimonav:
In Standalone, when using Wicos & passing `onChange` to RichContentEditor child, while maintaining editorState in Standalone (like other consumers). 
This is for backwards compatibility, since we have new API in Ricos now, but this allows consumers to update to consume new Rce-on-Rce plugins with the previous API.

PR #1987, opened 2021-01-13
[unsupported-blocks] tests

tamrat-bay:


shaulgo:
You also need to add it to fixtures in tests/settings.js file. We need to render it in the test.

PR #1990, opened 2021-01-13
[ricos-editor] feat: RICOS_PUBLISH event

maimonav:
New EditorEvent: RICOS_PUBLISH.
This event fires only when old PUBLISH event has been completed, PUBLISH now refers to the PLUGIN_PUBLISH event once Groups update to the new event.


packages/ricos-content/web/src/consts.ts
```diff
@@ -9,6 +9,11 @@ export const HEADER_BLOCK = Object.freeze({
 
 export const DEFAULT_TITLE_HEADINGS = ['h2', 'h3'];
 
+export const EditorEvents = {
```
lxgreen, line 12: why did you put _editor_ events into the _common_ common? the polls has the editor-common dependency, so there's no problem to import it from there

lxgreen (changes requested):
please see my comment

packages/ricos-content/web/src/consts.ts
oded-soffrin: also please note ricos-content shouldn't have any component related data, this is a pure js package which handles content structure only

maimonav: Screen Shot 2021-01-14 at 10 31 59

That's why..
I wanted to set declare the 'rce:publish' const in a package that will not increase poll's bundle size.

oded-soffrin: I see. but content is not the place.
Since all the consumers will want to load editorEvents, and won't want to load editor-common as well, maybe we should create yet another package or a lib ?
Just a reminder - our editor have to import it from the same place, and it has to be externalized in order for it to work

I suggest split into two different PR, keep the string for now so groups will be able to update version, and solve the editorEvents location on a different PR

maimonav: Sure, and I think creating a lib in editor-common of EditorEvents is a good idea.

PR #2006, opened 2021-01-18
Rich content to plain text converter

mharsat:


oded-soffrin:
Lets think of the exported API
currently its `import { toPlainText } from 'ricos-content/libs/toPlainText';` which expects new ricos schema

1. our current draft consumers should be able to use it seamlessly, so it should use the fromDraft internally if needed 
(could be implicit by content detection, or explicit by different params)

2. should we have a single class to do that? new RicosContent(content).toPlainText(); ? should be taken into consideration with content APIs @lxgreen

mharsat:
1. np I totally agree. It can accept both formats and convert internally
2. very interesting. we could have something like
```tsx
new RicosContent(content).toPlainText();
new RicosContent(content).toDraft();
RicosContent.fromDraft(content);
new RicosContent(content).getPluginList();
```

oded-soffrin:
True. anyhow this contentApi class can consume the lib. 
tomorrow we have the sync w/ FAQ, we should merge by then I generate a example (I assume docs is the best, but if it's complicated storybook can worok)

packages/ricos-content/web/src/toPlainText/toPlainText.ts
```diff
@@ -22,16 +22,21 @@ import {
...
 
 export const toPlainText = async (
-  ricosContent: RicosContent,
+  content: RicosContent | RicosDraftContent,
   options?: PlainTextOptions
 ): Promise<string> => {
+  let ricosContent = content;
+  if ('blocks' in ricosContent) {
+    ricosContent = fromDraft(ricosContent);
+  }
```
oded-soffrin, line 39: Like!
I assume we'll use it in all APIs (`migrateContentIfNeeded` ?) better to place it somewhere handy

mharsat: Added cool ensure functions

examples/storybook/stories/Components/PlainTextConverter.tsx
```diff
@@ -0,0 +1,34 @@
+import React, { FunctionComponent, useEffect, useState } from 'react';
+import { RichContentEditorBox, Section } from './StoryParts';
+import EditorWrapper from './EditorWrapper';
+import styles from '../Components/styles.scss';
+import sourceCode from '!!raw-loader!./PlainTextConverter';
+import { toPlainText } from 'ricos-content/libs/toPlainText';
+import { RichContentTheme, RicosContent } from 'wix-rich-content-common';
```
oded-soffrin, line 7: can be loaded from ricos-content right?
the consumers of content API shouldn't know common which has many draft thingies

packages/ricos-content/web/src/imageUtils.ts
```diff
@@ -1,5 +1,5 @@
 import imageClientAPI from 'image-client-api/dist/imageClientSDK';
-import { ComponentData, Helpers } from '../types';
+import { ComponentData, Helpers } from 'wix-rich-content-common/src/types';
```
oded-soffrin, line 2: same here, why are we dependent on ricos-common?

mharsat: these are types, this import line never reaches runtime code (`imageUtils.js`), and is run only in compilation time to create types (`imageUtils.d.ts`). You can see that `wix-rich-content-common/src/types` isn't a dep in package.json.
Adding a value to this import (like `isSSR`) would immediately fail in CI since it's not a dep

examples/storybook/stories/Components/PlainTextConverter.tsx
mharsat: `RichContentTheme` is only a type, and it comes from `wix-rich-content-common`. same as other comment

packages/ricos-content/web/src/imageUtils.ts
oded-soffrin: I understand that. but:
- encapsulaiton - ricos-content should be a leaf, which common is consuming, so it doesn't make much sense it will consume common back. if they have common needed, they probably need to sit in `content` 
- `componentData` is already on `ricos-content` https://github.com/wix/ricos/blob/9e7fd74e2bc61582dae9933e656e50b00adc0620/packages/ricos-content/web/src/types/index.ts#L8

oded-soffrin: regarding helpers, that's a different story, but I don't see how this type helps us in any way

oded-soffrin (approved):

PR #2013, opened 2021-01-19
[external-toolbar] add alignment buttons to external toolbar

noamtwix:


packages/editor/web/src/RichContentEditor/Toolbars/createEditorToolbars.ts
```diff
@@ -68,6 +69,10 @@ const createEditorToolbars = ({
     pluginTextButtons: pluginTextButtonMap,
     pluginButtonProps,
   });
+  const renderDefaultFormattingToolbar = !customSettings.find(a => a.name === 'FORMATTING');
```
shaulgo, line 72: use some instead of find

shaulgo: isCustomFormattingToolbarDefined

PR #2016, opened 2021-01-19
[e2e] applitools logs

shaulgo:


oded-soffrin:
@shaulgo still relevant?

PR #2020, opened 2021-01-20
Support nested lists in migration to new schema

mharsat:
Adds support for nested lists in new schema migration tools.
Tested on:
1. All depths - up to 5
2. Bullet list in ordered list
3. Nested list item aligned to right
4. Top level list continues numbering after nested list


packages/ricos-content/web/src/migrateSchema/fromDraft/migratedFixtures/intro.json
```diff
@@ -402,18 +431,22 @@
...
+        "ricosParagraph": {
+          "depth": 0
+        }
       }
     ],
     "lastEdited": {
-      "seconds": 1610367712,
-      "nanos": 425000000
+      "seconds": 1611134447,
+      "nanos": 582000000
```
lxgreen, line 449: what's purpose of this field?

packages/ricos-content/web/src/migrateSchema/fromDraft/fromDraft.spec.ts
```diff
@@ -11,11 +11,13 @@ import complexRicosFixture from './migratedFixtures/migration-content.json';
...
 describe('migrate from draft', () => {
   it('should migrate intro fixture', () => {
-    expect(compare(fromDraft(fixture), ricosFixture)).toEqual({});
+    expect(compare(fromDraft(fixture), ricosFixture, { ignoredKeys: ['key'] })).toEqual({});
   });
 
   it('should migrate complex fixture', () => {
-    expect(compare(fromDraft(complexFixture), complexRicosFixture)).toEqual({});
+    expect(
+      compare(fromDraft(complexFixture), complexRicosFixture, { ignoredKeys: ['key'] })
```
lxgreen, line 19: i see the `ignoreKeys` is parametrized now. when do we actually want to compare block keys? anchors and lists?

lxgreen (approved):
generally, looks good. see my questions

packages/ricos-content/web/src/migrateSchema/fromDraft/migratedFixtures/intro.json
mharsat: That's a proto `Timestamp` field, we defined the `doc.lastEdited` field as a timestamp and that's how a timestamp is saved

packages/ricos-content/web/src/migrateSchema/fromDraft/fromDraft.spec.ts
mharsat: It's parameterized now since our to/from migration tools work in a level of precision that preserves the relevant block keys of all types of block. This means that all block keys in `toDraft(fromDraft(content)` are identical to `content`.
We do still pass that field for the `fromDraft` test since `fromDraft(content)` creates many nodes with newly generated keys (those nodes that don't represent a draft block i.e. `text` & `list_item`.

Since the keys can be trusted we don't need any extra work for anchors

PR #2042, opened 2021-01-25
Schema changes after revision

mharsat:
Includes changes requested by Wix platformization

- Schema is scoped in `rich_contnet` package (name not final)
- Message renamed: `RicosContent` --> `RichContent`
- RicosSelection was removed from content (`RicosDocument` props hoisted to `RichContent`)
- Ricos prefix was removed: `RicosDivider`, `RicosImage` --> `DividerData`, `ImageData`
- Node & Decoration data fields renamed: `ricosImage` --> `imageData`
- Node type & Decoration type are now enums
- `version` split into `createdVersion` & `updatedVersion`
- Versions and `lastEdited` move to new `metadata`
- Constants for plugins like `RICOS_IMAGE_TYPE = 'ricos-image` still remain but are not saved in ricos content anymore (replaced by enums), will be used later for describing plugin names or in prose schema

Also, `ricos-content/libs/migrateSchema` now exposes atomic conversion functions for node & decoration data <--> draft entity data


lxgreen (approved):
cool!

PR #2056, opened 2021-01-27
Add ability to resize buttons

tomrimk:


packages/viewer/web/src/getPluginViewers.jsx
```diff
@@ -142,6 +146,13 @@ class PluginViewer extends PureComponent {
             customStyles = { width: src.width, maxWidth: '100%' };
           }
         }
+
+        if (type === ACTION_BUTTON_TYPE || type === LINK_BUTTON_TYPE) {
```
talevy17, line 150: this "customStyles" section gathers a lot of logic, seems a bit messy. 
its unrelated to this PR but maybe we should rework this mechanism, it doesn't scale well IMO.

talevy17 (approved):
overall seems solid, we should consider unit testing getPluginViewers - seems to gather a lot of logic.

packages/viewer/web/src/getPluginViewers.jsx
maimonav: Agree.

packages/plugin-button/web/src/createButtonPlugin.ts
```diff
@@ -61,6 +71,30 @@ const createButtonPlugin = (
...
+      }
+      return PLUGIN_DECORATION_PROPS[PLUGIN_DECORATIONS.RESIZEABLE](calulatedProps);
+    },
+    componentWillReceiveDecorationProps: (props, nextProps, onPropsChange) => {
+      const { width } = PLUGIN_DECORATION_PROPS[PLUGIN_DECORATIONS.RESIZEABLE](props);
+      const { width: nextWidth } = PLUGIN_DECORATION_PROPS[PLUGIN_DECORATIONS.RESIZEABLE](
+        nextProps
+      );
+      if (width !== nextWidth) {
+        onPropsChange({ width: nextWidth, size: ButtonType.CUSTOM });
```
mharsat, line 95: This size should be one of the known sizes right? Like in this case `inline`

maimonav: It doesn't really matter, but I changed it to size `inline` for consistency.
To make it work in new schema, `inline` should have automatically convert to `CUSTOM`.

mharsat: great I'll take care of that side

maimonav: TOTACH

mharsat (approved):
As long as `width` and `size` are located in the config this will work in schema conversion.
Just make sure to check the conversion in the example before merging

maimonav:
> As long as `width` and `size` are located in the config this will work in schema conversion.
> Just make sure to check the conversion in the example before merging

Cool, they are located like you described, but don't forget to convert "inline" to "CUSTOM", since right now, it converts it to "UNKNOWN".

PR #2071, opened 2021-01-31
[inner-rce] fix: improve toolbars appearance in table plugin

noamtwix:


packages/editor/web/src/RichContentEditor/InnerRCE.jsx
```diff
@@ -53,21 +53,15 @@ class InnerRCE extends PureComponent {
...
-
   onClickOutside = e => {
     if (
       this.state.showToolbars &&
       this.editorWrapper &&
       e.target &&
       !e.target.closest('[data-id=rich-content-editor-modal]') &&
       !e.target.closest('[class=ReactModalPortal]') &&
-      !this.editorWrapper.contains(e.target)
+      !this.editorWrapper.contains(e.target) &&
```
shaulgo, line 63: this could be slow. We need to test it. If you have a table 20x20. It will be called 400 times at once.
I'll call you later and we will try to test it.

PR #2076, opened 2021-02-01
[inner-modal] fix: vertical overflow

noamtwix:


packages/editor/web/src/RichContentEditor/InnerModal.jsx
```diff
@@ -5,16 +5,33 @@ import ClickOutside from 'react-click-outsider';
 import { getLangDir } from 'wix-rich-content-common';
 
 class InnerModal extends Component {
+  setInnerModalRef = ref => (this.innerModalRef = ref);
+
+  getVerticalPosition = top => {
+    const editor = document && document.querySelector('[data-id=rce]');
```
shaulgo, line 11: if there is more than one editor this will not work. Let's add the editorRef to the globalContext

noamtwix: @shaulgo 
I thought of another solution.
what do you think?

PR #2079, opened 2021-02-01
experiments infra

lxgreen:


packages/editor/web/src/RichContentEditor/RichContentEditor.tsx
```diff
@@ -618,12 +629,22 @@ class RichContentEditor extends Component<RichContentEditorProps, State> {
         modifiers: [],
         key: 'Escape',
       },
+      this.props.experiments?.barrelRoll === 'True' && typeof window !== 'undefined'
```
oded-soffrin, line 632: can the GlobalContext expose a class instead of object ? 
Than we can do:
- experiments.enabled('barrlRoll')
- experiments.clientEnabled('barrlRoll') //for ensuring window defined

lxgreen: 👍 yep forgot we want it that way

PR #2088, opened 2021-02-03
[table]fix: trap focus on editing cell

danarish4:


packages/editor/web/src/RichContentEditor/RichContentEditor.tsx
```diff
@@ -400,9 +401,16 @@ class RichContentEditor extends Component<RichContentEditorProps, State> {
...
       renderInnerRCE: this.renderInnerRCE,
       innerRCERenderedIn,
+      disableKeyboardEvents: this.disableKeyboardEvents,
     };
   };
 
+  disableKeyboardEvents = shouldDisable => {
+    if (!this.props.isInnerRCE) {
+      this.readOnly = shouldDisable;
+    }
```
shaulgo, line 411: This is used in render. So should probably be in state
```suggestion
    if (!this.props.isInnerRCE && readOnly !== this.state.readOnly) {
      this.setState({disableKeyboardEvents: shouldDisable});
    }
```

packages/plugin-table/web/src/components/CellRenderer.js
```diff
@@ -223,27 +219,24 @@ class Editor extends Component {
...
 
   render() {
-    const { children, editing } = this.props;
+    const { children, editing, selected } = this.props;
     return (
       // eslint-disable-next-line jsx-a11y/no-static-element-interactions
       <div
         className={classNames(styles.editor, editing ? styles.edit : styles.view)}
-        onKeyDown={this.onKeydown}
+        data-id={'cell-editor'}
```
shaulgo, line 233: is this used?

shaulgo (approved):

PR #2099, opened 2021-02-04
[ricos + legacy] fix(callbacks): onLoad events double trigger

Barackos:
Atm, `localeStrategy` deliberately remounts `RCE` & `RCV` in order to compensate the current inability to update `localeResource` on runtime (due to design).

As a result, `onViewerLoaded` and `onOpenEditorSuccess` cannot exist within `componentDidMount` of `RCV` & `RCE`, as these are triggered more than once because of the remount.

Therefore in this PR these callbacks are temporarily moved to `RicosEditor` and `RicosViewer`, to ensure they're triggered only once

We should consider replacing the current remount mechanism with Context / something equivalent, as these callbacks should exist within legacy components.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,79 +10,115 @@
 > - :nail_care: Polish
 
 ## [Unreleased]
+
 <details>
   <summary>
     Changes that have landed in master but are not yet released.
     Click to see more.
   </summary>
 
+### :bug: Bug Fix
+
+- `editor`
+  - [##2099](https://github.com/wix/ricos/pull/#2099) fix `onOpenEditorSuccess` unwanted triggers
+- `viewer`
+  - [##2099](https://github.com/wix/ricos/pull/#2099) fix `onViewerLoaded` unwanted triggers
+
 </details>
 <hr/>
 
 ## 8.19.1 (Feb 6, 2021)
+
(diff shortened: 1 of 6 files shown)


mharsat:
Do we necessarily need to have these callbacks in legacy components? There's only one consumer left in legacy
Also, in this case it was us internally remounting the component, but our consumers sometimes remount ricos too. This means that if a consumer has a remount mechanism then the number of BI events we get might be doubled, and bias our consumer analysis.
It might be good to research methods for aggregating events coming from the same end point over a small period of time

Barackos:
> Do we necessarily need to have these callbacks in legacy components? There's only one consumer left in legacy

Not necessarily, but as you inferred it is no good for our legacy API.

> Also, in this case it was us internally remounting the component, but our consumers sometimes remount ricos too. This means that if a consumer has a remount mechanism then the number of BI events we get might be doubled, and bias our consumer analysis.
> It might be good to research methods for aggregating events coming from the same end point over a small period of time

We should encourage consumers to not remount us, or at least know their cause for it & solve it if it's due to us.
I say it because we can't tell what the upper bound for "small period of time" is (client runtime), so no mechanism can assure there will be no bias if Ricos is remounted by consumers.

I'll add `debounce` which should solve it for the most part.

PR #2119, opened 2021-02-09
[link] fix anchors href bug in editor, and set cursor: text in editor

maimonav:
- event.stopPropagation() when clicking on anchor in Editor.
- set cursor: text when hovering over link in Editor.


shaulgo (approved):
Very good

PR #2123, opened 2021-02-09
[plugin-table] remove margin from code block & lists

noamtwix:


packages/editor/web/statics/styles/rich-content-editor.scss
```diff
@@ -154,6 +154,10 @@
   line-height: 26px;
   padding: 2px 25px;
   white-space: pre-wrap;
+
+  .renderedInTable & {
+    margin: 0 0;
```
shaulgo, line 159:
```suggestion
    margin: 0;
```

shaulgo (approved):

PR #2137, opened 2021-02-11
fix/ gallery and image native replace button

talevy17:


packages/plugin-commons/web/src/Base/toolbars/createAtomicPluginToolbar.jsx
```diff
@@ -400,10 +400,6 @@ export default function createAtomicPluginToolbar({
       ) : null;
     }
 
-    onClick = e => {
```
shaulgo, line 403: I'm guessing this was added for a reason. Did you check the commit message for this?

PR #2138, opened 2021-02-11
[viewer] improve performance

maimonav:


packages/viewer/web/src/utils/convertContentState.tsx
```diff
@@ -218,6 +214,7 @@ const normalizeContentState = (contentState: RicosContent): RicosContent => ({
     }
 
     const data = { ...block.data };
+    data.depth = block.depth;
```
oded-soffrin, line 217: suggestion:
    const data = { ...block.data, depth: block.depth };

oded-soffrin:
@maimonav @justame let's NOT merge to master, but have this under some feature toggle so we can measure it (important for performance review)

packages/viewer/web/src/utils/convertContentState.tsx
maimonav: Done.

packages/viewer/web/src/utils/convertContentState.tsx
```diff
@@ -212,16 +207,18 @@ const getEntities = (
...
-    const direction = getTextDirection(block.text);
-    if (direction === 'rtl') {
-      data.textDirection = direction;
-    }
+    const textDirection = getTextDirection(block.text);
+    const data = {
+      ...block.data,
+      depth: block.depth,
+      index,
+      textDirection: textDirection === 'rtl' ? textDirection : undefined,
```
oded-soffrin, line 220: I don't know if it makes any difference, 
but if textDirection != rtl, now data contains a texDirection (which is undefined) as oppose to previously where data didn't contain such key
you can achieve the same previous behavior like so:
```
const data = {
    ...block.data,
      depth: block.depth,
      index,
   ...(textDirection && {textDirection === 'rtl'}
)
```

because ...(false) won't add anything to the data object

maimonav: Pretty sure it won't make a difference, since data[textDirection] = undefined in both cases.
Regarding your suggestion, good to know, textDirection will not be false, it will be 'ltr' || 'rtl' || 'neurtal' , and then it will change the current behaviour.

oded-soffrin: alright `...(textDirection: textDirection === 'rtl' && {textDirection})` then :)

but as I said, only on "good to know" basis

maimonav: `...(textDirection === 'rtl' && {textDirection})`
Done, thanks :)

PR #2151, opened 2021-02-16
Use new content in example app

mharsat:
- Example app now supports a "Use New Content" config. When it's set, the content state editor section will display new content.
- Fixed issue in the example app where clicking on editor settings didn't change the setting button highlighting
- `bundle-analyzer` - added new bundle for schema


oded-soffrin:
Also changes here, example app is using ricos instead of wixRichContent, @shaulgo we're fine with that, right ?

mharsat:
> Also changes here, example app is using ricos instead of wixRichContent, @shaulgo we're fine with that, right ?

This is optional, you activate it with a toggle and the default is rich content with draft content as it is now

shaulgo:
> Also changes here, example app is using ricos instead of wixRichContent, @shaulgo we're fine with that, right ?

Awhile back matan told me he was doing this. I was already considering doing this myself. Very good timing :)

oded-soffrin (approved):

PR #2197, opened 2021-02-23
[soundCloud] remove soundCloud plugin

danarish4:


shaulgo:
You need to make sure to remove it from serverless first. Then you can merge this.

PR #2207, opened 2021-02-24
[image] fix(perf): quality preload

oded-soffrin:


oded-soffrin:
@Shaule1 @justame @lxgreen anxious to test on wix, so merging (under experiment)
*but* this branch contain alternative optimzation approach - render a better preloaded image to move the LCP to an earlier phase
(tested on e2e dev app, seem that blur, animation and replacement of image has no effect on LCP, only the quality and proportions of the preloaded image)

PR #2210, opened 2021-02-24
[internal] throw error if content is not serializable

danarish4:


examples/main/src/Components/ContentStateEditor.tsx
```diff
@@ -65,9 +65,12 @@ class ContentStateEditor extends PureComponent<Props> {
   componentWillReceiveProps(nextProps: Props) {
     const { contentState, shouldUseNewContent } = nextProps;
     if (!this.monaco?.editor.hasTextFocus()) {
-      this.setState({
-        value: stringifyJSON(shouldUseNewContent ? ensureRicosContent(contentState) : contentState),
-      });
+      const content = shouldUseNewContent ? ensureRicosContent(contentState) : contentState;
```
shaulgo, line 68: JSON.stringify replacer function. If you have function throw error.

PR #2214, opened 2021-02-24
[link] fix/ disable link button when atomic block is in selection

talevy17:


CHANGELOG.md
```diff
@@ -20,6 +20,10 @@
 - `ricos-content`
   -  [#2192](https://github.com/wix/ricos/pull/2192) Truncate - docs & finialize API
 
+### :bug: Bug Fix
+- `link`
+  - [#2214](https://github.com/wix/ricos/pull/2214) link button is disabled when there is an atomic block in the selection
```
shaulgo, line 25: this will be confusing when you can't use the button. I thought we would filter out all atomic blocks. And add link on selected text.
If we do it this way, we need a tooltip when it's disabled.

talevy17: yea I thought it wasn't ideal, but filtering out atomic blocks won't be very easy as well - draft applies links based on selectionState so it would require splitting the selection for every atomic block in the original selection and then apply the link to each new selection @shaulgo

maimonav (approved):
Looks good, please add changelog :)

PR #2217, opened 2021-02-24
render full screen on hover

justame:


justame:
@lxgreen, please review

oded-soffrin:
image

reproduce:
- load this https://rich-content-storybook-render-fullscreen-on-hover.surge.sh/?path=/story/plugins--image
- wait with the cursor outside of the viewer
- hover on image and click it (quickly)

also happen on master (https://rich-content-storybook-render-fullscreen-on-hover.surge.sh/?path=/story/plugins--image)
but only if you click when page is loading. this PR is making the issue worse IMO

justame:
@oded-soffrin , It means that there is a problem with the logic. I'll review it with Alex and we will fix it.

e2e/tests/media.e2e.js
```diff
@@ -343,13 +343,13 @@ describe('plugins', () => {
     after(() => cy.eyesClose());
 
     //TODO: fix this flaky test
-    it.skip('render upload modal', function() {
+    it('render upload modal', function() {
```
oded-soffrin, line 346: @justame  did you intentionally restore those flaky tests ?

justame: no, I only merged from master and pushed

PR #2218, opened 2021-02-24
use srcset

justame:


justame:
@lxgreen, please review

packages/plugin-image/web/src/image-viewer.jsx
```diff
@@ -203,11 +222,16 @@ class ImageViewer extends React.Component {
 
   getImage(imageClassNames, src, alt, props, opts = {}) {
     const { fadeIn = false, width, height } = opts;
+    let srcSet;
+    if (this.shouldUseSrcSet()) {
+      srcSet = replaceUrlFileExtenstion(src, '.webp');
+    }
```
oded-soffrin, line 228: didn't we said we need it only for preloaded image?

justame: I don't see a problem doing it for both

packages/plugin-image/web/src/image-viewer.jsx
```diff
@@ -203,11 +222,16 @@ class ImageViewer extends React.Component {
...
     return (
       <img
         {...props}
         className={imageClassNames}
         src={src}
+        srcSet={srcSet}
```
shaulgo, line 234: check with media when webp is supported

packages/plugin-image/web/src/image-viewer.jsx
```diff
@@ -22,6 +22,20 @@ import InPluginInput from './InPluginInput';
 
 const isSafari = () => /^((?!chrome|android).)*safari/i.test(navigator.userAgent);
 
+const replaceUrlFileExtenstion = (url, extensionTarget) => {
+  const originalExtensionFound = url.lastIndexOf(extensionTarget) > -1;
```
shaulgo, line 26: take the last 4/5 chars.

oded-soffrin:
It’s under experiment, so why not. I have two open questions before opening:
- will there always be a webp format available?
- the highres is client only, where we used the imageClientAPI. if it doesn’t return webp, why would we want to force replace? it’s probably not supported, so I think it’s safer to use only in preload, and the switch should happen [here](https://github.com/wix/ricos/blob/master/packages/ricos-content/web/src/imageUtils.ts#L67)  where we generate the URL

packages/plugin-image/web/src/image-viewer.jsx
oded-soffrin: see comment below

packages/plugin-image/web/src/image-viewer.jsx
lxgreen, line 26:
```suggestion
return url.replace(/(.*)\.(jp(e)?g|png)$/, `$1.${extensionTarget}`)
```

lxgreen (commented):
a shorter solution

justame:
I talked with Media and they always support webp for images

packages/plugin-image/web/src/image-viewer.jsx
justame: they support

packages/plugin-image/web/src/image-viewer.tsx
```diff
@@ -75,6 +80,21 @@ class ImageViewer extends React.Component<ImageViewerProps, ImageViewerState> {
 
   static contextType = GlobalContext;
 
+  shouldSkipImageThumbnail = () => {
```
oded-soffrin, line 83: did you merge from master? I've removed all the `shouldSkipImageThumbnail` experiment

justame: I had a big conflict, I'll clean it

PR #2227, opened 2021-02-28
[table] handle first and last blocks in table cells

noamtwix:


packages/plugin-table/web/src/components/CellRenderer.js
```diff
@@ -124,6 +124,31 @@ export default class Cell extends Component {
...
+      let firstBlockText;
+      if (setEditorRef) {
+        const currentContent = editorState.getCurrentContent();
+        lastBlockText = currentContent.getLastBlock().getText();
+        firstBlockText = currentContent.getFirstBlock().getText();
+      } else {
+        lastBlockText = blocks[blocks.length - 1].text;
+        firstBlockText = blocks[0].text;
+      }
+      const hideFirstBlock = firstBlockText === '' || firstBlockText === '​'; //zero-width space
```
shaulgo, line 144: zero-width space?

packages/plugin-table/web/src/components/CellRenderer.js
```diff
@@ -124,6 +124,31 @@ export default class Cell extends Component {
...
+      const hideLastBlock = lastBlockText === '';
+      return { hideFirstBlock, hideLastBlock };
+    } else {
+      return false;
```
shaulgo, line 148: let's return an {} object. So it's the same type.

packages/plugin-table/web/src/components/CellRenderer.js
```diff
@@ -196,6 +222,9 @@ export default class Cell extends Component {
       >
         <div
           className={classNames(
+            setEditorRef ? styles.editorWrapper : styles.viewerWrapper,
+            !isEditing && this.hideBlocks().hideFirstBlock && styles.hideFirstBlock,
```
shaulgo, line 226: don't call hideBlocks() twice. 
const { hideFirstBlock, hideLastBlock } = ...

packages/plugin-table/web/statics/styles/cell.scss
```diff
@@ -37,6 +37,40 @@
...
+                height: 1px;
+            }
+        }
+    }
+}
+
+.hideFirstBlock {
+    &.editorWrapper {
+        :global(.public-DraftEditor-content) > div > * {
+            &:first-child {
```
shaulgo, line 60:
```suggestion
        :global(.public-DraftEditor-content) > div > :first-child {
```

shaulgo: and the rest...

packages/plugin-table/web/src/components/CellRenderer.js
shaulgo: 1. maybe just change it to getHideBlocksClasses(). And get the classes. 
2. Split the classes to viewer and then have this function return the correct one.

packages/plugin-table/web/src/components/CellRenderer.js
```diff
@@ -196,6 +222,9 @@ export default class Cell extends Component {
       >
         <div
           className={classNames(
+            setEditorRef ? styles.editorWrapper : styles.viewerWrapper,
```
shaulgo, line 225: this is not needed

shaulgo (approved):

PR #2234, opened 2021-03-01
Content API - From Plain Text

mharsat:


packages/ricos-content/web/src/fromPlainText/fromPlainText.ts
```diff
@@ -0,0 +1,40 @@
+import { Node_Type, RichContent, Node, Common_TextAlignment } from 'ricos-schema';
+import { genKey } from '../migrateSchema/generateRandomKey';
+import { Version } from '../version';
+import { createTimestamp } from '../migrateSchema/createTimestamp';
+
+export const fromPlainText = (text: string): RichContent => {
+  const paragraphs = text.split('\n').map(
+    (paragraph: string): Node => ({
```
oded-soffrin, line 8: this functionality of building a paragraph node, can be used from the same place it is probably used in fromDraft function?

mharsat: The function that creates a paragrapg in `fromDraft` focuses mainly on parsing the block and creating text nodes from a draft-js entityMap, this use case is much lighter, the only thing they have in common is `type` and `key`

oded-soffrin: I'm talking of a common function that will serve both, and other builders to come
for example:

```
  const createParagrah = (paragraph: string, paragraphData: ParagraphData): Node => ({
    type: Node_Type.PARAGRAPH,
    key: genKey(),
    nodes: paragraph
      ? [
          {
            type: Node_Type.TEXT,
            key: genKey(),
            nodes: [],
            textData: {
              text: paragraph,
              decorations: [],
            },
          },
        ]
      : [],
    paragraphData,
  });
  ```

oded-soffrin (approved):

PR #2235, opened 2021-03-01
[workflows] feat: docs deploy on release

Barackos:
Created a workflow that will be triggered when a new version is released, excluding alpha releases


mharsat:
Did you follow the [docusaurus docs](https://v2.docusaurus.io/docs/deployment#triggering-deployment-with-github-actions) & check this works?
The deploy keys didn't work for me when I tried a few months ago.
Also, once we move to janet and deploy to our new domain this will change

PR #2256, opened 2021-03-03
[html]fix: html height

danarish4:


packages/plugin-html/web/src/HtmlComponent.jsx
```diff
@@ -71,6 +71,18 @@ class HtmlComponent extends Component {
     }
   };
 
+  getIframeHeight = () => {
+    const { height } = this.props.componentData.config || {};
+    const { iframeHeight } = this.state;
+    //for avoiding unnecessary scroll
```
shaulgo, line 77: Method is not used?
Also make it less complicated. Less if, less return.

danarish4: and now ?

shaulgo (approved):

PR #2259, opened 2021-03-03
translations-remount fix

justame:


packages/ricos-viewer/web/src/RicosViewer.tsx
```diff
@@ -1,5 +1,5 @@
 import React, { Component } from 'react';
-import { RicosEngine, shouldRenderChild, localeStrategy } from 'ricos-common';
+import { RicosEngine, shouldRenderChild } from 'ricos-common';
```
shaulgo, line 2: can't we delete the localeStrategy?

oded-soffrin: Still used in editor

packages/ricos-viewer/web/src/RicosViewer.tsx
```diff
@@ -29,18 +29,7 @@ export class RicosViewer extends Component<RicosViewerProps, State> {
...
 
-  updateLocale = async () => {
-    const { _rcProps: { experiments } = {} } = this.props;
-    await localeStrategy(this.getLocale(), experiments).then(localeData =>
-      this.setState({ localeData, remountKey: !this.state.remountKey })
-    );
-  };
-
   componentDidMount() {
-    if (this.getLocale() !== 'en') {
```
oded-soffrin, line 40: @justame no longer needed?

packages/ricos-viewer/web/src/RicosViewer.tsx
shaulgo: So we need two places that load the locale?

oded-soffrin: we don't need two places, but if this is what we want, editor should do what the viewer is doing in this PR before we can delete it.

packages/ricos-viewer/web/src/RicosViewer.tsx
oded-soffrin: Did you check if it doesn't have performance degradation?

justame: I haven't checked if there is performance degradation.
I did notice that If you change the locale in the runtime the translations do not update, so the viewer will render the initial locale that was defined.
Ricos Viewer(locale:he) => click a button and change locale to en => translations stay hebrew.

oded-soffrin: I understand. Not sure we have such use-case of updating translation in realtime outside the example app, but let's verify the performance before merging

packages/ricos-viewer/web/src/RicosViewer.tsx
oded-soffrin: also, I believe the usage in withi18next is not ideal. we had another remount mechanism there, and we wrap the whole editor/viewer with conusmer, instead of creating many small consumers for every place translation is needed

packages/common/web/src/Utils/withI18n.tsx
```diff
@@ -33,17 +36,40 @@ export default <T, P>(Component: ComponentType, defaultLocaleResource: LocaleRes
       };
     }
 
+    async componentDidMount() {
+      const { locale, localeResource } = await this.getResourceByLocale(this.props.locale);
```
oded-soffrin, line 40: minor: not needed if locale is 'en' (redundant call)

oded-soffrin:
for visibility, this changes breaks the full screen behavior on viewer, since it's assuming re-mount (see [loadEditorModalAfterLocaleResourceIsLoadedToPreventRemountHackFromBreakingModal](https://github.com/wix/ricos/blob/master/packages/ricos-viewer/web/src/modals/fullscreen/FullscreenProvider.tsx#L57)

There was no test coverage and we only found it in the last minute

@justame to add test on fullscreen in non-english language for viewer
@mharsat do we have a test on non-english for editor modal? in case of similar improvements in the future

oded-soffrin (changes requested):
Fix full screen in non-english

mharsat:
@oded-soffrin [render external modal in rtl](https://wixeyes.applitools.com/app/baselines/k~!7f68d511-7055-4f8d-b7ef-aa3fd4767490~!b3e35531-3d00-4ffc-858f-741205599629~!3d5c0fe5-d877-4b23-aea2-a5b6d046d5ef~!148cG7CZUs3TD9yhy5zv2bQq5pXFQjvbygpaXRTI_IU_~!/00000251787171709961/steps/5/edit?accountId=X3Sruf7w5kCEbUEEu1ooeQ~~)

justame:
All checks are green, should I update the changelog?

PR #2300, opened 2021-03-11
paste image support

oded-soffrin:
Paste image support, using editorCommands branch


oded-soffrin:
@maimonav @shaulgo that's the PR i was talking about, using editorCommands to add paste image support

shaulgo (commented):
Super cool. 
I'm pretty sure I don't want to expose `block`. This is the internal data structure. Seems like blockKey should be enough. 
Once I finish going over the editor commands. I will be wiser.

oded-soffrin:
> I'm pretty sure I don't want to expose `block`. This is the internal data structure. Seems like blockKey should be enough.

smart. changed return type to (string | void)

maimonav:
Opened a new PR, including only required changes to add this functionality, [here](https://github.com/wix/ricos/pull/2350)

PR #2312, opened 2021-03-18
[plugin-giphy] Add more descriptive alt text and missing required hotkeys

wix-heathercreech:
role="button" on a div tells screen readers that all hotkeys we'd use on buttons is supported for the element. That means we're missing the Space hotkey support on gifs in the giphy selector.

The important bits:
- Add missing Space hotkey
- Switch from depreciated event fields to non-depreciated ones
- Switch onKeyPress (depreciated) > onKeyDown
- Use the giphy title for the alt text instead of just gif


packages/plugin-giphy/web/src/toolbar/giphySelector.tsx
```diff
@@ -78,9 +114,9 @@ class GiphySelector extends Component {
       pubsub.update('componentData', { gif });
     }
 
-    if (helpers) {
-      helpers.openModal(data => pubsub.update('componentData', { metadata: { ...data } }));
-    }
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore: name doesn't exist in label
+    helpers?.openModal?.(data => pubsub.update('componentData', { metadata: { ...data } }));
```
oded-soffrin, line 119: @lxgreen I know this is very very legacy, but TS give me issues here
openModal everywhere on the project accept object, but here it's a function.
any idea?

PR #2342, opened 2021-04-05
[table] fix: code-block margin in viewer

noamtwix:


packages/viewer/web/src/RichContentViewer.tsx
```diff
@@ -242,6 +242,7 @@ class RichContentViewer extends Component<
       const editorClassName = classNames(
         styles.editor,
         renderedInTable && styles.renderedInTable,
+        renderedInTable && viewerStyles.renderedInTable,
```
danarish4, line 245: maybe you can do a little refactor .. something like:

const tableClassNames = classNames(styles.renderedInTable, viewerStyles.renderedInTable, draftDefaultStyles.renderedInTable)

and then
renderedInTable && tableClassNames

danarish4 (approved):

PR #2350, opened 2021-04-06
feat/ paste image using editor commands

maimonav:
Paste image support.


examples/main/shared/editor/Editor.tsx
```diff
@@ -299,6 +299,7 @@ export default class Editor extends PureComponent<ExampleEditorProps, ExampleEdi
               placeholder={'Add some text!'}
               plugins={this.ricosPlugins}
               linkPanelSettings={this.config.uiSettings.linkPanel}
+              _rcProps={{ experiments }}
```
oded-soffrin, line 302: @danarish4 FYI adding experiments support in playground for ricos use-case too

PR #2360, opened 2021-04-08
fix paste image in BM

maimonav:


oded-soffrin:
solved in a different branch

PR #2362, opened 2021-04-08
[bi] add button link bi

danarish4:


packages/plugin-button/web/src/toolbar/buttonInputModal.jsx
```diff
@@ -65,13 +72,29 @@ export default class ButtonInputModal extends Component {
...
+      settings: { rel, target, url },
+      validUrl,
+    } = this.state;
+    validUrl &&
+      this.props.helpers?.onPluginAction?.(ADD_PLUGIN_LINK_BI, {
+        plugin_id: LINK_BUTTON_TYPE,
+        link: url,
+        newTab: isNewTab(target),
+        category: WEB_ADDRESS_CATEGORY,
+        nofollow: rel,
```
maimonav, line 86: `rel` value contains 'nofollow', so it can lead to missinformation.

maimonav (approved):
Looks good, just take a look on the `rel` comment. :)

packages/plugin-button/web/src/toolbar/buttonInputModal.jsx
maimonav: IMO, the BI event need to be altered, having `rel` param instead of `nofollow`.

danarish4: I checked all 'rel' values and its always set to true/false

PR #2373, opened 2021-04-12
Use native proto timestamp

mharsat:
- Removed local `timestamp.proto` file and use native `google/proto/timestamp`. Uses `Date` type for `updatedTimestamp` and `createdTimestamp`
- Added "light" JSON conversion methods
- Conversion functions return `RichContent`. Saving to json requires running `RichContent.toJSON()` or `toJSONLight()`
- Use integer version number, folder structure updated to v1
- Stripped double quotes from toHTML output
- Exported migration-content.json
- Fixed conversion of `inline` plugin width to `CUSTOM` in schema


oded-soffrin:
as discussed:
- add: createdTimestamp, updatedTimestamp
- light-weight 'fromJson' which only handle Date type (to reduce bundlesize on consumption)

packages/ricos-content/web/src/converters/draft/fromDraft/fromDraft.ts
```diff
@@ -8,7 +8,7 @@ import { getTextNodes } from './getTextNodes';
 import { getEntity, getTextStyle } from './getRicosEntityData';
 import { createParagraphNode, initializeMetadata } from '../../nodeUtils';
 
-export const ensureRicosContent = (content: RichContent | DraftContent) =>
+export const ensureRicosContent = (content: RichContent | DraftContent): RichContent =>
```
lxgreen, line 11: by the way, there's the same `Node_Type !== Node_Type` incompatibility error in wix-data when they use `ensureRicosContent`

packages/ricos-content/web/src/converters/draft/fromDraft/fromDraft.ts
```diff
@@ -8,7 +8,7 @@ import { getTextNodes } from './getTextNodes';
...
   'blocks' in content ? fromDraft(content) : content;
```
lxgreen, line 12: we can implement TS type-guard functions in more reliable way

lxgreen (approved):

mharsat:
whoops I forgot Oded's comment, taking care of it

packages/ricos-content/web/src/converters/draft/fromDraft/fromDraft.ts
mharsat: please elaborate on that, if they define the content's type to be `RichContent` everything should work match typewise (unless a different enum type is used like when dealing with other proto-generated enums)

packages/ricos-content/web/src/converters/draft/fromDraft/fromDraft.ts
mharsat: WDYM?

packages/ricos-content/web/src/converters/draft/fromDraft/fromDraft.ts
mharsat: might need to cast (`as RichContent`). anyway after this PR they'll need to use `fromJSON` on their content before passing it to the `fromDraft` function, and types will be solved the right way

oded-soffrin (approved):

PR #2379, opened 2021-04-18
fix/gallery loader

talevy17:


packages/plugin-gallery/web/src/gallery-component.jsx
```diff
@@ -136,6 +134,7 @@ class GalleryComponent extends PureComponent {
         const item = data && createGalleryItem(data, Date.now().toString());
         uploadBIData && this.props.helpers?.onMediaUploadEnd(uploadBIData, error);
         this.setItemInGallery(item, error, itemIdx);
+        this.onLoad(false);
```
maimonav, line 137: Correct me if I'm mistaken, but this will remove loader after first img/video will upload.

talevy17: Good catch, fixing.

maimonav (approved):
Looks good :)

PR #2380, opened 2021-04-18
fix/delete quote block before atomic block

talevy17:


maimonav:
It might best to consult wix UX regarding this, but I would assume that if you delete a quote, it should change it first to unstyled block, and then if hit backspace again, it should remove the block, and put selection on previous block, like it works when there is text block before quote block.

maimonav (approved):

PR #2383, opened 2021-04-18
[schema] fix link

shaulgo:


packages/ricos-schema/web/src/main/proto/rich_content/common.proto
```diff
@@ -20,13 +20,31 @@ message TextStyle {
...
 message Link {
-    oneof data {
-        string url = 1;
-        string anchor = 2; // link in document
-        string custom_data = 3; // serialized object
+    message _Link {
+        message Rel {
+            google.protobuf.BoolValue nofollow = 1; //Indicates to search engine crawlers that the link should not be followed
+            google.protobuf.BoolValue sponsored = 2; //paid links
+            google.protobuf.BoolValue user_generated_content = 3; //UoU generated link
```
oded-soffrin, line 27:
```suggestion
            google.protobuf.BoolValue ugc = 3; //UoU generated link
```

packages/ricos-schema/web/src/main/proto/rich_content/common.proto
```diff
@@ -20,13 +20,31 @@ message TextStyle {
 }
 
 message Link {
-    oneof data {
-        string url = 1;
-        string anchor = 2; // link in document
-        string custom_data = 3; // serialized object
+    message _Link {
+        message Rel {
```
oded-soffrin, line 24: noreferrer & noopener are not needed?
(taken from [spec](https://docs.google.com/document/d/1Mzz28MNYnIdiuV5G1ql61fY92NcSWVf-NXipehMpmIk/edit#))

packages/ricos-schema/web/src/main/proto/rich_content/common.proto
```diff
@@ -20,13 +20,31 @@ message TextStyle {
...
+        oneof data {
+            string url = 1;
+            string anchor = 2; // id of node in document
+        }
+        Target target = 4;
+        Rel rel = 5;
+    }
+
+    oneof linkData {
+        _Link link = 1;
```
oded-soffrin, line 45: Suggestion: rename _Link message to something like [HTMLAnchorElement](https://developer.mozilla.org/en-US/docs/Web/API/HTMLAnchorElement)

oded-soffrin (commented):
some renaming suggestions, and question about 'noreferrer' & 'noopener' necessity

packages/ricos-schema/web/src/main/proto/rich_content/common.proto
oded-soffrin: also - notice that usage from imagePlugin will be, e.g.:
**imageData.config.link.link.target** 
maybe we should have
HTMLAnchorElement attributes = 1;

so types will match and usage will be **imageData.config.link.attributes.target**

mharsat: Also message names must follow guidelines and that means UpperCamelCase

packages/ricos-schema/web/src/main/proto/rich_content/common.proto
shaulgo: I considered ugc. But why not have a name that explains itself?

packages/ricos-schema/web/src/main/proto/rich_content/common.proto
shaulgo: I can't think of case where they are not "true". So no need to save them. 
If a case arises. Then we can add them in the future.

packages/ricos-schema/web/src/main/proto/rich_content/common.proto
shaulgo: * link.link is definitely not what we want. 
* Let's go with imageData.config.link.[url/anchor/rel/target/custom_data]
* Remove the oneof. So it will be possible in the schema to have custom_data alongside the rest of the link attributes. Although this will probably not be used.
* This should be most readable.

packages/ricos-schema/web/src/main/proto/rich_content/common.proto
oded-soffrin: I believe ugc is a well known abbreviation

shaulgo: Changed it. 
But I did a quick survey in the office, at least here, it's not known at all. 
Yet it doesn't matter much.

shaulgo:
Seems like we are done. Good comments, thanks :)
Matan will continue with the converters.

packages/ricos-content/web/src/converters/draft/fromDraft/convertRicosPluginData.ts
```diff
@@ -146,13 +149,41 @@ const convertButtonData = (data, blockType) => {
...
+}): Link => {
+  const relValues =
+    rel
+      ?.split(' ')
+      .filter(key => ['nofollow', 'sponsored', 'ugc'].includes(key))
+      .map(key => [key, true]) || [];
+  return {
+    anchor,
+    url,
+    rel: relValues.length > 0 ? Object.fromEntries(relValues) : undefined,
```
shaulgo, line 184: relValues = rel?.xxx
This means relValues can be undefined. This will error. Shouldn't ts error on this?

mharsat (approved):

packages/ricos-content/web/src/converters/draft/fromDraft/convertRicosPluginData.ts
mharsat: It's `relValues = rel?.xxx || []` so rel can be undefined, but relValues is always an array

PR #2388, opened 2021-04-21
chore(ricos-content): remove process exit

oded-soffrin:
`process.exit(1)` make the server goes bye bye when using converters function with invalid content
This PR merely replaces the oh-so-harsh command with a more subtle traditional error throwing


mharsat (approved):
With great power comes great responsibility 🕸

PR #2398, opened 2021-04-22
[plugin-image] Avoid exception when image src missing

ramunaspi:
### Issue
In some cases Ricos viewer crashes (along with the rest of the app) when image source in image plugin is undefined. Rest of the code has checks for that, but `isPNG` regex match does not.

### Solution
Added conditional check to avoid that

### Additional info
Yes, it is a one symbol PR :) After this check regex function get `undefined`, and returns false. It is correct, since `undefined` is definitely not a PNG. Why `src` is undefined there is another question, but in any case, it should not crash here.


oded-soffrin:
isPng check should be moved to inner function `getImageSrc` where file_name usage is being checked
https://github.com/wix/ricos/blob/e73141d2c7d146fe18aac53a4e5ca88e2c2e0583/packages/ricos-content/web/src/imageUtils.ts#L110

PR #2412, opened 2021-04-28
[html plugin] new ricos schema

danarish4:


shaulgo (approved):
only add url example to the fixtures

shaulgo:
@mharsat Please take a look

packages/ricos-content/web/src/converters/draft/toDraft/convertDraftPluginData.ts
```diff
@@ -208,7 +209,12 @@ const convertButtonData = (data: Partial<ButtonData> & { button }) => {
 };
 
 const convertHTMLData = data => {
-  delete data.config?.size;
+  const { html, url, config = {} } = data;
+  const srcType = html ? 'html' : 'url';
+  data.srcType = srcType;
+  data.src = html ? html : url;
```
mharsat, line 215: also nice
```suggestion
  data.src = html || url;
```

packages/ricos-schema/web/src/main/proto/rich_content/v1/common.proto
```diff
@@ -65,8 +65,9 @@ message PluginContainerData {
     }
 
     Width width = 1;
-    Alignment alignment = 2;
-    Spoiler spoiler = 3;
+    google.protobuf.Int32Value custom_height = 2;
```
mharsat, line 68: This is a breaking change to field indices in common.proto, better add it as index 4

mharsat (approved):
Amazing! 🚀
Just need a fix in common.proto

packages/ricos-schema/web/src/main/proto/rich_content/v1/common.proto
shaulgo: basa

PR #2462, opened 2021-05-06
[ricos-content] ContentApi: clean undone events

Barackos:


lxgreen (changes requested):
looks good. just don't commit commented code, we have git, so we can restore it even if it's removed.

Barackos:
> link to usage sends to theming docs?

old habits ;)

PR #2459, opened 2021-05-06
[link]: feat/custom anchor scroll

talevy17:

PR #2468, opened 2021-05-06
Allow custom logic for scrolling to anchors

carl-e:
We're using the plugin-link in our editors and views to allow scrolling to anchors within post but our headers aren't the same sizes as the standard wix headers.

We'd like to have a config option `customAnchorScroll` for the LINK plugin that would be called instead of the built in `anchorScroll` and `addAnchorTagToUrl` that's fired in `LinkViewer`'s `handleClick`. That way we can calculate a better set of margins in order to scroll to the right place (and do other custom logic with the url).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,12 +15,16 @@
     Changes that have landed in master but are not yet released.
     Click to see more.
   </summary>
-  
+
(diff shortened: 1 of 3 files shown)


carl-e:
Looks like it'll be handled with #2459

PR #2508, opened 2021-05-13
[twitter-plugin] support twitter plugin by wix-ricos

danarish4:


shaulgo:
1. move to rich content
2. preload
3. api

PR #2536, opened 2021-05-24
[static-toolbar] fix: static toolbar style

maimonav:


packages/editor/web/src/RichContentEditor/Toolbars/StaticToolbar/createStaticToolbar.jsx
```diff
@@ -70,7 +70,7 @@ export default (data = {}) => {
       return editorState;
     },
     Toolbar: props => (
-      <div className={staticToolbarClassName} dir={getLangDir(locale)}>
+      <div className={staticToolbarClassName} dir={getLangDir(locale)} style={{ flex: 0 }}>
```
shaulgo, line 73: 1. why is this needed?
2. The toolbar is also in consumers, can it break them? (is it also used in mobile?)
3. Why inline style and not css class?

maimonav: 1. We need this to limit the size of the toolbar's height, using `flex: 0`,this `flex: 0` used to be in the surge example under `toolbar-wrapper` class name, but since we delete the legacy RichContent components, we delete this also, and then the issue recreated (the toolbar height is really big, it takes more space that it needs to).
2. I think it can only fix the issue and not break anything. Moreover, I'm guessing that consumers that use our static toolbar already fixed this issue by the same css, but I'll check it.
3. You're right, I'll add it to the css class.

shaulgo:
Sounds good.

packages/editor/web/src/RichContentEditor/Toolbars/StaticToolbar/createStaticToolbar.jsx
maimonav: @shaul I've search this usage of static toolbar across our consumers, no one but WixData is using our built in static toolbar from Ricos, `useStaticTextToolbar` prop.
Wix Data is using it, but they also pass `textToolbarContainer`.
We can do the same in our exampleApp, pass container which has the `flex: 0` fix, WDYT?

maimonav: Nevermind, this issue is occurring only in the ExampleApp, it took some time to figure it out, this issue is related to `styles.global.scss`.

shaulgo (approved):

PR #2542, opened 2021-05-25
Get rid of direct usage of i18next in plugins

mbutsykin:

### DIFF
--- a/packages/common/web/src/Utils/withI18n.tsx
+++ b/packages/common/web/src/Utils/withI18n.tsx
@@ -1,5 +1,5 @@
 import React, { PureComponent, Ref, ComponentType } from 'react';
-import { I18nextProvider, translate } from 'react-i18next';
+import { I18nextProvider, translate, Trans } from 'react-i18next';
 import i18n from './i18n';
 import createHocName from './createHocName';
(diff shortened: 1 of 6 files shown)


packages/common/web/src/Utils/withI18n.tsx
```diff
@@ -1,5 +1,5 @@
 import React, { PureComponent, Ref, ComponentType } from 'react';
-import { I18nextProvider, translate } from 'react-i18next';
+import { I18nextProvider, translate, Trans } from 'react-i18next';
```
oded-soffrin, line 2: @mbutsykin if the problem is react-i18next without `translate` , won't it crash here?

mbutsykin: `common` has own `react-i18next^7.0` which has `translate` so this PR simply gets rid of all usages of `react-i18next` not from `common`

mbutsykin: i mean common bundles with own `react-i18next` with strict version. 
so i changed everywhere direct imports of `react-i18next` & `i18next` to use ones from `common`

PR #2543, opened 2021-05-25
[ricos-schema] containerData width refactor

danarish4:


packages/ricos-schema/web/src/generated/proto/wix/rich_content/v1/common.proto
```diff
@@ -50,28 +50,25 @@ message PluginContainerData {
...
     }
 
-    Width width = 1;
-    Alignment alignment = 2;
-    Spoiler spoiler = 3;
-    google.protobuf.Int32Value custom_height = 4;
+    oneof data {
+        Width width = 1;
+        google.protobuf.Int32Value custom_width = 2;
+    }
```
oded-soffrin, line 68: I think we still need a `Width` object here, because now we have either:
container.width = SMALL /  container.custom_width = 23;
if we encapsulate we'll have:
container.width.size = SMALL / container.width.custom = 23 
I think it's clearer

also let's change custom_height to be simetric (container.height.custom)
so we could add more attributes if need in the future (like max_height you suggested)

PR #2586, opened 2021-06-03
[anchor] support anchor from outer editor into inner editors

noamtwix:


packages/editor-common/web/src/Components/AnchorComponents/anchorUtils.js
```diff
@@ -14,13 +14,13 @@ import {
   ACTION_BUTTON_TYPE,
 } from 'wix-rich-content-common';
 
-export const getAnchorableBlocks = editorState => {
+export const getAnchorableBlocks = (editorState, selectedBlock = undefined) => {
```
maimonav, line 17: Is `selectedBlock = undefined` needed? isn't it undefined by default if you don't pass the prop in the calling function?

packages/editor-common/web/src/Components/AnchorComponents/anchorUtils.js
```diff
@@ -14,13 +14,13 @@ import {
...
   const anchorableBlocks = [];
   const typesWithIndexes = {};
 
   const contentState = editorState.getCurrentContent();
   const selection = editorState.getSelection();
-  const selectedBlockKey = selection.getStartKey();
+  const selectedBlockKey = selectedBlock || selection.getStartKey();
```
maimonav, line 23: If I understand correctly, `selectedBlock` is a `selectedBlockKey`, but the var name is taken,
I would've call it `blockKey`, just to emphasize that it's a key, rather than a block, so others won't be confused by it.

maimonav (approved):

packages/editor-common/web/src/Components/AnchorComponents/anchorUtils.js
noamtwix: this one is to avoid ts error without changing the file to ts

packages/editor-common/web/src/Components/AnchorComponents/anchorUtils.js
noamtwix: I don't think that blockKey is better than selectedBlock

packages/editor-common/web/src/Components/AnchorComponents/anchorUtils.js
maimonav: If so, I would suggest to change it to `.ts`, should be quick, there are only methods here.

PR #2671, opened 2021-07-04
[ricos-content] create new embed schema and consider social-embed as …

danarish4:
…embed in the new schema


packages/ricos-content/web/src/converters/plainText/toPlainText/convertNodes.ts
```diff
@@ -105,3 +105,7 @@ export const parseLinkPreview = ({ linkPreviewData }: Node): string => {
   const { url } = linkPreviewData?.link || {};
   return url || '';
 };
+
+export const parseEmbed = ({ embedData }: Node, delimiter?: string): string => {
+  return [embedData?.oembed?.title || '', embedData?.src].filter(Boolean).join(delimiter);
```
danarish4, line 110:
```suggestion
  return [embedData?.oembed?.title, embedData?.src].filter(Boolean).join(delimiter);
```

PR #2680, opened 2021-07-05
[ricos-schema] collapsible-list schema

maimonav:


packages/ricos-schema/web/src/main/proto/wix/rich_content/v1/plugin_collapsible_list.proto
```diff
@@ -0,0 +1,22 @@
...
+package wix.rich_content.v1;
+
+import "wix/rich_content/v1/common.proto";
+import "google/protobuf/wrappers.proto";
+
+message CollapsibleListData {
+    enum ExpandState {
+        FIRST_EXPANDED = 0;
+        EXPANDED = 1;
+        COLLAPSED = 2;
```
oded-soffrin, line 12: thought, maybe change semantic here to make it clearer:
DefaultExpandedItems {
   FIRST = 0;
   ALL = 1;
   NONE = 2;
}

packages/ricos-schema/web/src/main/proto/wix/rich_content/v1/rich_content.proto
```diff
@@ -62,8 +63,10 @@ message Node {
         VERTICAL_EMBED = 18;
         BUTTON = 19;
         COLLAPSIBLE_LIST = 20;
-        TABLE = 21;
-        EMBED = 22;
+        COLLAPSIBLE_LIST_PAIR = 21;
+        COLLAPSIBLE_LIST_PAIR_ITEM = 22;
```
shaulgo, line 67: @oded-soffrin  @maimonav 
Why do this need to be Nodes? This makes them "global" type.
I thought it would something like so:
collapsibleData.items[0].title.nodes
collapsibleData.items[0].content.nodes

maimonav: If we do that, we are breaking the "node in node" structure, don't you think? @shaulgo

maimonav: Similar to the way @mharsat has implemented the LIST_ITEM node.

maimonav: Plus, it gives us the possibility, in the future, to add data to the pair/item, features such as styles and design that are planned for Collapsible phase 2.

shaulgo: 1. We could still add the data in title/content.
2. I don't understand the node in node structure. What are the benefits?

maimonav: I think that the first benefit, is consistency through the schema, instead of saving nodes inside the NodeData, we can avoid it and put it in the nodes array, just like all other Nodes.
And I'm not entirely sure how we are going to render each Node using TipTap, so that can be answered later when we reach that point.

packages/ricos-content/web/src/converters/draft/fromDraft/fromDraft.ts
```diff
@@ -60,12 +60,37 @@ export const fromDraft = (draftJSON: DraftContent): RichContent => {
...
+      pairs.map((pair: CollapsibleListPair) => parseCollapsibleListPair(pair));
+
     if (entityRanges && entityRanges.length) {
       const entity = getEntity(entityRanges[0].key, entityMap);
       if (entity) {
+        const nodes =
+          entity.type === Node_Type.COLLAPSIBLE_LIST
+            ? // eslint-disable-next-line @typescript-eslint/no-explicit-any
+              parseCollapsiblePairs((entity as any).collapsibleListData.pairs)
+            : [];
```
mharsat, line 90: Let's keep `parseAtomicBlock` as lean an generic as possible.
All plugin specific logic should be moved to some designated handler

maimonav: Done.

packages/ricos-schema/web/src/main/proto/wix/rich_content/v1/plugin_collapsible_list.proto
maimonav: @shaulgo @mharsat WDYT?

mharsat: I'm in favor, it makes it easier to understand.
I do suggest changing terminology. We don't handle state in the saved content to `ExpandState` doesn't fit. Default values are related to consumer configs so `DefaultExpandedItems` doesn't work either.
If I understand correctly this setting sets the initial expanded items state per list, so I'd go for `InitialExpandedItems`

packages/ricos-schema/web/src/main/proto/wix/rich_content/v1/rich_content.proto
```diff
@@ -62,8 +63,11 @@ message Node {
         APP_EMBED = 18;
         BUTTON = 19;
         COLLAPSIBLE_LIST = 20;
-        TABLE = 21;
-        EMBED = 22;
+        COLLAPSIBLE_ITEM = 21;
```
lxgreen, line 66: it is breaking change to modify the index of previously existing field. here, embed was 22, now it is 25. schema extension should always be appending to end

packages/ricos-schema/web/src/main/proto/wix/rich_content/v1/rich_content.proto
```diff
@@ -87,7 +91,8 @@ message Node {
         TextData text_data = 18;
         AppEmbedData app_embed_data = 19;
         VideoData video_data = 20;
-        EmbedData embed_data = 21;
+        CollapsibleListData collapsible_list_data = 21;
+        EmbedData embed_data = 22;
```
lxgreen, line 95: also here

lxgreen (changes requested):
please fix the breaking

packages/ricos-schema/web/src/main/proto/wix/rich_content/v1/rich_content.proto
maimonav: Sure, thanks

packages/ricos-schema/web/src/main/proto/wix/rich_content/v1/plugin_collapsible_list.proto
maimonav: @mharsat Good points, agree.

maimonav:
@lxgreen Done, anything else?

bundle-analyzer/bundlesSizesBaseline.json
```diff
@@ -32,10 +32,10 @@
   "ricos-viewer-no-modal": 185,
   "ricos-viewer-with-plugins": 274,
   "rce": 774,
-  "rce-with-plugins": 978,
+  "rce-with-plugins": 1051,
   "ricos-editor": 823,
   "ricos-editor-no-modal": 823,
-  "ricos-editor-with-plugins": 1028,
+  "ricos-editor-with-plugins": 1099,
```
mharsat, line 38: Is this only schema changes?

mharsat (approved):
LGTM 🔥

bundle-analyzer/bundlesSizesBaseline.json
maimonav: Yes.

packages/ricos-content/web/src/converters/draft/fromDraft/fromDraft.ts
```diff
@@ -59,13 +59,36 @@ export const fromDraft = (draftJSON: DraftContent): RichContent => {
...
   };
 
+  const parseCollapsiblePairs = (entity): Node | undefined =>
+    entity?.collapsibleListData?.pairs?.map(
+      (pair: { key: string; title: DraftContent; content: DraftContent }) => ({
+        id: pair.key,
+        type: Node_Type.COLLAPSIBLE_ITEM,
+        nodes: [
+          {
+            id: generateId(),
```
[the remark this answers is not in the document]
maimonav: It's possible, but then we will have two pairs of nodes that contains the same key:
(COLLAPSIBLE_ITEM_TITLE & FIRST NESTED NODE, COLLAPSIBLE_ITEM_BODY & FIRST NESTED NODE)
It's best to keep it unique :)

packages/ricos-content/web/src/converters/html/toHtml/toHtml.ts
```diff
@@ -34,9 +34,12 @@ const createHtmlAttrs = (decoration: Decoration): Record<string, string> => {
 };
 
 const createHtmlTag = (tag: string, child: string, attrs: Record<string, string> = {}): string => {
+  const isFragment = tag === 'fragment';
   const prefix = tag === 'li' ? '\n  ' : '';
   const suffix = ['ul', 'ol'].includes(tag) ? '\n' : '';
-  return `${prefix}<${tag}${attrsToString(attrs)}>${child}${suffix}</${tag}>`;
+  return isFragment
```
lxgreen, line 40: i think the whole suffix/prefix thing is redundant here. why would we try to **format** html while conversion to html? formatting is unrelated to conversion, as it is matter of visual representation vs data structure. Actually, suffix/prefix become `\n` textual nodes in html, which later complicates parsing
@mharsat @maimonav

lxgreen (changes requested):

packages/ricos-content/web/src/converters/html/toHtml/toHtml.ts
maimonav: If I understand you correctly, you're referring to formatting html in **general** and not regarding this changes.
What I've done is to allow to parse collapsible to html (IMO, we need to define the html representation of every plugin first), but this can be done later and I'm sure that we will have changes in the future.
Regarding your subject, I'm not really familiar with the usage of this converters, who is using it and how? @lxgreen @mharsat

lxgreen: right, my question is addressed to @mharsat, and it is not related directly to these changes. 

just wrote it down now, before i forget it.

packages/ricos-schema/web/src/main/proto/wix/rich_content/v1/plugin_collapsible_list.proto
shaulgo: ya this looks good 
InitialExpandedItems {
  FIRST = 0;
  ALL = 1;
  NONE = 2;
}

maimonav: Great, it's implemented already :)

bundle-analyzer/bundlesSizesBaseline.json
shaulgo: This pr is small. Where does this originate from?

packages/ricos-content/web/src/converters/html/toHtml/toHtml.ts
mharsat: I agree that formatting can be separated from the conversion, especially if it's growing.
Initially the idea was to match the existing HTML baseline, like we do with all converters (as we get it from RichText). We can have a different task for separating this

packages/ricos-content/web/src/converters/draft/toDraft/convertDraftPluginData.ts
```diff
@@ -29,6 +30,7 @@ import {
 import { WRAP, NO_WRAP } from '../../../consts';
 import { ComponentData, FileComponentData } from '../../../types';
 import { parseLink } from '../../nodeUtils';
+import { toDraft } from './toDraft';
```
mharsat, line 33: I think the bundle issue comes from here.
`convertDraftPluginData.ts` is part of the `toDraftData` bundle - a leaner version containing only the plugin conversion functions, instead of the whole `toDraft` logic. It's used in every `create<PLUGIN_NAME>Data.ts` file, that might explain why we see the increase only in the `rce-with-plugins` & `ricos-editor-with-plugins` bundles, and not in the schema bundle or the non-plugin rce/ricos-editor bundles.

@maimonav please consider if we can work around this somehow in the plugins, this will rise again when table schema is ready.
Also, a new `toDraftData` bundle in our analyzer can help prevent this in the future

@lxgreen @shaulgo

bundle-analyzer/bundlesSizesBaseline.json
```diff
@@ -32,10 +32,10 @@
   "ricos-viewer-no-modal": 185,
   "ricos-viewer-with-plugins": 274,
   "rce": 774,
-  "rce-with-plugins": 978,
+  "rce-with-plugins": 1051,
```
shaulgo, line 35: @oded-soffrin When you feel better. We need to discuss.

packages/ricos-content/web/src/converters/draft/toDraft/convertDraftPluginData.ts
mharsat: Updating that @maimonav is trying to solve by dynamically loading the `toDraftData` bundle. It can generally improve performance and reduce even the existing bundle size. All that is left is to verify that editor commands can load this data asynchronously

bundle-analyzer/bundlesSizesBaseline.json
mharsat: Notice the `toDraft` discussion above ⬆️

packages/ricos-content/web/src/converters/draft/toDraft/convertDraftPluginData.ts
maimonav: @mharsat Correct, but after discussing it with Shaul, we are not sure it's the right move.

maimonav:
@oded-soffrin @shaulgo @mharsat Please take a look again, bundle size grew a bit, but way better than before.

mharsat:
> @oded-soffrin @shaulgo @mharsat Please take a second look, bundle size grew a bit, but way better than before.

I think the editor's bundle size should not grow for all kinds of rich content just because of 2 very advanced plugins.
If dynamic loading is off the table we should consider a different approach that narrows this increase only to contents/consumers that use collapsible. Could be achieved by importing `toDraft` from the collapsible package, or maybe by using loadable.

shaulgo:
> > @oded-soffrin @shaulgo @mharsat Please take a second look, bundle size grew a bit, but way better than before.
> 
> I think the editor's bundle size should not grow for all kinds of rich content just because of 2 very advanced plugins.
> If dynamic loading is off the table we should consider a different approach that narrows this increase only to contents/consumers that use collapsible. Could be achieved by importing `toDraft` from the collapsible package, or maybe by using loadable.

As I understand it this growth is mainly the support of inner rce. If so, it's acceptable by me at the moment. We will definitely do more work on this in the future.
@maimonav please rebase the table schema pr on this one. And let's see how much the bundle grows when we have both collapsible and table.

PR #2695, opened 2021-07-07
Finalize Plugins Schema

maimonav:

### DIFF
--- a/packages/ricos-schema/web/src/main/proto/wix/rich_content/v1/common.proto
+++ b/packages/ricos-schema/web/src/main/proto/wix/rich_content/v1/common.proto
@@ -39,6 +39,12 @@ message Link {
     google.protobuf.StringValue custom_data = 5; // Serialized object, used for custom/external link panel
 }
 
+enum Alignment {
+    CENTER = 0;
+    LEFT = 1;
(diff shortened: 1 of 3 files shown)


packages/ricos-schema/web/src/main/proto/wix/rich_content/v1/plugin_divider.proto
```diff
@@ -5,7 +5,7 @@ package wix.rich_content.v1;
 import "wix/rich_content/v1/common.proto";
 
 message DividerData {
-    enum Type {
+    enum LineStyle {
```
mharsat, line 8: Nice!

packages/ricos-schema/web/src/main/proto/wix/rich_content/v1/common.proto
```diff
@@ -39,6 +39,12 @@ message Link {
     google.protobuf.StringValue custom_data = 5; // Serialized object, used for custom/external link panel
 }
 
+enum Alignment {
```
mharsat, line 42: Preferably, enums should be part of a message, otherwise they will be global between different files (including dependency proto files)

PR #2696, opened 2021-07-07
fix tiptap translation

justame:


justame:
@lxgreen @shaulgo

packages/common/web/src/Utils/withI18n.tsx
```diff
@@ -87,4 +87,27 @@ export default <T, P>(
   return React.forwardRef<T, P>((props, ref) => <I18nWrapper {...props} forwardedRef={ref} />);
 };
 
-export { translate, Trans };
+const RicosTranslate = ({ children, locale, localeResource }) => {
```
shaulgo, line 90: withTranslaiton from react-i18next, is not good for us?

PR #2699, opened 2021-07-08
[ricos-schema] add title to video schema

danarish4:

### DIFF
--- a/packages/ricos-content/web/statics/json/migratedFixtures/migration-content.json
+++ b/packages/ricos-content/web/statics/json/migratedFixtures/migration-content.json
@@ -6,7 +6,7 @@
       "nodes": [
         {
           "type": "TEXT",
-          "key": "7jebh",
+          "key": "",
           "nodes": [],
           "textData": {
             "text": "Lorem ipsum dolor sit amet",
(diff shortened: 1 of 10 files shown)


mharsat (approved):
Nice!

packages/ricos-content/web/statics/json/migratedFixtures/migration-content.json
```diff
@@ -6,7 +6,7 @@
       "nodes": [
         {
           "type": "TEXT",
-          "key": "7jebh",
+          "key": "",
```
shaulgo, line 9: is this supposed to remove the keys?

shaulgo: @mharsat The keys are making it much harder to find the diff. 
We should find a way to check this without changing the key every time. 
I'm not familiar enough with the test. But maybe, changing it to an ordered key (1,2,3). Or empty keys, but with a check that they do exist.

mharsat: Prose doesn't support a `key` attribute on `text` nodes, so we made it an empty string.
This should also create a lot less key changes when updating baselines, since most of them were text keys

PR #2717, opened 2021-07-14
getImgSrc removed from converters

lxgreen:

### DIFF
--- a/packages/ricos-content/web/src/converters/plainText/toPlainText/convertNodes.ts
+++ b/packages/ricos-content/web/src/converters/plainText/toPlainText/convertNodes.ts
@@ -1,5 +1,4 @@
 import { Node, Node_Type } from 'ricos-schema';
-import { getImageSrc } from '../../../imageUtils';
 import { LINK_TYPE } from '../../../consts';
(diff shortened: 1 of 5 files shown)


packages/ricos-content/web/src/converters/plainText/toPlainText/convertNodes.ts
```diff
@@ -56,15 +55,7 @@ export const parseImage = async (
...
-  const imageUrlOptions = Object.assign(
-    {
-      imageType: 'highRes',
-      requiredQuality: 90,
-    },
-    width && { requiredWidth: width },
-    height && { requiredHeight: height }
-  );
-  let url: string = getImageSrc({ file_name: src?.custom }, undefined, imageUrlOptions);
+  let url = `https://static.wixstatic.com/media/${src?.custom?.replace('media/', '')}`;
```
mharsat, line 58: Maybe this function could be common

mharsat (approved):
Great 🚀

packages/ricos-content/web/src/converters/plainText/toPlainText/convertNodes.ts
lxgreen: agree

PR #2758, opened 2021-07-21
[ricos-content -> tiptap-editor] converters relocation

Barackos:


packages/ricos-content/web/lib/converters.ts
```diff
@@ -6,6 +6,5 @@ export {
   TO_RICOS_NODE_TYPE,
 } from '../src/converters/draft';
 export { fromPlainText, toPlainText } from '../src/converters/plainText';
-export * from '../src/converters/tiptap';
```
Barackos, line 9: @lxgreen Am I breaking someone by doing this?

PR #2769, opened 2021-07-25
[html-plugin] hotfix(HtmlComponent): validate src to prevent crash

oded-soffrin:
@shaulgo FYI 
(fix for https://github.com/wix-private/auto-views/issues/3352)


packages/plugin-html/web/src/HtmlComponent.jsx
```diff
@@ -123,7 +123,7 @@ class HtmlComponent extends Component {
           />
         )}
 
-        {srcType === SRC_TYPE_URL && isValidUrl(src) && (
+        {srcType === SRC_TYPE_URL && src && isValidUrl(src) && (
```
shaulgo, line 126: Also possible to add it to the isValidUrl. falsy value is definitely not valid url.

PR #2770, opened 2021-07-25
[ui-components] refactor: button/textinput/inputwithlabel

tamrat-bay:


danarish4 (approved):

packages/ui-components/web/src/Components/InputWithLabel.jsx
```diff
@@ -15,8 +15,8 @@ class InputWithLabel extends Component {
   }
 
   handleOnChange = e => {
-    const { onChange, getTarget } = this.props;
-    onChange(getTarget ? e.target : e.target.value);
+    const { onChange, getEvent } = this.props;
+    onChange(getEvent ? e : e.target.value);
```
shaulgo, line 19: @tamrat-bay @danarish4 you like this?

danarish4: You're right, but in this case, the external package we use (Downshift) expects to get event object (otherwise it fails on Downshift source code because of trying to get event.target..)

tamrat-bay: Screen Shot 2021-07-25 at 14 20 16

PR #2774, opened 2021-07-25
[ricos-schema] map schema

maimonav:


packages/ricos-schema/web/src/generated/proto/wix/rich_content/v1/plugin_map.proto
```diff
@@ -16,6 +16,13 @@ message MapData {
   MapSettings mapSettings = 2;
 }
 
+enum MapType {
+  ROADMAP = 0;
+  SATELINE = 1;
```
shaulgo, line 21:
```suggestion
  SATELLITE = 1;
```

maimonav: Thanks

packages/ricos-content/web/src/converters/draft/toDraft/convertDraftPluginData.ts
```diff
@@ -281,7 +281,7 @@ const convertMapData = data => {
   data.mapSettings.isZoomControlShown = zoomControl;
   data.mapSettings.locationDisplayName = locationName;
   data.mapSettings.zoom = initialZoom;
-  data.mapSettings.mode = mapType;
+  data.mapSettings.mode = mapType?.lowerCase();
```
mharsat, line 284: Typo - should be `toLowerCase`
Maybe you can add basic typing on `convertMapData`'s `data` object, it helped us in the past with cases like this

mharsat (changes requested):
Can you add comments to the fields in the proto file? (like we have in `rich-content.proto`)

mharsat (approved):

PR #2794, opened 2021-07-29
[editor] fix genFragment of draftConvertFromHtml

maimonav:


packages/editor/web/src/RichContentEditor/utils/draftConvertFromHtml.js
```diff
@@ -313,9 +300,6 @@ function genFragment(
       return getEmptyChunk();
     }
 
-    if (text.trim() === '' && inBlock !== 'code-block') {
-      return getWhitespaceChunk(inEntity);
```
shaulgo, line 317: This looks like it should be when not in code block? 
What does this code do?

maimonav: This `getWhitespaceChunk` function returns chunk of single space, no matter if the text was more than one space.
They did it do differ between code-block & not, but in our case - we don't want to alter the spaces, what we get is what we paste.
That is all draft-convert code.

shaulgo:
sounds good

PR #2807, opened 2021-08-02
add capabilities to editor commands (indent, line spacing, save state…

noamtwix:
…, save selection)


packages/common/web/src/types/editorCommandsType.ts
```diff
@@ -92,11 +94,13 @@ interface PluginsDataMap {
   [IMAGE_TYPE]?: any;
 }
 
-interface DecorationsDataMap {
+export interface DecorationsDataMap {
```
maimonav, line 97: Is it needed?

packages/editor/web/src/RichContentEditor/EditorCommands.ts
```diff
@@ -215,6 +257,11 @@ export const createEditorCommands = (
     hasLinkInSelection: () => hasLinksInSelection(getEditorState()),
     getLinkDataInSelection: () => getLinkDataInSelection(getEditorState()),
     getSelectedData: () => getEntityData(getEditorState()) || {},
+    getBlockSpacing,
+    saveEditorState,
+    loadEditorState,
+    saveSelectionState,
+    loadSelectionState,
```
maimonav, line 264: I know it has no difference, but can you please change it to look like all other functions in the editorState object? (declare the function inside the object)

packages/editor/web/src/RichContentEditor/EditorCommands.ts
```diff
@@ -169,6 +177,13 @@ const deleteDecorationsMapFuncs = {
   [RICOS_TEXT_HIGHLIGHT_TYPE]: setHighlightColor,
 };
 
+const lineHeight = 'line-height';
+const spaceBefore = 'padding-top';
+const spaceAfter = 'padding-bottom';
```
maimonav, line 182: It might be better to define it in a different file and export it, or if you only use it in the `getBlockSpacing`, I would use it there and not declare it as a global var.

packages/common/web/src/types/editorCommandsType.ts
```diff
@@ -115,6 +119,11 @@ export interface EditorCommands {
   getLinkDataInSelection: () => any;
   getSelectedData: () => any;
   getPluginsList: (settings?: { isRicosSchema?: boolean }) => PluginsList;
+  getBlockSpacing: () => any;
```
maimonav, line 122: Can you create a type for it?

maimonav (changes requested):

packages/common/web/src/types/editorCommandsType.ts
```diff
@@ -42,7 +44,7 @@ import {
 import { MentionData } from './pluginTypes';
 import { TextAlignment, InlineStyle } from './commonTypes';
 
-type ColorType = typeof RICOS_TEXT_COLOR_TYPE | typeof RICOS_TEXT_HIGHLIGHT_TYPE;
+export type ColorType = typeof RICOS_TEXT_COLOR_TYPE | typeof RICOS_TEXT_HIGHLIGHT_TYPE;
```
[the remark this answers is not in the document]
noamtwix: yes

packages/common/web/src/types/editorCommandsType.ts
noamtwix: yes

packages/common/web/src/types/editorCommandsType.ts
noamtwix: no

packages/common/web/src/types/editorCommandsType.ts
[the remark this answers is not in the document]
maimonav: Ok, but this file is the type of the EditorCommands, if you need this type elsewhere, you should move it to type file in this package (common), and export if from there, and use wherever you need.

packages/common/web/src/types/editorCommandsType.ts
maimonav: Same here :)

packages/common/web/src/types/editorCommandsType.ts
maimonav: Ok, why?

maimonav (approved):
Looks good, just a kind reminder - create a type for `getBlockSpacing` when you use it in the toolbar :)

PR #2828, opened 2021-08-05
[editor-common, common, gallery] fix: link data values

maimonav:
1. Gallery modal fix using `uiSettings`.
2. Link values save fix.


packages/common/web/src/test/linkConverters.spec.ts
```diff
@@ -29,9 +29,9 @@ describe('Test link converters', () => {
     expect(converted).toEqual(relObject);
   });
 
-  it('should convert target undefined to undefined', () => {
+  it('should convert target undefined to true', () => {
```
shaulgo, line 32: This is very unintuitive with the name of the function.

shaulgo (changes requested):

packages/plugin-gallery/web/src/components/gallery-controls/gallery-image-settings.js
```diff
@@ -74,12 +74,16 @@ class ImageSettings extends Component {
       visibleRightArrow,
       uiSettings,
```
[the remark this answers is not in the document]
maimonav: I'm not sure if it always was that way, but it has been like that for a while.

packages/common/web/src/test/linkConverters.spec.ts
maimonav: This is the default result, if target isn't defined then by default it's true.

shaulgo: I don't love it. But this should will be removed when we migrate completely to new schema?

maimonav: No, this is needed for the link panels.

shaulgo (approved):

PR #2837, opened 2021-08-09
[storybook] upgrade to v6

Barackos:


oded-soffrin:
On the effort of trying to remove securty warning driven by older storybook version.
this PR provide the infra of new storybook v6, but due to version syntax change all the stories have to be migrated manually

PR #2839, opened 2021-08-09
[tooltip] refactor: tooltip style

tamrat-bay:


packages/ricos-common/web/src/themeStrategy/generators/palette.spec.ts
```diff
@@ -23,10 +23,16 @@ describe('Palette', () => {
     settingsActionColorTuple: undefined,
     focusActionColor: undefined,
     focusActionColorTuple: undefined,
+    borderColorFallback: undefined,
   };
 
-  const wixExpected = {
+  const tooltipExpected = {
```
tamrat-bay, line 29: Better name
borderFallbackExpected

PR #2874, opened 2021-08-18
[table] override sideToolbar onClick config when use it from table

danarish4:


shaulgo (approved):
I hope you know this is bad coding. But that's what I wanted. 
We will change it when we move completely to the new toolbars

PR #2876, opened 2021-08-19
[ricos-content] improve(html convertors): preserve formatting

Barackos:


packages/ricos-content/web/src/converters/html/fromHtml/core/rules.ts
```diff
@@ -80,6 +93,41 @@ export const strongEmUToDecoration: Rule = [
...
+});
+
+const toBackgroundData = (color: string): Decoration['colorData'] => ({
+  background: color,
+});
+
+const toColorDecoration = (ctx: Context, el: Element) => (colorData: ColorData) =>
+  ctx.addDecoration(Decoration_Type.COLOR, { colorData }, el);
+
+export const colorStyleToTextColor: Rule = [
```
lxgreen, line 107: it is preferred to get separate rules for different styles, in sake of flexibility, so text-color and text-highlight have separate rules.

The rule naming convention is `htmlTagOrAttr-To-RicosNodeOrElementType`

packages/ricos-content/web/src/converters/html/fromHtml/core/rules.ts
```diff
@@ -80,6 +93,41 @@ export const strongEmUToDecoration: Rule = [
...
+  concatApply(MonoidAll)([hasTag('span'), hasStyleFor('color')]),
+  context => (el: Element) =>
+    pipe(
+      el,
+      getStyle,
```
lxgreen, line 112: this util already exists, no need to implement again ;)

packages/ricos-content/web/src/converters/html/toHtml/toHtml.ts
```diff
@@ -44,13 +61,37 @@ const attrsToString = (attrs: Record<string, string>): string =>
...
+const eitherColorDecoration = (d: Decoration): E.Either<Decoration, Decoration> =>
+  d.type === Decoration_Type.COLOR ? E.right(d) : E.left(d);
+
+const normalizeDecorations: (decorations: Decoration[]) => Decoration[] = flow(
+  A.map(flow(eitherColorDecoration, E.fold(A.of, expandColorDecoration))),
+  A.flatten
+);
+
+const decorateTextElement = ({ text, decorations }: TextData): string => {
+  const normalizedDecorations = normalizeDecorations(decorations);
```
lxgreen, line 86: for simplicity and reuse, it is better to get 1:1 relation between tag and attribute. the `normalizeDecorations` splits color decorations into array of decoration-per-style.

Currently, this is a quick fix. We'll rewrite `toHtml` to be more generic like its counterpart

lxgreen (changes requested):
see my changes and comments; some Translatable tests have to be updated

PR #2900, opened 2021-08-31
polls-plugin: remove settings (`siteToken`) from contentState

mbutsykin:


packages/plugin-social-polls/web/src/createPollPlugin.ts
```diff
@@ -1,3 +1,4 @@
+import { omit } from 'lodash';
```
maimonav, line 1: Please revert :)

mbutsykin: sure)

PR #2946, opened 2021-09-29
fix fullscreen a11y issues

talevy17:


package.json
```diff
@@ -30,7 +30,6 @@
     "build:dev": "lerna run --stream --ignore 'wix-rich-content-@(test-env|example|storybook)' build:dev",
     "build:core": "npm run build:dev -- --scope 'ricos-content' && npm run build:dev -- --scope 'wix-rich-content-@(common|editor|viewer)'",
     "prewatch": "lerna run build --scope wix-rich-content-*common* && lerna run build --scope wix-rich-content-editor && lerna run build --scope ricos-common",
-    "prebuild": "node scripts/mergeTranslations",
```
maimonav, line 33: It is on purpose?

packages/fullScreen/web/src/InnerFullScreen.jsx
```diff
@@ -101,7 +112,7 @@ export default class InnerFullscreen extends Component {
         tabIndex={0}
         className={styles.close}
         onClick={this.onClose}
-        onKeyDown={this.onClose}
+        onKeyDown={e => e.keyCode === 13 && this.onClose()}
```
maimonav, line 115: I suggest to extract keyCode === 13 to a method

packages/fullScreen/web/src/InnerFullScreen.jsx
```diff
@@ -113,7 +124,7 @@ export default class InnerFullscreen extends Component {
   onFullscreenToggleKeyDown = event => {
     if (event.key === 'Escape') {
       this.onClose();
-    } else {
+    } else if (event.keyCode === 13) {
```
maimonav, line 127: Check comment in line 115. :)

package.json
talevy17: mistake reverting

packages/fullScreen/web/src/InnerFullScreen.jsx
```diff
@@ -121,14 +123,6 @@ export default class InnerFullscreen extends Component {
     );
   };
 
-  onFullscreenToggleKeyDown = event => {
-    if (event.key === 'Escape') {
```
maimonav, line 114: Isn't this use case is missing now?

talevy17: @maimonav nope, its already a registered event listener in `document` and works either way.

maimonav (approved):

PR #2954, opened 2021-10-03
[editor] support shortcuts with key code

noamtwix:


packages/editor/web/src/RichContentEditor/keyBindings.js
```diff
@@ -68,7 +68,11 @@ function getModifiers(e) {
 
 function getCommandByShortcut(shortcut, bindingMap) {
   const commands = bindingMap
-    .filter(mapped => mapped.key === shortcut.key && isEqual(mapped.modifiers, shortcut.modifiers))
+    .filter(
+      mapped =>
+        (mapped.key === shortcut.key && isEqual(mapped.modifiers, shortcut.modifiers)) ||
+        (mapped.key === shortcut.keyCode && isEqual(mapped.modifiers, shortcut.modifiers))
```
oded-soffrin, line 74: a) logic wise, it can be simplified 
(a && c) || (b && c) === (a || b) && C 

b) why are we comparing key to keyCode? can't we change in the structure to keyCode as well so we could compare mapped.keyCode === shortcut.keyCode ?

oded-soffrin (changes requested):
small comment

PR #2956, opened 2021-10-04
[link-preview-popover]fix: display popover on link plugin only

danarish4:


packages/common/web/src/consts.js
```diff
@@ -1,3 +1,5 @@
 export const ADD_PLUGIN_LINK_BI = 'addPluginLink';
 export const WEB_ADDRESS_CATEGORY = 'web_address';
 export const ANCHOR_CATEGORY = 'section';
+export const LINK_VIEWER_DATA_HOOK = 'linkViewer';
+export const ANCHOR_VIEWER_DATA_HOOK = 'linkViewer';
```
danarish4, line 5:
```suggestion
export const ANCHOR_VIEWER_DATA_HOOK = 'anchorViewer';
```

PR #2960, opened 2021-10-05
[ricos-content] improve(fromDraft): handle unsupported block & decoration types

mharsat:


packages/ricos-content/web/src/converters/draft/fromDraft/fromDraft.ts
```diff
@@ -19,43 +19,47 @@ export const fromDraft = (draftJSON: DraftContent): RichContent => {
...
-          nodes.push(node);
-          parseBlocks(nextIndex);
-          break;
-        case BlockType.Unstyled:
-          nodes.push(parseTextBlock(block));
-          parseBlocks(index + 1);
-          break;
-        default:
-          throw Error(`ERROR! Unknown block type "${block.type}"!`);
+      if (UNSUPPORTED_BLOCK_TYPES.includes(block.type)) {
```
lxgreen, line 22: hmm. this approach provides us control over "what is allowed and what's not"; however, there's downside -- we have to maintain this as more unsupported cases arise. we cannot know in advance how much cases there are in the wild, and AFAIU we're gonna allow them all at the end. so, maybe it is worth to allow any unsupported value to be ignored, but this whole "allow unsupported fields"  mechanism has to be opt in? like a param to fromDraft:
```js
fromDraft(content, opts = { ignoreUnsupportedValues = false })
```

lxgreen (commented):
there's an open question to discuss

packages/ricos-content/web/src/converters/draft/fromDraft/fromDraft.ts
mharsat: OK we can make that opt in.
Consumers that use the getContent service will need us to extend that too in this case

lxgreen: which `getContent` service do you refer to? there are multiple `getContent`s

mharsat: RichContentService

lxgreen: you mean the other `fromXXX` convertors also may get such option to skip invalid content rather throw error?

mharsat: I meant [RichContentService](https://github.com/wix-private/wix-ricos/blob/master/serverless/rich-content-services/src/RichContentService.ts)

PR #3008, opened 2021-10-17
chore(LinkPanelWrapper): component tests

oded-soffrin:
Adding component tests for link panel wrapper 
@maimonav as dicussed after this change - https://github.com/wix/ricos/commit/dc0bd63f30b395d7e6189e38b163568444e17129 

Serves as an example of much needed and missing component tests in our beloved project, as a way to develop and protect from future changes

@shaulgo FYI


oded-soffrin:
@noamtwix it seems the new toolbars is duplicating this file:
https://github.com/wix/ricos/blob/master/packages/toolbars-new/web/src/modals/link/LinkComponents/LinkPanelWrapper.tsx

this is troubling since fixes made to origin files after duplication will be lost with new toolbars, what's the plan here?

PR #3046, opened 2021-10-30
[link + tiptap] plugin relocation + styling

Barackos:


packages/plugin-link/web/src/tiptap.ts
```diff
@@ -0,0 +1,225 @@
...
+
+      defaultOptions: {
+        openOnClick: true,
+        linkOnPaste: true,
+        HTMLAttributes: { link: {} },
+        ...defaultOptions,
+      },
+
+      addAttributes() {
+        // return LinkData.fromJSON({});
```
Barackos, line 95: @lxgreen We don't have a link json data

PR #3074, opened 2021-11-07
[ricos-content] improve(extract API): extraction by type

Barackos:
2 updates:
- `getText` is out of the `adapter` entry, as it is unrelated to it
- `getText` now returns a joined `string` object
- refactoring


packages/ricos-content/web/src/RicosContentAPI/extract-by-type.ts
```diff
@@ -9,9 +9,10 @@ type PluginData = NonNullable<
 export const getText = (content: RichContent) =>
   extract(content.nodes)
     .filter(node => node.type === Node_Type.TEXT)
-    .map(node => node.textData)
-    .filter(data => !!data?.text)
-    .get() as TextData[];
+    .map(node => node.textData?.text)
+    .filter(data => !!data)
+    .get()
+    .join('\n');
```
lxgreen, line 15: `join('\n')` is very application domain logic, why not just array? also, this code ignores the fact that new line symbol is [platform dependent](https://stackoverflow.com/questions/10864486/node-js-constant-for-platform-specific-new-line).

packages/ricos-content/web/src/RicosContentAPI/extract-by-type.ts
```diff
@@ -9,9 +9,10 @@ type PluginData = NonNullable<
 export const getText = (content: RichContent) =>
   extract(content.nodes)
     .filter(node => node.type === Node_Type.TEXT)
-    .map(node => node.textData)
-    .filter(data => !!data?.text)
-    .get() as TextData[];
+    .map(node => node.textData?.text)
```
lxgreen, line 12: all this could be written as 
```
extract(content.nodes).map(({ textData }) => textData?.text || '').get()
```

explanation: `textData` exists on TEXT nodes only, and `get()` filters falsy results internally.

packages/ricos-content/web/src/RicosContentAPI/extract-by-type.ts
```diff
@@ -1,6 +1,6 @@
-import { RichContent, Node_Type, TextData, Node, Media } from 'ricos-schema';
+import { RichContent, Node_Type, Node, Media } from 'ricos-schema';
```
lxgreen, line 1: what's the point to include this in the content API core level? it is more relevant to an application like velo

lxgreen (changes requested):

packages/ricos-content/web/src/RicosContentAPI/extract-by-type.ts
Barackos: no real reason. convinced

packages/ricos-content/web/src/RicosContentAPI/extract-by-type.ts
Barackos: consider it as an explicit api to do things. Velo is one consumer. But we also have services that may want it, so it is unrelated to velo specifically

lxgreen: then let's call it like `application-wrappers` instead of velo, but keep them separately

Barackos: You mean, a lib with the name `application-wrappers`? or separate libs for each use case?

lxgreen: a lib with the name `application-wrappers`

PR #3075, opened 2021-11-07
[toolbars-new] fix: disable link when atomic block is selected

noamtwix:


maimonav (approved):
Great, please don't forget to update the changelog :)

PR #3076, opened 2021-11-07
[example] new toolbar - default in example + add dedicated button in …

noamtwix:
…settings


oded-soffrin:
Like
but let's make some order
Order should be:
- Mobile 
- Use New Content (prose mirror effort)
- and than New toolbar

(we can push Native Mobile mock below - I believe we don't use it very often)

PR #3080, opened 2021-11-08
[tiptap] add insert, set and update node core commands

danarish4:

### DIFF
--- a/packages/tiptap-editor/web/src/core/commands/insertNode.ts
+++ b/packages/tiptap-editor/web/src/core/commands/insertNode.ts
@@ -1,0 +1,19 @@
+import { RawCommands } from '@tiptap/core';
+import { generateId } from 'ricos-content';
+import { toTiptap } from '../../converters';
+
+declare module '@tiptap/core' {
+  interface Commands<ReturnType> {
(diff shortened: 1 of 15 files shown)


packages/tiptap-editor/web/src/core/commands/insertNode.ts
```diff
@@ -0,0 +1,19 @@
...
+       * Insert a node.
+       */
+      insertNode: (type: string, attrs: Record<string, unknown>) => ReturnType;
+    };
+  }
+}
+
+export const insertNode: RawCommands['insertNode'] = (type, attrs = {}) => ({ commands }) => {
+  const id = generateId();
+  return commands.insertContent({ type, attrs: { id, ...toTiptap(attrs) } });
```
oded-soffrin, line 18: we can use the id we got in the attrs instead of generating twice, right ?

oded-soffrin, line 18:
```suggestion
const { id, rest } = toTiptap(attrs)
return commands.insertContent({ type, attrs: { id: id || generateId(), ...rest } });
```

PR #3112, opened 2021-11-14
[image] fix a11y on enter

oded-soffrin:


oded-soffrin:
Notice image expand stopped working when tabbing into it and hitting enter, this fixed it.
(couldn't remove the other onKeyDown because it has onClick)
@maimonav WDYT ?

maimonav (approved):
Please update changelog :)

PR #3157, opened 2021-11-23
[devex] ricos-content improve build speed

Barackos:


oded-soffrin:
@justame should we keep this one? (even for future referenec)

justame:
Yes , shaul might need it

PR #3159, opened 2021-11-24
[video-modal/ui-components] refactor: move upload modal into ui-components

tamrat-bay:


packages/ui-components/web/src/Modals/MediaUploadModal.tsx
```diff
@@ -0,0 +1,66 @@
...
+}) => {
+  const Styles = mergeStyles({ styles, theme });
+
+  const uploadSection = (
+    <div>
+      <div className={Styles.mediaUploadModal}>
+        <input
+          id={id}
+          type="file"
+          accept="video/*"
```
talevy17, line 38: just a thought - not necessarily on you rn
maybe we should align our accepted video formats with MediaManager or ones that `ReactPlayer` can play for sure, so we won't try uploading formats MM or ReactPlayer can't handle.

talevy17 (changes requested):
Looks solid, make sure to check that the theme <=> cssOverride doesn't break you

PR #3168, opened 2021-11-29
[tiptap] check latest version android issues

oded-soffrin:


oded-soffrin:
doesn't seem to fix bold/link issues :(

PR #3176, opened 2021-11-29
[tiptap]: unique-id extension & fix selection callbback

danarish4:


packages/tiptap-editor/web/src/extensions/extension-unique-id/unique-id.tsx
```diff
@@ -0,0 +1,69 @@
...
+export const createUniqueId = (): RicosFunctionalExtension => ({
+  type: 'extension',
+  createExtensionConfig: () => {
+    return {
+      name,
+      priority: 1,
+      addGlobalAttributes() {
+        return [
+          {
+            //TODO: take node extensions names from extensions config
```
justame, line 16: let's talk about this list.
if I develop a third-party extension, would my extension get an id?

PR #3181, opened 2021-11-30
[general] wire settings box-shadow and border-radius

tamrat-bay:


packages/plugin-headings/web/statics/styles/panelStyles.scss
```diff
@@ -22,10 +23,11 @@
         background-color: rgba(0, 0, 0, .05);
       }
     }
+  }
```
tamrat-bay, line 26: remove

PR #3209, opened 2021-12-05
[ricos-editor] feat: Ricos Essentials

maimonav:


examples/main/src/Components/ToolbarWithEssentials.tsx
```diff
@@ -0,0 +1,113 @@
+import React from 'react';
+import { IRicosEditorEssentials } from 'wix-rich-content-common';
```
lxgreen, line 2: why does it reside in `common`?  it is for editor only, right? the `common` should contain stuff that is relevant for both editor and viewer (i know we made it just a pile of garbage for last couple of years, but still -- let's start making _seder_ in _balagan_)

packages/common/web/src/types/IRicosEditorEssentials.ts
```diff
@@ -0,0 +1,23 @@
...
+export interface IRicosEditorCommands {
+  insertNode: (node: Node) => string;
+  updateNode: (id: string, node: Node) => boolean;
+  deleteNode: (id: string) => boolean;
+}
+
+export interface IRicosEditorModel {
+  getNodesBy: (predicate: (node: Node) => boolean) => Node[];
+  getSelectedNodes: () => Node[];
+  hasDecoration: (type: Decoration_Type) => boolean;
```
lxgreen, line 18: what scope `hasDecoration` API applies to? single node or the whole content? both `getSelectedNodes` and `getNodesBy` refer the whole content, and this one?

packages/common/web/src/types/IRicosEditorEssentials.ts
```diff
@@ -0,0 +1,23 @@
...
+}
+
+export interface IRicosEditorState {
+  hasFocus: () => boolean;
```
lxgreen, line 22: same question regarding the scope here

packages/common/web/src/types/IRicosEditorEssentials.ts
```diff
@@ -0,0 +1,23 @@
+import { Node, Decoration_Type } from 'ricos-schema';
+
+export interface IRicosEditorEssentials {
```
lxgreen, line 3: so the whole `essentials` API is exposed, and also each one of `model`, `commands` and `state` exposed? i think we should stick with either exposing only the `essentials` or just its components. 

By the way, the difficulty to find a good name for these `essentials` is just a symptom of its redundancy as it has no added value beyond being a wrapper to 3 well-defined elements (`model`, `commands` and `state`). 🤔

packages/ricos-content/web/src/consts.ts
```diff
@@ -123,13 +123,21 @@ export const RICOS_EXTERNAL_LINK_TYPE = 'ricos-external-link'; // TODO: is this
 export const RICOS_EXTERNAL_MENTION_TYPE = 'ricos-external-mention'; // TODO: is this type needed?
 
 // tiptap plugins types
+export const TIPTAP_BUTTON_TYPE = 'button';
```
lxgreen, line 126: _unrelated to these changes only:_

the tiptap node types are used by tiptap only (so even the tiptap converters are not defined in `ricos-content` since they are internal). We don't want to expose any tiptap internals here, so we should not see any `tiptap` outside packages that have `tiptap` in their names. So, all these consts ideally are just name fields inside tiptap extension definitions, that's it. However, due to centralized approach that we have currently with all the commands, converters, etc. we have to store them elsewhere. Please move them into `ricos-tiptap-types`, for now. Later, we'll decompose this centralization, hopefully even schema gonna be distributed.

packages/ricos-editor/web/src/RicosEditor.tsx
```diff
@@ -13,7 +13,13 @@ import {
   getBiCallback as getCallback,
 } from 'ricos-common';
 import { DraftContent } from 'ricos-content';
-import { RichContentEditor, RichContentEditorProps } from 'wix-rich-content-editor';
+import {
+  RichContentEditor,
+  RichContentEditorProps,
+  DraftEditorCommands,
+  DraftEditorModel,
+  DraftEditorState,
```
lxgreen, line 21: can we just load these dynamically based on experiment? if yes, we can do the same trick with the whole RCE

lxgreen (changes requested):
please see my comments, let's discuss them some time

Barackos (commented):
Awesome!
The `essentials` naming still doesn't feel right for me, however I have no better suggestion. I'm in favor of adopting Alex's note regarding this naming

packages/common/web/src/types/IRicosEditorEssentials.ts
maimonav: `hasDecoration` is applied on the selected nodes.

packages/common/web/src/types/IRicosEditorEssentials.ts
maimonav: `hasFocus` on the Editor.

examples/main/src/Components/ToolbarWithEssentials.tsx
maimonav: It should reside in `ricos-common`, or if we would like to open `ricos-editor-common`, and then, `IRicosEditorEssentials` can be consumed from `ricos-common`, IMO that should be the appropriate way to use.
But in the current implementation, package `ricos-common` is dependent on `editor`, so refactor needs to be done to make a seder in the balagan :)

packages/common/web/src/types/IRicosEditorEssentials.ts
maimonav: That is only the Type representation, the exposed `essentials` is accessible through the editor's ref.

packages/ricos-editor/web/src/RicosEditor.tsx
maimonav: For sure, it needs some adjustments, but yeah, we can refactor it to dynamically load all RCE related components.

packages/common/web/src/types/IRicosEditorEssentials.ts
```diff
@@ -0,0 +1,23 @@
...
+
+export interface IRicosEditorCommands {
+  insertNode: (node: Node) => string;
+  updateNode: (id: string, node: Node) => boolean;
+  deleteNode: (id: string) => boolean;
+}
+
+export interface IRicosEditorModel {
+  getNodesBy: (predicate: (node: Node) => boolean) => Node[];
+  getSelectedNodes: () => Node[];
```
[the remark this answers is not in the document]
maimonav: Good notes, indeed!

At first, my thoughts regarding `Model` module are that this module should expose every content related (Ricos schema), but it might be better to move `getSelectedNodes` into `State` module since it's data based selection, and sure, selection is state related.
BTW, `hasDecoration` in the current implementation (and most use cases if not every use case) is applied on selected nodes, so it might also be a good idea to move it to the `State` module as well.

packages/common/web/src/types/IRicosEditorEssentials.ts
lxgreen: then it should be reflected somehow (in its name?)

packages/common/web/src/types/IRicosEditorEssentials.ts
lxgreen: editor ref is disputable way to expose API. In the past, we already saw its limitations, e.g. publish API usage in blog, where Publish button and editor are placed in different layers, so we ended with really ugly ref forwarding. On other hand, the commands require a live editor instance. This could be achieved if we'd initialize such API with editor instance argument like:

```ts

const commands = configureEditorCommands(editor);
commands.addNode(...);

```

 This way, it is easy to move/consume this API, it is easy to test it (just pass any mock implementation you want), easy to distinguish among underlying implementations (draft/tiptap/ckeditor).

maimonav: I see, I like this approach.
But what is the Type of editor? is it editor's ref?
If so, it will have the same issue as you described with publish button and editor.

lxgreen: ideally, it is pure logic entity like Editor instance of tiptap, which is easily passed everywhere, including EditorContent component that actually renders the content. In our case, the draft functionalities are coupled to editor component, and we don't want to invest time in it, as it is gonna be deprecated. Can you find a way to create a logic draft-based entity that is bound to draft?

examples/main/src/Components/ToolbarWithEssentials.tsx
maimonav: Make a package `ricos-types`

PR #3236, opened 2021-12-09
[toolbars-new] fix: color picker buttons in static toolbar when colla…

noamtwix:
…psed


packages/editor-common/web/src/Utils/inlineStyleUtils.ts
```diff
@@ -100,6 +100,17 @@ export const removeCurrentInlineStyle = (
 ) => {
   const selection = editorState.getSelection();
   const currentStyles = getSelectionStyles(editorState, styleSelectionPredicate);
```
talevy17, line 102: `currentStyles` is defined twice (before and inside the if) its a bit confusing, maybe move it below the if statement?

talevy17 (approved):

PR #3242, opened 2021-12-10
perf(ricos-editor): limit initial content change calls

oded-soffrin:
Due to performance issues, slack convo https://wix.slack.com/archives/C02Q305P07Q/p1639151433032100

other than boolean value reverse (@Barackos tell me if I missed something, but isn't it the _poet meaning_?)
this PR goal is to have a single expensive `getContentTraits` call


Barackos:
@oded-soffrin Oops 😄 yes the value reverse is definitely right. Should've respected the poet a little more

oded-soffrin:
> @oded-soffrin Oops 😄 yes the value reverse is definitely right. Should've respected the poet a little more

And regarding saving extra calls, seems right?

Barackos:
> And regarding saving extra calls, seems right?

Yes, in fact - following your change, you can also remove the `if` clause inside `onInitialContentChanged`, as it is always true

packages/ricos-editor/web/src/RicosEditor.tsx
```diff
@@ -225,17 +225,18 @@ export class RicosEditor extends Component<RicosEditorProps, State> {
 
   onInitialContentChanged = () => {
     const { initialContentChanged } = this.state;
```
Barackos, line 227:
```suggestion
```

packages/ricos-editor/web/src/RicosEditor.tsx
```diff
@@ -225,17 +225,18 @@ export class RicosEditor extends Component<RicosEditorProps, State> {
...
-    if (initialContentChanged) {
+    if (!initialContentChanged) {
       const contentId = this.getContentID();
       this.getBiCallback('onContentEdited')?.({ version: Version.currentVersion, contentId });
```
Barackos, line 230:
```suggestion
    const contentId = this.getContentID();
    this.getBiCallback('onContentEdited')?.({ version: Version.currentVersion, contentId });
```

packages/ricos-editor/web/src/RicosEditor.tsx
```diff
@@ -225,17 +225,18 @@ export class RicosEditor extends Component<RicosEditorProps, State> {
...
-      this.setState({ initialContentChanged: false });
+      this.setState({ initialContentChanged: true });
     }
```
Barackos, line 232:
```suggestion
    this.setState({ initialContentChanged: true });
```

Barackos (approved):

packages/ricos-editor/web/src/RicosEditor.tsx
Barackos: Sorry for the 3 code suggestions - Github doesn't allow 1 suggestion when some lines are "deleted" 😄

lxgreen:
@Barackos are you sure it is necessary to store the `initialContentChanged` on state? Could be just a class property to save re-renders, right?

Barackos:
> @lxgreen: Are you sure it is necessary to store the `initialContentChanged` on state? Could be just a class property to save re-renders, right?

True, moved to class variable

PR #3247, opened 2021-12-12
[ricos-content] fromDraft: fix crash on no-data plugins

oded-soffrin:
Input coming from @andrius-n while trying to use fromDraft
seems like blog invalid content is crashing, we can just skip invalid plugins when converting

@lxgreen WDYT?

also notice there are many empty ID (screenshot attached) in converted content, is it valid output?
image


packages/ricos-content/web/src/converters/draft/fromDraft/__tests__/fromDraft.spec.ts
```diff
@@ -25,6 +27,7 @@ import {
 } from 'ricos-schema';
 import { convertBlockDataToRicos } from '../convertRicosPluginData';
 import { IMAGE_TYPE, FILE_UPLOAD_TYPE, WRAP } from '../../../../consts';
+import { convertDocumentStyleDecorationTypes } from '../../toDraft/decorationParsers';
```
maimonav, line 30: Looks like it's not being used

maimonav (approved):

PR #3251, opened 2021-12-13
[tiptap] move code block extension to ricos side

danarish4:


packages/tiptap-editor/web/src/components/RicosNode.tsx
```diff
@@ -11,8 +11,10 @@ export const RicosNode = ({ Component, tiptapNodeProps }) => {
     ...tiptapNodeProps,
   };
   return (
-    <NodeViewWrapper>
-      <Component {...componentProps} />
-    </NodeViewWrapper>
+    Component && (
```
lxgreen, line 14: no need in this protection, the component-less RicosNodeExtensions are handled differently from react-based ones, if you define it as a RicosExtension (see my next comment for details)

packages/tiptap-extensions/web/src/codeBlock.ts
```diff
@@ -0,0 +1,211 @@
...
+       */
+      toggleCodeBlock: (attributes?: { language: string }) => ReturnType;
+    };
+  }
+}
+
+export const backtickInputRegex = /^```(?<language>[a-z]*)?[\s\n]$/;
+export const tildeInputRegex = /^~~~(?<language>[a-z]*)?[\s\n]$/;
+
+export const CodeBlock = Node.create<CodeBlockOptions>({
```
lxgreen, line 30: this whole extension should be defined as RicosExtension, and moved to the code-block plugin `tiptap.ts`. Please take a look at heading [extension](https://github.com/wix/ricos/blob/master/packages/plugin-headings/web/src/tiptap.ts) for reference, it walked the same path from tiptap => headings plugin.

lxgreen (changes requested):
please see the comments and align this extension with other migrated extensions

PR #3254, opened 2021-12-13
[monitoring] fix: contentId not sent in some events

Barackos:
## Fix
Several events did not contain `contentId` field. 2 reasons:
- Wasn't included in `onPluginAction` (inside `RichContentEditor.tsx`)
- Occurred in events triggered within RicosEditorModal, which is outside the scope of `RichContentEditor` and `RichContentViewer`, but the business logic related to adding this field existed only on their scope. Therefore, moved all of that business logic to `RicosEngine`.

## Refactoring
- renamed `getContentID` to `getContentId`
- `RicosEngine` is now responsible for adding `version` & `contentId` into events (instead of `RichContentEditor` & `RichContentViewer`
- `uuid` from `wix-rich-content-common` now has its own entry lib. This is to reduce bundle size for `ricos-common`


examples/main/shared/editor/Editor.tsx
```diff
@@ -82,10 +80,7 @@ export default class Editor extends PureComponent<ExampleEditorProps> {
 
   initEditorProps() {
     /* eslint-disable no-console */
-    const onPluginAction: OnPluginAction = async (
-      eventName: EventName,
-      params: PluginEventParams
-    ) => console.log(eventName, params);
+    const onPluginAction: OnPluginAction = async (...args) => console.log(...args);
```
lxgreen, line 83: degraded type?

packages/common/web/src/types/biCallbacksTypes.ts
```diff
@@ -1,7 +1,7 @@
 import { ToolbarType } from './toolbarEnums';
 import { OnPluginAction } from './pluginsBiCallbacksTypes';
 import { getContentSummary } from '../Utils/contentAnalytics';
-interface biCallbackParams {
+export interface biCallbackParams {
```
lxgreen, line 4: naming -- interface should be in CapitalCase

packages/ricos-common/web/src/utils.ts
```diff
@@ -17,3 +26,91 @@ export function getBiCallback<T extends keyof BICallbacks>(key: T): BICallbacks[
   const { children, _rcProps } = this.props;
   return children?.props.helpers?.[key] || _rcProps?.helpers?.[key];
 }
+
+export const applyGenericParams = (getContentId: () => string | undefined) => props => {
```
lxgreen, line 30: whose props are these `props`? do we already have their type definition? if yes -- it is worth using

lxgreen (changes requested):

packages/ricos-common/web/src/utils.ts
Barackos: We'll have to get back to it. Technically these are `RichContentProps`, which are later destructed into `RicosModal` props, however `RicosModal` is not aware of them, causing build errors

PR #3261, opened 2021-12-14
[general] add BI to link modal in new toolbar

noamtwix:


packages/common/web/src/types/pluginsBiCallbacksTypes.ts
```diff
@@ -30,3 +33,5 @@ export type OnPluginAction = <K extends keyof EventsParamsMappers>(
   eventName: K,
   params: EventsParamsMappers[K]
 ) => void;
+
+export type OnAddPluginLink = (data: AddLinkData, plugin_id: string) => void;
```
Barackos, line 37: The `plugin_id` naming is confusing & not descriptive enough. I think something like `triggeredBy` would better explain its purpose.

Also, these are the only possible options (see [TextLinkModal.jsx](https://github.com/wix/ricos/blob/68dd04aa1b4ea14f5c7ebb35b1a189dc7279bd6d/packages/editor/web/src/RichContentEditor/Toolbars/TextLinkModal.jsx#L109-L116))
```suggestion
export type OnAddPluginLink = (data: AddLinkData, triggeredBy: 'TEXT' | 'IMAGE' | 'UNSUPPORTED') => void;
```

Barackos (approved):

PR #3280, opened 2021-12-19
[plugin-common] fix: plugin toolbar position in rtl

noamtwix:


shaulgo (approved):
I'm not completely sure what's going on. But if you are fine with it then great. We are always looking for a quick fix.

PR #3310, opened 2021-12-25
[*] typescript@^4.0.0

Barackos:


package.json
```diff
@@ -213,7 +213,7 @@
     "ts-jest": "^26.4.4",
     "ts-node": "^9.0.0",
     "ts-proto": "1.72.0",
-    "typescript": "^3.8.3",
+    "typescript": "^4.0.0",
```
lxgreen, line 216:
```suggestion
    "typescript": "^4.5.0",
```

lxgreen: let's get to the edge

packages/editor-common/web/src/Utils/isiOS.ts
```diff
@@ -2,6 +2,7 @@ export function isiOS() {
   return (
     typeof navigator !== 'undefined' &&
     /iPad|iPhone|iPod/.test(navigator.userAgent) &&
-    !window.MSStream
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    !(window as any).MSStream
```
lxgreen, line 6: this one is old -- window.MSStream fails on IE11, there's better [alternative](https://stackoverflow.com/a/9039885)

lxgreen (commented):
see my comments

PR #3311, opened 2021-12-26
[schema] fix: gallery mobile options

talevy17:


packages/ricos-content/web/src/converters/draft/consts.ts
```diff
@@ -222,6 +222,7 @@ export type DraftGalleryStyles = {
   cubeRatio?: number;
   isVertical?: boolean;
   numberOfImagesPerRow?: number;
+  m_numberOfImagesPerRow?: number;
```
oded-soffrin, line 225: why not mobileNumberOfImages?

oded-soffrin (approved):

packages/ricos-content/web/src/converters/draft/consts.ts
talevy17: in draft styles we usually use the PG namings, no reason to change it now I guess

PR #3383, opened 2022-01-13
chore(ts): remove ignore lines

oded-soffrin:
Noticed in @noamtwix we presented code with @ts-ignore, which kinda hurt my eyes 
I've checked why we have those lines and see it's not indeed (or at least this PR attempt to check it)


noamtwix:
@lxgreen told me that he will check this.
what do you think?

lxgreen:
hmm the tiptap-editor didn't build for me when i moved the extensions out to `tiptap-extensons` package. so i added those `ts-ignore` lines. if it is ok, then fine. me thought was is to duplicate command declarations into `tiptap-editor` so they will be known there.

lxgreen:
ok i see it is still failing. i'll try the declarations trick today

oded-soffrin:
I wonder why TS didn't shut on me on the IDE. Closing

PR #3396, opened 2022-01-16
[image] add aspect ratio to img tag

danarish4:


packages/plugin-image/web/src/image-viewer.tsx
```diff
@@ -249,8 +250,8 @@ class ImageViewer extends React.Component<ImageViewerProps, ImageViewerState> {
         onError={this.onImageLoadError}
         onLoad={fadeIn ? e => this.onImageLoad(e.target) : undefined}
         ref={fadeIn ? this.imageRef : this.preloadRef}
-        width={width}
-        height={height}
+        width={width || componentData.src.width}
+        height={height || componentData.src.height}
```
oded-soffrin, line 254: I think this might cause unwanted behavior - what if we send width only?

(if width & height isn't supplied, than we should use `componentData.src.width` & `componentData.src.height` to keep aspect ratio)

oded-soffrin:
Looks great other than the small comment
Tried to see the behavior in example app and still saw content jump, but i'm not sure that's the right test...

PR #3408, opened 2022-01-19
[image-plugin] preload image width by size config

danarish4:
Consider the size of the image (taken from `componentData.config`) when loading preload image.

In the case of `small` size, we almost always make a profit (we set the image width to 350px - I notice Blog override this class and set 360px width but it is pretty close)

In case of `fullWidth` size, it is smarter guess


oded-soffrin (approved):
:

PR #3416, opened 2022-01-23
[editor] insert atomic plugins in new line

oded-soffrin:
I want to add a RCE prop that will allow to add atomic plugins in new line instead of overriding existing plugins for propsals
(later on, we can control this private prop by experiment on ricos level)

not sure how to pass it in the context down below


danarish4:
> 

Not sure this is the best practice but this code is going to die anyway

oded-soffrin:
> 
> Not sure this is the best practice but this code is going to die anyway

awesome 🚀

danarish4 (approved):

PR #3464, opened 2022-02-03
[vertical-embed] fix: dynamically change base url

danarish4:
Following [this](https://github.com/wix-private/wix-ricos/pull/2090) PR


packages/plugin-vertical-embed/web/src/convertDuration.ts
```diff
@@ -0,0 +1,19 @@
+import { isNumber } from 'lodash';
```
danarish4, line 1: exported from utils file for organization

packages/plugin-vertical-embed/web/src/verticalsDataBuilder.ts
```diff
@@ -0,0 +1,36 @@
...
+const contentTypeMapGetter = {
+  [verticalEmbedProviders.booking]: getBookingContent,
+  [verticalEmbedProviders.event]: getEventContent,
+  [verticalEmbedProviders.product]: getProductContent,
+};
+
+export const verticalsDataBuilder = (type, data, t, changeBaseUrl) => {
+  const { imageSrc, pageUrl, imageWidth, imageHeight } = data;
+  return {
+    url: changeBaseUrl ? changeBaseUrl(pageUrl) : pageUrl,
```
oded-soffrin, line 30: alternatively - 
```suggestion
export const verticalsDataBuilder = (type, data, t, changeBaseUrl = () => {}) => {
  const { imageSrc, pageUrl, imageWidth, imageHeight } = data;
  return {
    url: changeBaseUrl(pageUrl),
```

also. - how do you know whether this code is working? a unit-test on verticalDataBuilder will be good here

oded-soffrin (changes requested):
comments in the PR
also - there must be some TS we need to change when adding more props to vertical embed config, right?

oded-soffrin (approved):
Pagaz

PR #3467, opened 2022-02-04
[ricos-content] split migrateSchema entry

danarish4:


lxgreen (approved):
looks great! 

i'm surprized it didn't affect any bundle size.

PR #3474, opened 2022-02-06
[media-plugins] fix: throw generic error on file reader error

talevy17:


oded-soffrin:
going over the code i wanted to ask - why do we promisify this ?
```
const fileReader = file => {
  return new Promise(resolve => {
    resolve(URL.createObjectURL(file));
  });
};
```

also, how did you test?
to reproduce the issue I had to connect my phone and enable USB debugging
I guess we can do the same when the example app will be live

talevy17:
> going over the code i wanted to ask - why do we promisify this ?
> 
> ```
> const fileReader = file => {
>   return new Promise(resolve => {
>     resolve(URL.createObjectURL(file));
>   });
> };
> ```
> 
> also, how did you test? to reproduce the issue I had to connect my phone and enable USB debugging I guess we can do the same when the example app will be live

I'm not sure about the promise, intuitively I think its unnecessary.
About the testing, I just attempted reading a corrupt file, not exactly a one on one equivalent case but figured its similar

packages/plugin-commons/web/src/Base/createBaseMediaPlugin.jsx
```diff
@@ -43,33 +44,35 @@ const handleUploadStart = (
...
+            getComponentData,
+            data,
+            error,
+            onUploadFinished,
+            itemPos,
+            fileType
+          );
+        });
+      })
+      .catch(() => commonPubsub.set('onMediaUploadError', { key: MediaUploadErrorKey.GENERIC }));
```
oded-soffrin, line 75: let display the error
```suggestion
      .catch(e => {
          console.error('handleUploadStart error', e);
          commonPubsub.set('onMediaUploadError', { key: MediaUploadErrorKey.GENERIC }));
      }
```

talevy17: done

oded-soffrin (approved):

PR #3476, opened 2022-02-06
[viewer] perf: include PNG & mobile

oded-soffrin:


danarish4 (commented):
I'm not sure what `removeUsm` does, but imageUrl.preload always calc imageSrc with `removeUsm:false` and imageUrl.highres depended on the experiment - is that acceptable?

maimonav:
Yes, `removeUsm` is a config prop by `ImageClientApi` to remove unsharp mask, and we only use `ImageClientApi` for highres image url's.

maimonav:
For preload & qualityPreload we construct it's url manually.
Side note:
We might need to think if this legacy code should be moved to wix-ricos & we probably can refactor it so we won't need to construct it's url, and use the ImageClientApi instead.
@danarish4 @oded-soffrin

maimonav (approved):

PR #3481, opened 2022-02-06
[ricos-types]

oded-soffrin:
WIP


packages/ricos-common/web/src/biCallbacks/mediaUploadBI.ts
```diff
@@ -1,4 +1,4 @@
-import type { MediaUploadError } from 'wix-rich-content-common';
+import type { MediaUploadError } from 'ricos-content';
```
danarish4, line 1: Is there a reason why this type doesn't sit in `ricos-types` package?

packages/ricos-common/web/src/index.ts
```diff
@@ -5,4 +5,4 @@ export { RicosEngine } from './RicosEngine';
 export { default as themeStrategy } from './themeStrategy/themeStrategy';
 export { default as localeStrategy } from './localeStrategy/localeStrategy';
 export { DRAFT_EDITOR_PROPS } from './consts';
-export type { DraftContent, RicosContent } from 'wix-rich-content-common';
+export type { DraftContent, RicosContent } from 'ricos-content';
```
danarish4, line 8: Why not ricos-type?

danarish4 (changes requested):
MOSHLAM :)
Except for the fact I didn't understand why `DraftContent, RicosContent, MediaUploadError` are not inside ricos-types package

oded-soffrin:
@danarish4 they were already in `ricos-content` so they shouldn't be an issue for shahar
I bet there are tons more types we can move, this is just to unblock

danarish4 (approved):

shahardavid:
This PR os for removing `wix-rich-content-common` from `ricos-common`?

PR #3486, opened 2022-02-07
fix fromHtml/ck-editor: empty list item protection

lxgreen:


lxgreen:
> Looks good, I suggest to add a unit test for this issue.
don't worry, we have plenty of them.
there's `migration-failures.json` and `crashes.json` files with 200+ test cases.

PR #3488, opened 2022-02-07
[video] fix facebook duration data in mobile

shaulgo:


packages/plugin-video/web/src/video-component.jsx
```diff
@@ -52,6 +52,17 @@ class VideoComponent extends React.Component {
     if (!this.state.isLoaded && !this.props.componentData.tempData) {
       this.setState({ isLoaded: true });
     }
+    this.durationFix();
+  };
+
+  durationFix = () => {
```
oded-soffrin, line 58: should we limit this workaround for isMobile only?

shaulgo: completely changed since. I'm not sure what would be the best choice. Every source can work differently. And also change over time. I thought about limiting it to mobile facebook. But decided better to have it for all.

PR #3494, opened 2022-02-09
[theme] ricos font size inhertiance

oded-soffrin:


danarish4 (approved):
We just need to keep in mind this is breaking change for whoever consumes us through the open-source.
Do we want to announce it somewhere?

oded-soffrin:
> We just need to keep in mind this is breaking change for whoever consumes us through the open-source. Do we want to announce it somewhere?

The only meaning that on OOTB usage, instead of Helvetica they will get the inherited font in the editor. It worth mentioning in ricos-updates but I don't think it's breaking

PR #3504, opened 2022-02-10
[video-plugin] fix: insert video crash the editor

danarish4:


danarish4:
> gather that the crash we saw happen during conversion does it make sense to add error handling?
We absolutely have to handle errors here, this doesn't make sense that the editor will crash

> Did the convert happens on the Native side or Web side?
Happens on both sides.
Shahar convert to ricos schema before he calls the editor commands
And we convert from-to draft inside tiptap `draftHoc` extension

maimonav:
@danarish4 @oded-soffrin Just saw this, looks good 🔥 
In general, errors handling is great, but I believe that these conversion errors should not happen in the first place, I will add a dedicated task for it.

PR #3507, opened 2022-02-13
[audio/schema] revert - audio-schema

tamrat-bay:


maimonav (approved):
Please merge it & release a version ( since @oded-soffrin just released a version with a wip version of audio schema) and we don't want to have breaking changes :)
And afterwards, open a PR of the revert of the revert (Recursive something 😃 ) so we can go over it :)

PR #3513, opened 2022-02-14
[ui-components] fix settings secondary color

tamrat-bay:


oded-soffrin:
image
Should we address all the other 757575 in ricos? not many left  🔢

packages/plugin-commons/web/src/Icons/HashtagIcon.jsx
```diff
@@ -21,7 +21,7 @@ const HashtagIcon = props => (
       <g
         id="Text-Formatting-Toolbar-"
         transform="translate(-914.000000, -4635.000000)"
-        fill="#757575"
+        fill="#646464"
```
oded-soffrin, line 24: how will that work when we'll override secondary-text-color ?
can we use fill="currentColor" like in other icons in this folder?

tamrat-bay: Yes you are right,
I moved the styling to the icons css file

PR #3521, opened 2022-02-15
[editor] settings modals use editor commands

danarish4:


packages/plugin-social-polls/web/src/components/settings/SettingsModal.jsx
```diff
@@ -125,7 +125,7 @@ export class SettingsModal extends Component {
     } = this.props;
     const { languageDir } = this.context;
 
-    const componentData = this.modalsWithEditorCommands
+    const componentData = this.modalBaseActionHoc
```
danarish4, line 128:
```suggestion
    const componentData = this.modalsWithEditorCommands
```

PR #3525, opened 2022-02-16
[ricos schema] add audio-schema 

tamrat-bay:


oded-soffrin:
can be closed, right ?

PR #3552, opened 2022-02-21
[internal-tiptap] remove defensive code from richContentAdapter inser…

danarish4:
…tBlock


packages/tiptap-editor/web/src/components/RichContentAdapter/RichContentAdapter.tsx
```diff
@@ -95,7 +95,7 @@ export class RichContentAdapter implements TiptapAPI {
             .insertContent([{ type: 'paragraph' }, { type, attrs }, { type: 'paragraph' }])
             .command(({ tr, commands }) => {
               const nodesWithPos = findNodeById(tr, id);
-              const { pos } = nodesWithPos?.[0] || {};
+              const { pos } = nodesWithPos[0];
```
oded-soffrin, line 98: cant nodesWithPos be empty array?
const { pos } = nodesWithPos[0]
=> const { pos } = [][0]
=> const { pos } = undefined
=> Uncaught TypeError: Cannot destructure property 'pos' of 'undefined' as it is undefined.

justame: @oded-soffrin we first create a node with a specific ID and then look for it. This is a synchronization operation, meaning the node will always be found and the array will never be empty.

danarish4: We don't expect it to be an empty array (it never should be) - and if it happens, we want to know (easy to debug)
I will add try & catch

justame: If you know for sure that the node should exist, neither try-catch is needed

oded-soffrin: what happen if insertContent failing to insert node? it won't reach said code ?

danarish4: it will reach this code.
The purpose was to know easily if that kind of error occurred
I added console.error and I think this is enough

PR #3554, opened 2022-02-21
[list-item](fix): list-item content change

justame:


packages/tiptap-extensions/web/src/list-item.ts
```diff
@@ -15,8 +15,8 @@ export const createListItem = (): RicosExtension => ({
     },
 
     // Note: this should remain paragraph until draft-js is supported
-    content: 'paragraph bulletedList listItem',
-
+    content: 'paragraph bulletedList* listItem*',
```
danarish4, line 18:
```suggestion
        content: '(paragraph|bulletedList*|listItem*)+',
```

danarish4: 'paragraph bulletedList* listItem*' 
This means that content must have exact one paragraph at the top, and one or more bulletedList below and one or more listItem below 

I'm not sure this is what we want to do.. we want to say "list item can include paragraph/list item/ bullet list"

PR #3555, opened 2022-02-21
[themeTypes.ts] CustomStyles API expanded

tamrat-bay:


packages/tiptap-extensions/web/src/statics/styles.scss
```diff
@@ -11,15 +11,30 @@
...
+  line-height: $ricos-code-block-line-height;
+  padding: $ricos-code-block-padding;
+  margin: $ricos-code-block-margin;
+  white-space: pre-wrap;
+
+  .renderedInTable & {
+    margin: 0;
+  }
+}
+
```
danarish4, line 28: need to move to code-block package
replace this `.code` class 
https://github.com/wix/ricos/blob/master/packages/plugin-code-block/web/statics/styles/code-block.scss#L14

PR #3556, opened 2022-02-21
[tiptap] uniqueId optimisation: set id only if needed

danarish4:


oded-soffrin (commented):
If this PR was to remove redundant `setNodeMarkup` then it looks solid to me
does it have any negative side effects BTW?

danarish4:
> If this PR was to remove redundant `setNodeMarkup` then it looks solid to me does it have any negative side effects BTW?

It does not work :(
I'm considering whether to investigate what the reason is or leave it that way for now

PR #3559, opened 2022-02-22
[ui-components] add edit icon

tamrat-bay:


oded-soffrin:
Are all the 5 edit icons are in use ?
image

tamrat-bay:
> Are all the 5 edit icons are in use ? image
> 
> @oded-soffrin looks like that apart the one in gallery they are all in use.

At some point, the icons should be available in the UI-components

PR #3567, opened 2022-02-27
Toolbars v3 color picker buttons

noamtwix:


packages/toolbars-v2/web/src/resolvers/draftResolvers.spec.ts
```diff
@@ -342,4 +344,24 @@ describe('draft resolvers', () => {
       jest.resetModules();
     });
   });
+
+  describe('getTextColorInSelectionResolver', () => {
+    const wixRichContentEditorExported = require('wix-rich-content-editor');
+    it('should check if the function getColor from "wix-rich-content-editor" is called', () => {
+      const spy = jest.spyOn(wixRichContentEditorExported, 'getColor');
+      getTextColorInSelectionResolver.resolve(editorState);
+      expect(spy).toBeCalled();
```
justame, line 353: getTextColorInSelectionResolver and getHighlightColorInSelectionResolver check the same thing.
it's not enough to check if a function was called it's important to know which arguments it received. 
please change to - toBeCalledWith(...).

PR #3576, opened 2022-03-02
[ricos-editor] tiptap detach commands poc (translator & repo)

maimonav:


packages/toolbars-v2/web/src/components/RicosTiptapToolbar.tsx
```diff
@@ -6,23 +6,26 @@ import { RicosToolbar } from '../RicosToolbar';
...
+  experiments?: AvailableExperiments;
 }
 interface RicosToolbarWrapperState {}
 
 class RicosToolbarWrapper extends Component<RicosToolbarWrapperProps, RicosToolbarWrapperState> {
   toolbar: RicosToolbar | null = null;
 
   componentDidMount() {
-    const { content, editorCommands } = this.props;
+    const { content, editorCommands, experiments } = this.props;
```
maimonav, line 23: use Context

maimonav: It appears more changes need to be done in RicosToolbarWrapper/Tiptap in order to use the context, so I'll leave it like that for now

PR #3579, opened 2022-03-03
spoiler tiptap extension priority

lxgreen:


danarish4 (approved):

danarish4:
Maybe we need to manage the extension's priority in one place? 
We need to ensure spoiler extension will get higher priority than all regular extensions

justame (approved):
Looks great

PR #3581, opened 2022-03-06
commit #8000 GET!

lxgreen:


packages/tiptap-editor/web/src/patch-extensions.ts
```diff
@@ -1,16 +1,17 @@
...
+import * as M from 'fp-ts/Monoid';
 import * as S from 'fp-ts/string';
+import { deepLog } from 'ricos-content';
 import type { RicosMarkExtension, RicosNodeExtension } from 'ricos-tiptap-types';
 import { getUnsupportedNodeConfig } from 'wix-rich-content-plugin-unsupported-blocks';
 import {
-  createUniqueId,
   createTextAlign,
   createTextDirection,
+  createUniqueId,
```
justame, line 12: Does the order matter?

lxgreen: here it does not matter

PR #3597, opened 2022-03-09
[media-plugins] add loading percentage support

danarish4:


oded-soffrin:
loved the storybook and example app examples 
If i understand correctly, there's a new command `setLoadingPercentage` works on draft only, so only whoever is using tiptap can use it (mobile)
my question though - since loading percent event is triggering after loading start, does it mean the fake loader will start running until the first time it gets loading percentage ? or we have to signal somehow not to use fake loader ?

danarish4:
> loved the storybook and example app examples If i understand correctly, there's a new command `setLoadingPercentage` works on draft only, so only whoever is using tiptap can use it (mobile) my question though - since loading percent event is triggering after loading start, does it mean the fake loader will start running until the first time it gets loading percentage ? or we have to signal somehow not to use fake loader ?

This API works only on Tiptap for now (if we want it to work also for Draft without putting a lot of effort it means loadingPercentage will sit on the content)

The consumer who calls the initial `insertBlock` commands must add `loadingPercentage: 0` to the data (currently they only add `loading: true`) in order to not display a fake loader until loadingPercentage starts to change.

danarish4:
> > loved the storybook and example app examples If i understand correctly, there's a new command `setLoadingPercentage` works on draft only, so only whoever is using tiptap can use it (mobile) my question though - since loading percent event is triggering after loading start, does it mean the fake loader will start running until the first time it gets loading percentage ? or we have to signal somehow not to use fake loader ?
> 
> This API works only on Tiptap for now (if we want it to work also for Draft without putting a lot of effort it means loadingPercentage will sit on the content)
> 
> The consumer who calls the initial `insertBlock` commands must add `loadingPercentage: 0` to the data (currently they only add `loading: true`) in order to not display a fake loader until loadingPercentage starts to change.


After giving it a second thought, I realize it might be confusing to send it once through the data and once through the appropriate command.
We should guide them to chain commands (or we can create an additional API for the combination `insertBlockWithLoadingProgress`)

```
let nodeId;
editor
.chain()
.command(({commands}) => (nodeId = insertBlock(type, data)))
.setLoadingPercentage(nodeId, 0)
```

packages/plugin-gallery/web/src/gallery-component.jsx
```diff
@@ -46,6 +48,7 @@ GalleryComponent.propTypes = {
   error: PropTypes.object,
   t: PropTypes.func.isRequired,
   isLoading: PropTypes.bool,
+  loadingPercentage: PropTypes.number,
```
lxgreen, line 51: wow! didn't see the `PropTypes` for a looong time!

lxgreen (approved):

PR #3616, opened 2022-03-16
[docs] update customStyles docs

danarish4:


oded-soffrin:
Nice!
We should mention `RicosCustomStyles` type (https://github.com/wix/ricos/blob/master/packages/ricos-types/web/src/themeTypes.ts#L78) which consumers can also import if needed
(actually, every time we'll update this type we'll need to update the docs, so might as well add a code reference)

oded-soffrin (approved):

PR #3622, opened 2022-03-17
Poll schema

oded-soffrin:
Regarding https://wix.slack.com/archives/C8QHV6UM9/p1647524955771879
added a test and hacked the schema
@maimonav it looks like those fields (size, width) doesn't make much sense because they can never change anyhow?
also as discussed with @danarish4 , missing an e2e test that would've discover that schema validation


danarish4 (approved):

maimonav:
@oded-soffrin you're correct, this was a future feature for Polls.
We can either double check that it's not being used and remove it, or keep it if we intend to implement this feature in the future.

oded-soffrin:
> @oded-soffrin you're correct, this was a future feature for Polls. We can either double check that it's not being used and remove it, or keep it if we intend to implement this feature in the future.

How can we check?

oded-soffrin:
@danarish4 the failing tests doesn't seem relevant, are they flaky? should we remove them ?

maimonav:
Go over the code and check if there is a usage, and we can also contact maksym and check if he knows/if they make use of it on their side.

danarish4:
> @danarish4 the failing tests doesn't seem relevant, are they flaky? should we remove them ?

The tests are indeed irrelevant to this PR, I have never seen them fail before (everything is green in master) 
I wouldn't remove them, since they are necessary - we can set ignore region for the image inside the table (in that case, we only care about the focus)
We will consider what to do if these tests fail repeatedly.

oded-soffrin:
> Go over the code and check if there is a usage

Couldn't find any use case, not in product or in code (mainly looked here - https://github.com/wix/ricos/blob/master/packages/plugin-social-polls/web/src/components/PollEditor.jsx#L48)

if that's the case i'll minimize the schema only to that config (and generally will stop generating it)

PR #3630, opened 2022-03-22
[toDraft] fix documentStyle conversion

oded-soffrin:
@andrius-n is converting to and from

given content below our converter crashed due to 
`TypeError: Cannot read properties of null (reading 'decorations')`

this PR is fixing it

```js
{
      "nodes": [
        {
          "type": "PARAGRAPH",
          "id": "foo",
          "nodes": [
            {
              "type": "TEXT",
              "id": "",
              "nodes": [],
              "style": null,
              "textData": {
                "text": "Simple sentence.",
                "decorations": []
              }
            }
          ],
          "style": null,
          "paragraphData": {
            "textStyle": {
              "textAlignment": "AUTO",
              "lineHeight": null
            },
            "indentation": 0
          }
        }
      ],
      "metadata": {
        "version": 1,
        "createdTimestamp": "2022-03-21T16:40:53.970Z",
        "updatedTimestamp": "2022-03-21T16:40:53.970Z",
        "id": "69f4c2f6-e563-4872-b406-5750d87c23c0"
      },
      "documentStyle": {
        "headerOne": null,
        "headerTwo": null,
        "headerThree": null,
        "headerFour": null,
        "headerFive": null,
        "headerSix": null,
        "paragraph": null
      }
    }
```


packages/ricos-content/web/src/converters/draft/toDraft/toDraft.ts
```diff
@@ -142,8 +142,10 @@ const convert = (ricosContent: RichContent): DraftContent => {
   const parseDocStyle = documentStyle => {
     const draftDocStyle = {};
     Object.entries(documentStyle).forEach(([header, values]) => {
-      const { decorations } = values as TextNodeStyle;
```
maimonav, line 145: @talevy17 @oded-soffrin Is it expected that h1,h2.. in docStyle are null values?
If not, maybe we need to add it a normalizer to it.

oded-soffrin: Not sure if expected, we can talk with @andrius-n to understand how this content was generated with null.

talevy17: @maimonav not expected. the fields should exist only if the header type has been customized, It's probably a proto thing I didn't understand correctly.

oded-soffrin: might be a proto thing or proto definition in the blog. 
we can investigate further, but any reason not merge this defensive code ?

maimonav: If we found the source of the issue, we might not need this defensive code.
The converters should convert content, and not fixing it (of course there is always exceptional, and priorities).

oded-soffrin: So I checked and I believe blog using it via gRPC and fromJSON
and one of them is filling out the schema

and since no optional fields in proto, documentStyles get its values...

https://github.com/wix/ricos/blob/master/packages/ricos-schema/web/src/main/proto/wix/rich_content/v1/rich_content.proto#L168

oded-soffrin: also another point to consider - is the schema in the test is invalid? (proto wise)

maimonav: @oded-soffrin 
1. Regarding Blog's usage, if they use it via gPRC, we should do it as well to make sure we won't have these errors in the future.

2. Regarding the test, not sure which one?

maimonav (approved):

packages/ricos-content/web/src/converters/draft/toDraft/toDraft.ts
oded-soffrin: https://github.com/wix/ricos/pull/3630/files#diff-7329518f05df4e25f559e33a07cb99cd513c410e7b9d3b3c442625bdacb0ab13R34

PR #3669, opened 2022-03-28
Use struct as is

oded-soffrin:


packages/ricos-content/web/src/converters/draft/fromDraft/convertRicosPluginData.ts
```diff
@@ -201,9 +200,7 @@ const convertDividerData = (data: {
   data.containerData.width = { size: PluginContainerData_Width_Type.CONTENT };
 };
 
-const convertExternalData = data => {
-  data.fields = convertJsonToStruct(data).fields;
-};
+const convertExternalData = data => data;
```
oded-soffrin, line 203: @lxgreen maybe here we need FromJSON ?

PR #3671, opened 2022-03-29
[ricos-schema] fix: textStyle in Lists

maimonav:


lxgreen:
i don't understand the rational here. list item can contain non-text like image or poll. it makes no sense to add text style to list item, conceptually.

maimonav:
@lxgreen @oded-soffrin This is what we talked in the meeting with Guy, as described in the ticket [here](https://app.asana.com/0/1201536258950155/1201857423198198), to add indentation & alignment for Paragraph wrappers.
It might be wrong to call it textAlignment, but as simply alignment - of the whole listItemNode.
WDYT?

lxgreen:
it is not just the textAlignment name issue, it is the whole TextStyle being part of list item data which is irrelevant to non-textual content. if we talk about list item alignment, it could be expressed as PluginContainerData which includes alignment for non-textual content, or extending the Node [Style](https://github.com/wix/ricos/blob/ce8fe05bd55d0db3d277e46e9d0ec07916f2ddac/packages/ricos-schema/web/src/main/proto/wix/rich_content/v1/rich_content.proto#L123)

maimonav:
Agree, we should add alignment to a containerData, and treat it like a plugin - for future features.

lxgreen:
if we'd implement idea to expose PluginContainerData at the Node level rather than adding it to all the plugin data messages back then, we'd separate pure plugin data from container, decrease duplication and list item node would have the proper way to get the alignment. i still believe we should do it.

maimonav:
> if we'd implement idea to expose PluginContainerData at the Node level rather than adding it to all the plugin data messages back then,

Interesting, one thing that can be misleading is that this means that every node has a containerData, even if it does not have a container.

lxgreen:
> Interesting, one thing that can be misleading is that this means that every node has a containerData, even if it does not have a container.
indeed. This means we're gonna have a stronger schema (v2) that defines nodes separately, like the RefinedNode does. This is one of improvements we plan to do in the Q2/Q3 -- distributed plugin schemas along with their converters, models for editor/viewer and so on.

lxgreen:
> Agree, we should add alignment to a containerData, and treat it like a plugin - for future features.

iirc we already have the alignment inside container data. let's for now add container data to list item data? WDYT?

maimonav:
> iirc we already have the alignment inside container data. let's for now add container data to list item data? WDYT?

For sure.

PR #3690, opened 2022-04-03
[adsense]: fix adsense doesnt show in the editor

danarish4:


maimonav:
@danarish4 Looks great, how about adding a test for it?

maimonav (approved):

PR #3703, opened 2022-04-04
[schema] replace duplicated struct w/ google struct (schema files only)

oded-soffrin:


oded-soffrin:
First phase solution:
having the wix-schema using google struct to unblock platformized efforts (auto-velo)
converting to current solution on OS to keep same behavior as today (bundle size, converters, etc)

next steps:
use google struct in OS as well and add mutate function to fix existing externalData content to match new format

packages/ricos-schema/web/src/generateSchema.ts
```diff
@@ -35,6 +35,11 @@ const run = async () => {
         .replace('import "wix/api/entity.proto";\n', '')
         .replace('import "wix/api/validations.proto";\n', '')
         .replace('import "wix/api/annotations.proto";\n', '')
+        .replace(
+          'import "google/protobuf/struct.proto"',
```
lxgreen, line 39: does this replacement mean we use google struct on wix-ricos only?

lxgreen: if yes, then how can we update the draft converters?

lxgreen (approved):
so, AFAIU the OS has no changes here?

PR #3717, opened 2022-04-07
[e2e] add toolbars v2 tests

noamtwix:


e2e/tests/text.e2e.ts
```diff
@@ -1,18 +1,23 @@
...
 
-const changeTextColor = (title: string) => {
-  cy.loadRicosEditorAndViewer('plain')
-    .setTextStyle(INLINE_TOOLBAR_BUTTONS.COLOR, [20, 15])
+const useNewToolbar = useExperiments({
+  newFormattingToolbar: { namespace: 'ricos', value: 'true', enabled: true },
+});
+
+const changeTextColor = (title: string, toolbarsV2?: boolean) => {
+  const testTitle = `${title} - ${toolbarsV2 ? 'toolbars v2' : 'old toolbar'}`;
```
oded-soffrin, line 10: the same logic behind using `v2` and not `new` , 
hold for `v1` and not `old`

(when we'll move to v3, what will it mean old? v1? v2?)

PR #3733, opened 2022-04-12
split ListNode type per ordered/bulleted

GuyHaviv37:
Each list node type can have its own "data" property (for indentation) I need to consume it with a ListNode type.


maimonav (approved):
@GuyHaviv37 Thank you, missed this one.

PR #3743, opened 2022-04-14
[tiptap] add html extension

danarish4:


packages/plugin-html/web/src/tiptap/tiptap.ts
```diff
@@ -0,0 +1,24 @@
+import htmlDataDefaults from 'ricos-schema/dist/statics/html.defaults.json';
+import type { CreateRicosExtensions } from 'ricos-tiptap-types';
+import { Image as Component } from './component';
```
maimonav, line 3:
```suggestion
import { Html as Component } from './component';
```

packages/plugin-html/web/src/tiptap/component.tsx
```diff
@@ -0,0 +1,36 @@
+import React from 'react';
+import { Component as HtmlComponent } from '../HtmlComponent';
+import { HTML_TYPE } from '../types';
+import type { PluginProps } from 'ricos-tiptap-types';
+
+export const Image: React.FC<PluginProps> = ({
```
maimonav, line 6:
```suggestion
export const Html: React.FC<PluginProps> = ({
```

packages/plugin-html/web/src/tiptap/tiptap.ts
```diff
@@ -0,0 +1,24 @@
+import htmlDataDefaults from 'ricos-schema/dist/statics/html.defaults.json';
+import type { CreateRicosExtensions } from 'ricos-tiptap-types';
+import { Html as Component } from './component';
+import { TIPTAP_HTML_TYPE } from 'ricos-content';
+
+export const createRicosExtensions: CreateRicosExtensions = defaultOptions => [
```
maimonav, line 6: General question, how do we handle `pluginsConfig` from consumers?
I see it's being added as `defaultOptions`, but the default data ignores it.
For example, user passed defaultData: { width: 100, height: 300}.
I'm aware that we can't test it yet since we have no toolbar so we can't insert plugins in Tiptap.

danarish4: In the case of pluginEditorConfig - defaultOptions is exactly this (maybe we should to rename it) 
You can see we call createRicosExtensions from editor.ts file with the plugin config.

In the case of general editor config (such as domainSandbox) - @justame is working on infrastructure to pass editor config to extensions. After he finishes, we will pass it to the relevant extensions.

maimonav (approved):
Looks great, just update changelog :)

PR #3755, opened 2022-04-19
[toolbars-v3] feat: Mega Toolbar

maimonav:


packages/toolbars-v3/web/src/toolbarItemConfig/draftMegaToolbarItemConfig.ts
```diff
@@ -0,0 +1,595 @@
+// import { ContentResolver } from './ContentResolver';
```
noamtwix, line 1: Do we need this file?

maimonav: Now we don't :)
Previously, I made this Mega Toolbar POC for draft, 2 months gone by
I've changed it so it will be supported via Tiptap

PR #3759, opened 2022-04-19
[tiptap] add insertText command

danarish4:


oded-soffrin:
merged to here https://github.com/wix/ricos/pull/3763, thanks!

PR #3773, opened 2022-04-24
keyboard shortcuts: domain model

lxgreen:


packages/ricos-editor/web/src/RicosEditorTiptap/RicosEditorTiptap.tsx
```diff
@@ -77,26 +77,17 @@ class RicosEditorTiptap extends React.Component<RicosEditorProps & IRicosContext
...
-        placeholder,
-        textAlignment,
-        iframeSandboxDomain,
-        isTextWrap: textWrap,
-        maxTextLength,
-        anchorTarget,
-        relValue,
-        rel,
-      });
+    return extensions.configure({
```
justame, line 82: I'm not sure if the configure on the aggregator would be the best place because except for our special case there is no need to configure all the extensions with the same config

packages/ricos-editor/web/src/keyboard-shortcuts/common-shortcuts.ts
```diff
@@ -0,0 +1,11 @@
+import type { EditorCommands, KeyboardShortcut } from 'ricos-types';
+
+export const boldShortcut: KeyboardShortcut = {
+  keys: 'Cmd+B',
+  command(editorCommands: EditorCommands) {
+    editorCommands.toggleInlineStyle('bold');
+  },
+  contexts: [],
+  tooltipHint: '',
```
justame, line 9: maybe 'name' would be better? tooltipHint is for the "toolbar" case

packages/ricos-editor/web/src/RicosEditorTiptap/RicosEditorTiptap.tsx
lxgreen: actually, there are couple of cases more:

- the "dynamicConfiguration" gonna use it
- the RicosExtension `addKeyboardShortcuts` field removal as part of Keyboard Shortcuts

packages/ricos-editor/web/src/keyboard-shortcuts/common-shortcuts.ts
lxgreen: hmmm

this is exactly the keys hint for tooltip like `(⌘B)`. AFAIK, it is gonna be used in tooltips only.

PR #3778, opened 2022-04-25
[ricos-modals] add modal service

danarish4:


packages/ricos-types/web/src/modalTypes.ts
```diff
@@ -0,0 +1,12 @@
...
+
+export type ModalGroup = 'dialog' | 'fullscreen';
+
+export type ModalOptions = { referenceElement?: HTMLElement; groups?: ModalGroup[] };
+
+export type IModal = {
+  Component: ComponentType;
+  modalId: string; //change to id
+  options?: ModalOptions;
+  state?: { open: boolean; nodeId?: string };
```
lxgreen, line 11:
```suggestion
  Component: ComponentType;
  id: string;
  groups: Group[]; // we can omit the 'Modal' prefix everywhere since we're in context of Modal type
  positioning: ModalPositioning; // this should be well-defined type that admits everything related to positioning
```

lxgreen: `ModalOptions` type is meaningless, as it contains unrelated fields. The `groups` and `positioning` can just be listed under root.

The `state` field is irrelevant to this type, 
as it refers runtime component state, and this type describes modal configuration. Modal state deserves a separate well-defined type.

modals

lxgreen: The `IModal` called `ModalConfig` here, and the Modal entity represents all the modals in editor in runtime. So, the runtime entity has state.

lxgreen (changes requested):

PR #3791, opened 2022-04-28
[ricos-editor] polish: FullRicosEditorTiptap impl RicosEditorRef

maimonav:
FullRicosEditorTiptap implements RicosEditorRef to support editor ref API.


packages/tiptap-editor/web/src/components/RichContentAdapter/RichContentAdapter.tsx
```diff
@@ -40,6 +40,23 @@ export class RichContentAdapter implements TiptapAPI {
...
+
+  getContentPromise: TiptapAPI['getContentPromise'] = async ({ flush, publishId } = {}) => {
+    throw new Error('Method not implemented.');
+  };
+
+  getContentTraits: TiptapAPI['getContentTraits'] = () => {
+    throw new Error('Method not implemented.');
+  };
+
+  //@ts-ignore
```
justame, line 55: why ts-ignore?

maimonav: Because this is a temp solution, we won't support getToolbarProps in this current api, external toolbar planned to be deprecated.
If we do keep external toolbar, we'll have to change it's type because it's coupled to draft.
So for now, it's good enough.

PR #3812, opened 2022-05-02
[text-color,highlight-color] feat: support consumer's colors in tiptap (color1,...)

maimonav:


packages/plugin-text-color/web/src/tiptap.ts
```diff
@@ -57,10 +59,19 @@ export const getTiptapExtensions: CreateRicosExtensions = defaultOptions => [
...
+          const parseColor = (color: Color) =>
+            defaultOptions?.customStyleFn?.(OrderedSet([color]))?.color;
+
+          const getColor = (color: Color) => (shouldParseColor(color) ? parseColor(color) : color);
           return [
             'span',
             {
-              style: `color: ${HTMLAttributes.foreground}; background-color: ${HTMLAttributes.background}`,
+              style: `color: ${getColor(HTMLAttributes.foreground)}; background-color: ${getColor(
+                HTMLAttributes.background
```
talevy17, line 73: this might not work properly for highlight color, we get different `customStyleFn` & `styleSelectionPredicate` for `text-highlight-plugin` since we treat it as a different plugin

maimonav: After checking with consumers, it looks like it behaves the same for text-color & highlight-color.
I'll merge it for now, if we see it's an issue - we might need to separate it to two extensions.

PR #3817, opened 2022-05-02
[toolbars-v3] feat: tiptap gallery layout selector

talevy17:


packages/toolbars-v3/web/src/resolvers/tiptapResolvers.ts
```diff
@@ -330,3 +329,13 @@ export const isPluginSelectedResolver = TiptapContentResolver.create(
...
   content => pluginSelectedResolvers.some(isPluginSelected => isPluginSelected.resolve(content))
 );
+
+export const getGalleryLayoutResolver = TiptapContentResolver.create(
+  RESOLVERS_IDS.GET_GALLERY_LAYOUT,
+  content => {
+    if (content?.length === 1) {
+      return content[0].attrs.options?.layout?.type;
+    }
+    return false;
```
justame, line 339: if you want a falsy value , try to return an empty string instead false.
it's unpredicted that one time it can return a string and one time it can return a boolean

PR #3840, opened 2022-05-10
[tiptap-full-editor] add upload provider

danarish4:


packages/ricos-editor/web/src/RicosContext/UploadProvider.tsx
```diff
@@ -0,0 +1,20 @@
...
+        <UploadContextWrapper
+          locale={locale}
+          editorCommands={getEditorCommands()}
+          isMobile={isMobile}
+          helpers={helpers}
+        >
+          {children}
+        </UploadContextWrapper>
+      )}
+    </RicosContextConsumer>
```
lxgreen, line 18:
```suggestion
      const { getEditorCommands, isMobile, locale } = useContext(RicosContext);
      return (
        <UploadContextWrapper
          locale={locale}
          editorCommands={getEditorCommands()}
          isMobile={isMobile}
          helpers={helpers}
        >
          {children}
        </UploadContextWrapper>
      )
```

lxgreen: While in functional component, it is way more elegant to use hook. Strongly suggest to adopt `useContext` instead of consumer/provider wrappers, and functional components instead of classes, as well.

packages/ricos-editor/web/src/RicosEditorTiptap/FullRicosEditorTiptap.tsx
```diff
@@ -200,21 +201,23 @@ export class FullRicosEditorTiptap
...
-                  />
-                  <FloatingAddPluginMenu
-                    pluginsButtons={plugins
-                      ?.filter(plugin => plugin.addButtons)
-                      .map(plugin => plugin.addButtons)}
-                    editor={this.editor}
-                  />
-                </ToolbarContext.Provider>
-              </ModalProvider>
+              <UploadProvider helpers={_rcProps?.helpers}>
```
lxgreen, line 204: here, the only reason to use provider wrapper is initialization with `helpers` value. 

Another context related note: if you examine `ModalProvider`, it does not require any initialization, but renders the `ModalPopper`, so it got its right to wrap. The `withModalContext` HOC exists due to the fact that all the `InsertModal`s are class components.

packages/ricos-editor/web/src/RicosContext/UploadProvider.tsx
lxgreen: Note: the `UploadServiceContext` renders additional components, so it is OK to use it as wrapper.

lxgreen (changes requested):

lxgreen (commented):
cool

PR #3866, opened 2022-05-16
[ricos-editor] feat: ricos portal

maimonav:
Currently, **all toolbars/modals** related are using `document.body` as a portal, which makes theme CSS variables **unavailable to them (when parentClass is defined).**
Ricos Portal fixes that, it uses the parentClass if defined.

~@lxgreen need to discuss the following:~
 - ~[Fullscreen](https://github.com/wix/ricos/blob/master/packages/fullScreen/web/src/FullScreen.jsx#L11)~
 - ~[EditorModalProvider](https://github.com/wix/ricos/blob/master/packages/ricos-editor/web/src/modals/EditorModalProvider.tsx#L143) @oded-soffrin~
 - ~[StaticToolbar](https://github.com/wix/ricos/blob/master/packages/ricos-editor/web/src/tiptap/RicosToolbars.tsx#L189)~
 - ~[TextFormattingToolbar](https://github.com/wix/ricos/blob/master/packages/ricos-editor/web/src/toolbars/TextFormattingToolbar.tsx#L200) @noamtwix~
 - ~[ModalPopper](https://github.com/wix/ricos/blob/master/packages/ricos-modals/web/src/utils/ModalPopper.tsx#L33) @danarish4~


packages/ricos-editor/web/src/tiptap/RicosPortal.tsx
```diff
@@ -0,0 +1,10 @@
+import React, { forwardRef } from 'react';
+import ReactDOM from 'react-dom';
+
+interface Props {
+  className?: string;
+}
+
+export default forwardRef<HTMLDivElement, Props>((props, ref) => {
+  return ReactDOM.createPortal(<div ref={ref} className={props?.className} />, document.body);
```
maimonav, line 9: Maybe should generate some id with className? to avoid having same classNames for parent and portal?

packages/ricos-types/web/src/contexts.ts
```diff
@@ -9,4 +10,5 @@ export interface GeneralContext {
   t: TranslationFunction;
   languageDir: TextDirection;
   theme: RicosTheme;
+  portal: RicosPortal;
```
oded-soffrin, line 13: shouldn't it be:

```suggestion
  portal?: RicosPortal;
```

(to sort out all the `null as unknown as RicosPortal` ?)

maimonav: It will fix the unknown, but then every context usage I'll need to check that portal is exist.
I've added doc.body as default value so it will always be defined.

danarish4:
> Currently, **all toolbars/modals** related are using `document.body` as a portal, which makes theme CSS variables **unavailable to them (when parentClass is defined).** Ricos Portal fixes that, it uses the parentClass if defined.
> 
> @lxgreen need to discuss the following: ~[Fullscreen](https://github.com/wix/ricos/blob/master/packages/fullScreen/web/src/FullScreen.jsx#L11)~ ~[EditorModalProvider](https://github.com/wix/ricos/blob/master/packages/ricos-editor/web/src/modals/EditorModalProvider.tsx#L143) @oded-soffrin~
> 
> * [StaticToolbar](https://github.com/wix/ricos/blob/master/packages/ricos-editor/web/src/tiptap/RicosToolbars.tsx#L189)
> * [TextFormattingToolbar](https://github.com/wix/ricos/blob/master/packages/ricos-editor/web/src/toolbars/TextFormattingToolbar.tsx#L200)
> * [ModalPopper](https://github.com/wix/ricos/blob/master/packages/ricos-modals/web/src/utils/ModalPopper.tsx#L33) @danarish4

Greate :) 
We faced a similar issue with business manager styles for modals.
@tamrat-bay will consult with you about this when he will reach to handle it

danarish4 (approved):
Great refactor
Thanks

packages/ricos-editor/web/src/tiptap/RicosPortal.tsx
lxgreen: is this a problem if they get same classes?

lxgreen (approved):
lux gud

PR #3868, opened 2022-05-17
[mentions] refactor mentions ui in tiptap

tamrat-bay:


packages/plugin-mentions/web/statics/mention-list.scss
```diff
@@ -1,25 +1,19 @@
...
+  padding-left: 12px;
+  white-space: nowrap;
+  overflow: hidden;
+  text-overflow: ellipsis;
 
-  &.selected {
-    border-color: #000;
+  svg {
+    min-width: 28px;
+    min-height: 28px;
```
danarish4, line 17: why this is not sit under ui-component?

tamrat-bay: I didn't want the **<ListItem />** component to have a **fixed** width and **position** for its content 
I think this will make it more flexible and reusable

However, I can make 'mention-item' class slimmer

image

danarish4: you totally right

danarish4 (approved):

PR #3875, opened 2022-05-18
[ricos-editor] fix: force update to make portal ref available

maimonav:


maimonav:
I've tried to find a more elegant solution rather than use forceUpdate, if you think of anything - I would be glad to hear :)

PR #3910, opened 2022-05-26
[toolbars-v2] fix: modal buttons on atomic + text selected

noamtwix:


packages/toolbars-new/web/src/Toolbar/ButtonComponents/ModalButton.tsx
```diff
@@ -59,8 +59,11 @@ interface State {
 class ModalButton extends Component<ModalButtonProps, State> {
   modalRef?: HTMLDivElement | null;
 
+  buttonRef?: React.RefObject<any> | undefined;
```
noamtwix, line 62:
```suggestion
  buttonRef = React.createRef();
```

PR #3912, opened 2022-05-26
[mentions] fix white space issue & refactor ui

tamrat-bay:


packages/plugin-mentions/web/src/tiptap/suggestion.ts
```diff
@@ -70,6 +82,11 @@ export default (
             return true;
           }
 
+          if (props.event.key === 'Enter') {
+            const { items, query } = component.props;
+            shouldEndMentioningProcess(query, items) && this.onExit();
+          }
```
danarish4, line 88:
```suggestion
          if (shouldEndMentioningProcess(query, items)) {
              this.onExit();
          }
```

check & end mention process on every char typed and not only on Enter.
This need to be the first condition because I think arrow-up\arrow-down will also not work

danarish4 (changes requested):
please follow comment

PR #3933, opened 2022-05-29
Image legacy type convert

danarish4:


oded-soffrin (approved):
Looks good!
let's add a test though , I think it belongs here - https://github.com/wix/ricos/blob/master/packages/ricos-content/web/src/converters/draft/toDraft/toDraft.spec.ts#L25

PR #3961, opened 2022-06-01
[toolbars-v2] modal overflow logic using only getBoundingClientRect (…

noamtwix:
…under experiment)


packages/editor-common/web/src/Utils/overflowUtils.ts
```diff
@@ -21,11 +21,23 @@ const getElementCoordsInWindow = (elem: HTMLElement): ElementCoords => {
...
+};
+
 export const elementOverflowWithEditor = (
   element: HTMLElement,
-  rootEditorElement?: HTMLElement
+  rootEditorElement?: HTMLElement,
+  modalOverflowByBoundingClientRect?: boolean
 ): Record<string, number> => {
-  const elementCoordsInWindow = getElementCoordsInWindow(element);
+  const elementCoordsInWindow = modalOverflowByBoundingClientRect
```
oded-soffrin, line 38: might need to change in other `getElementCoordsInWindow` usage

packages/editor-common/web/src/Utils/overflowUtils.ts
```diff
@@ -21,11 +21,23 @@ const getElementCoordsInWindow = (elem: HTMLElement): ElementCoords => {
   return { top: Math.round(top), left: Math.round(left) };
 };
 
+const getElementCoordsInWindowByBoundingClientRect = (elem: HTMLElement): ElementCoords => {
+  const box = elem.getBoundingClientRect();
+
+  const top = box.top;
+  const left = box.left;
+
```
oded-soffrin, line 29:
```suggestion
  const {top, left} = elem.getBoundingClientRect();
```

oded-soffrin:
`modalOverflowByBoundingClientRect` appears a lot for this change, not ideal
(we should either encapsulate into a groups of params (to include future params), or pass the whole epxeriments object)
better yet - use context

(not super important because all of this code will soon be dead)

PR #3990, opened 2022-06-07
[toolbars-v3] data hooks

noamtwix:


packages/toolbars-v3/web/src/components/SizeCalculator/SizeCalculator.tsx
```diff
@@ -1,13 +1,13 @@
 import React, { useState, useRef, useEffect } from 'react';
 
 type SizeResult = {
-  visible: React.Component[];
-  overflowed: React.Component[];
+  visible: JSX.Element[];
```
noamtwix, line 4: use ComponentType

oded-soffrin:
@noamtwix @lxgreen  please ping me when this is merged

packages/toolbars-v3/web/src/models/toolbarButtonsTypes.ts
```diff
@@ -0,0 +1,14 @@
+import type { ToolbarButton } from './index';
+
+export interface IToolbarButton {
+  renderComponent: () => JSX.Element;
+  renderComponentWithDataHook: () => JSX.Element;
+}
+
+export interface IToolbarButtons {
```
lxgreen, line 8: i can't see what's the responsibility of this aggregate.

`getButtonByIndex`, `isEmpty` are just utility methods;
`addButton` accepts **entity** that this aggregate supposed to **abstract** from user.

i suggest to stop using these models in visible/overflow calculation as it only complicates things. make them responsible to get you elements with/without dataHook, and use them only there. this will distill interfaces better, and will help you to find a good name.

packages/toolbars-v3/web/src/models/toolbarButtonsTypes.ts
```diff
@@ -0,0 +1,14 @@
+import type { ToolbarButton } from './index';
+
+export interface IToolbarButton {
+  renderComponent: () => JSX.Element;
```
lxgreen, line 4: worth to check if all occurrences of `JSX.Element` can be replaced with `ReactElement` type. It is more convenient, as JSX talks about certain HTML-in-JS templating mechanism.

lxgreen (changes requested):

PR #3996, opened 2022-06-07
[tiptap-audio] audio add button

tamrat-bay:


packages/plugin-audio/web/src/editor.ts
```diff
@@ -18,5 +18,6 @@ export const pluginAudio: EditorPluginCreator<AudioPluginEditorConfig> = config
     createPluginData: createAudioData,
     tiptapExtensions,
     toolbarButtons: getToolbarButtons(config),
+    addButtons: getAddButtons(config),
```
danarish4, line 21:
```suggestion
    toolbarButtons: getToolbarButtons(pluginConfig),
    addButtons: getAddButtons(pluginConfig),
```

packages/plugin-audio/web/src/modals/AudioInsertModal.tsx
```diff
@@ -8,13 +8,13 @@ import {
   MEDIA_POPOVERS_BUTTONS_NAMES_BI,
   MEDIA_POPOVERS_TABS_NAMES_BI,
 } from 'wix-rich-content-common';
-import { AUDIO_TYPE } from '../../types';
-import styles from '../../../statics/styles/audio-modal.scss';
+import { AUDIO_TYPE } from '../types';
+import styles from '../../statics/styles/audio-modal.scss';
 
-const AudioModal = props => {
+const AudioInsertModal = props => {
```
danarish4, line 14: Please add types to props

packages/plugin-audio/web/src/modals/AudioUploadModal.tsx
```diff
@@ -54,12 +53,8 @@ const AudioUploadModal = props => {
     }
   };
 
-  const closeModal = () => {
-    helpers.closeModal();
-  };
-
   const setComponentData = data => {
-    pubsub.set('componentData', data, blockKey);
+    pubsub?.set('componentData', data, blockKey);
```
danarish4, line 57: what happens in the case of tiptap? is this function doesn't do anything?
we need to pass setComponentData as props and support both cases

packages/plugin-audio/web/src/modals/AudioUploadModal.tsx
```diff
@@ -96,7 +91,7 @@ const AudioUploadModal = props => {
       html = null,
       coverImage = null,
       ...rest
-    } = { ...componentData, ...pubsub.get('componentData') };
+    } = { ...componentData, ...pubsub?.get('componentData') };
```
danarish4, line 94: same here- in case of tiptap pubsub?.get('componentData') will return undefined and the app will crash because you tey to spread undefined

danarish4 (changes requested):
כםךךם׳ בםצצקמאד

packages/plugin-audio/web/src/modals/AudioUploadModal.tsx
tamrat-bay: In case of tiptap it won't do anything and it shouldn't . 
 spreading undefined wouldn't crash the app
{...undefined} // {}

PR #4001, opened 2022-06-08
TS bump

oded-soffrin:


oded-soffrin:
@noamtwix 

Trying to bump TS, I see it fails because 'HEADINGS' (https://github.com/wix/ricos/blob/8413a6175ab668148864a20141774b80a7908b47/packages/toolbars-new/web/src/Toolbar/buttonsListCreatorConsts.tsx#L493) isn't part of TextBlockType (https://github.com/wix/ricos/blob/d5092353e1fe949854fd23bba9eab6623aa4bc8c/packages/ricos-types/web/src/editorCommandsType.ts#L60)


Should it is HEADINGS_TYPE - https://github.com/wix/ricos/blob/cc4bfeb20c4da88ffe74d77e394aa0ae2b212a78/packages/ricos-content/web/src/consts.ts#L38

notice there a change from captial letters to lower case

PR #4029, opened 2022-06-12
add text background color support to theme

talevy17:


packages/common/web/statics/styles/fonts.scss
```diff
@@ -97,6 +97,15 @@ $text-decorations: (
   min-height: var(--ricos-custom-#{$name}-min-height, map-get($min-heights, $name));
   @if $name == 'p' {
     color: $ricos-p-color;
+    @if $isViewer {
+      @at-root #{'& > * > span'} {
+        background-color: $ricos-p-background-color;
+      }
+    } @else {
+      @at-root #{'& > span'} {
```
maimonav, line 105: Shouldn't it be only the first span in the dom?
For example, 

```suggestion
      @at-root #{'& + span'} {
```

packages/plugin-headings/web/src/tiptap.ts
```diff
@@ -71,7 +71,7 @@ export const tiptapExtensions = [
             mergeAttributes(this.options.HTMLAttributes, HTMLAttributes, {
               class: styles[`header${HEADER_NAMES[level - 1]}`],
             }),
-            0,
+            ['span', { style: 'display: inline-block;' }, 0],
```
maimonav, line 74: Why is this style: 'display: inline-block;' needed?

packages/plugin-text-color/web/src/tiptap.ts
```diff
@@ -60,10 +61,14 @@ export const tiptapExtensions = [
...
-              tag: 'span',
+              tag: 'span > span',
               getAttrs: element => {
                 const { color, backgroundColor } = (element as HTMLElement).style || {};
-                return color || backgroundColor ? {} : false;
+                const attributes: { foreground?: string; background?: string } = {};
+                color && (attributes.foreground = rgbToHex(color) || color);
+                backgroundColor &&
+                  (attributes.background = rgbToHex(backgroundColor) || backgroundColor);
+                return color || backgroundColor ? attributes : false;
```
maimonav, line 71: Let's extract this to another PR, it's text-color copy/paste related

packages/plugin-headings/web/src/tiptap.ts
talevy17: in order to keep the parent's height and the highlight gets the same height as the inline style

examples/storybook/src/shared/resources/fontsExample.ts
```diff
@@ -1,5 +1,8 @@
 export const FONTS = [
-  { h2: { fontFamily: 'Arial' }, p: { fontFamily: 'Comic Sans MS' } },
+  {
+    h2: { fontFamily: 'Arial', color: 'white', backgroundColor: 'black' },
+    p: { fontFamily: 'Comic Sans MS', color: 'white', backgroundColor: 'black' },
+  },
```
maimonav, line 5: don't forget to remove when it's done :)

PR #4035, opened 2022-06-13
[ricos-context] move plugins context to ricos-context

danarish4:


packages/ricos-context/web/package.json
```diff
@@ -36,6 +36,7 @@
   "dependencies": {
     "ricos-tiptap-types": "8.71.23",
     "ricos-types": "8.71.23",
+    "ricos-plugins": "8.71.23",
```
maimonav, line 39: Is this needed because of the `EditorPlugins` type? if so, I believe this type should be extracted, probably to `ricos-types`.
`ricos-context` should not have `ricos-plugins` dependency, I believe this will fix the bundle issue as well.

PR #4109, opened 2022-07-03
[tiptap] floating toolbar click outside fix

danarish4:


oded-soffrin:
@danarish4 this PR is introducing breaking e2e test (!)

PR #4110, opened 2022-07-03
[divider/tiptap] add toolbar size buttons

tamrat-bay:


danarish4:
Please use the current divider icons instead of duplicate

PR #4113, opened 2022-07-03
[divider/tiptap] add toolbar linestyle buttons

tamrat-bay:


danarish4:
Looks good.
why not reuse current divider icons?

PR #4115, opened 2022-07-04
[toolbars-v3] feat: add plugin menu button in tiptap mobile toolbar

maimonav:


maimonav:
@danarish4 Please take a look to see that regarding modal service usage it's ok, plus, there is a css issue, so please take a look when you can.
@noamtwix please take a look at the draftStaticToolbarItemConfigs, I've removed the old addPlugin config, just making sure it's ok and there is no risk in removing this.

danarish4 (approved):
🔥

PR #4119, opened 2022-07-05
fix(html): use srcDoc instead of ref write api

oded-soffrin:
This PR fixed tiptap behavior where html isn't working when not using iframesandboxdomain
(for some reason, iframe ref is being triggered but iframe window document isn't available, so we can't write into it)

switching to srcDoc fixes it. 
NOTE - srcDoc isn't supported in IE, but wix not longer supported IE, and anyhow iframesandboxdomain should be use anywhere.


danarish4 (approved):
👑

PR #4153, opened 2022-07-10
Tiptap/html width height

talevy17:


packages/plugin-html/web/src/toolbar/DimensionSliderButton.tsx
```diff
@@ -0,0 +1,76 @@
...
+    toolbarItem.commands?.click({ data: { ...componentData, containerData } });
+  };
+
+  const onClick = () =>
+    modalService.openModal(id, {
+      componentProps: {
+        ...props,
+        getValue,
+        onChange,
+        close: () => modalService.closeModal(id),
```
danarish4, line 58: I dont think we need this
The modal shouldnt need to "close itself" you need to do something like this: 

```suggestion
```

const onClick = () => {
   if (modalService?.isModalOpen(id)) {
          modalService.closeModal(id);
        } else {
          modalService?.openModal(id, {
            componentProps: {
              ...props,
              getValue,
             onChange,
            },
            positioning: { placement: 'bottom', referenceElement },
            layout: isMobile ? 'drawer' : 'popover',
          });
        }
      }
}

danarish4 (changes requested):
follow comments

PR #4178, opened 2022-07-13
replace uploadContext with uploadService in commands

talevy17:


maimonav (approved):
Looks good :). Please update changelog

PR #4182, opened 2022-07-13
[tiptap]: fix drag and drop in collapsible lists

talevy17:


danarish4 (approved):
The end solution is SO beautiful, I'm so happy! 👑 
It still needs some UI indications, let's merge it and continue working on the UI in a separate PR

PR #4197, opened 2022-07-17
Add text buttons to formatting plugins

noamtwix:


packages/plugin-code-block/web/src/getTextButtons.ts
```diff
@@ -0,0 +1,35 @@
...
+    {
+      id: 'codeBlock',
+      type: 'toggle',
+      presentation: {
+        dataHook: 'TextCodeBlockButton',
+        tooltip: 'TextCodeBlockButton_Tooltip',
+        tooltipShortcut: {
+          MacOS: ' (⌘⇧C)',
+          Windows: ' (Ctrl+⇧+C)',
+        },
```
lxgreen, line 19: we'll have to replace this API => shortcuts API

packages/plugin-code-block/web/src/getTextButtons.ts
```diff
@@ -0,0 +1,35 @@
...
+        visible: alwaysVisibleResolver,
+        active: isTextContainsCodeblockResolver,
+      },
+      commands: {
+        toggleCodeblock:
+          ({ editorCommands }) =>
+          () => {
+            editorCommands.chain().focus().toggleCodeBlock().run();
+          },
+      },
```
lxgreen, line 32: it is better to create dedicated tiptap extensions for commands. otherwise, tiptap code is spread all over the plugin. later this Q we plan to replace all the editor commands with model-based comands. so please make sure the commands for buttons defined as 

```ts
commands: {
  toggleCodeblock:
    ({ editorCommands }) => () => editorCommands.toggleCodeBlock(),
}  
```

i.e. all the command logic is moved to extension that implements `toggleCodeBlock`.

lxgreen: this way, we'll be able to reuse this command in other places, text button definition gonna be simpler, and the future commands replacement should be easier

lxgreen: @maimonav WDYT?

lxgreen (changes requested):
notes:
1. tiptap commands should move to extensions
2. toolbar shortcuts API --> shortcuts API
3. ricos-plugins should feature TextButton entity + TextButtons aggregate (let's discuss)

PR #4241, opened 2022-07-24
plugin toolbar buttons order and seperators

talevy17:


danarish4 (changes requested):
Looks good.
In the toolbars-v3 package, there is already a `ToolbarButtonSeparator` component.
Use your component in toolbars-v3 and remove it from there, so it will only exist in toolbars-ui

PR #4260, opened 2022-07-25
[ricos-plugins] get formatting toolbar item config from plugins

noamtwix:


packages/ricos-types/web/src/editorPluginTypes.ts
```diff
@@ -185,8 +185,12 @@ export interface FormattingToolbarButton {
  * @interface FormattingToolbarButtons
  */
 export interface FormattingToolbarButtons {
-  asArray: () => FormattingToolbarButton[];
-  toToolbarItemsConfigs(editorCommands: EditorCommands): IToolbarItemConfigTiptap[];
+  toToolbarItemsConfig: (
+    ToolbarType: ToolbarType,
+    isMobile: boolean,
+    editorCommands: EditorCommands
+  ) => IToolbarItemConfigTiptap[] | null;
```
maimonav, line 192: Let's remove the `null` if possible

maimonav (changes requested):

maimonav (approved):

PR #4267, opened 2022-07-26
[tiptap-plugins]internal: inject services from plugin entity

danarish4:


packages/ricos-types/web/src/editorPluginTypes.ts
```diff
@@ -275,3 +264,12 @@ export interface IPluginToolbarButton {
 
   unregister: () => void;
 }
+
+export type PluginServices = {
```
lxgreen, line 268: we already [have](https://github.com/wix/ricos/blob/master/packages/ricos-types/web/src/index.ts#L40) it, please reuse

lxgreen (changes requested):
niiiice! just see the type

packages/ricos-types/web/src/editorPluginTypes.ts
danarish4: I thought of reusing it as `Partial<PluginServices>`
But PluginServices type is missing `modalService` & `t`

Do you think the type should be 
`Partial<PluginServices> & { t: TranslationFunction, modalService: ModalService } `

lxgreen: nope, just let's add **all** available services to RicosServices, and use it everywhere

PR #4284, opened 2022-07-28
[toolbars-ui] fix dropdown button selection state

tamrat-bay:


danarish4:
I find it strange that alignment modals (for example) are registered as `DropdownModal`.
For each standard button (e.g. alignmentPanel, sizePanel, which will also have options as props), I think we should create a unique dropdown. That way, alignmentButton will register alignmentPanel as its modal.
By doing so, alignmentButton will be a simple button that opens alignmentPanel when clicked.
In addition, DropdownButton component will no longer be required.

packages/plugin-audio/web/src/getToolbarButtons.tsx
```diff
@@ -34,15 +34,27 @@ export const getToolbarButtons = (config, services): ToolbarButton[] => {
   return [
     {
       id: PLUGIN_TOOLBAR_BUTTON_ID.SIZE,
+      modal: {
+        Component: DropdownModal,
```
danarish4, line 38:
```suggestion
        Component: decorateComponentWithProps(SizePanel, { options:{['SMALL', 'CONTENT'] as PluginContainerData_Width_Type[]} }),
```

packages/plugin-audio/web/src/getToolbarButtons.tsx
```diff
@@ -34,15 +34,27 @@ export const getToolbarButtons = (config, services): ToolbarButton[] => {
...
+        id: AUDIO_BUTTONS.size,
+      },
       renderer: toolbarItem => (
         <NodeSizeButton
+          id={AUDIO_BUTTONS.size}
           toolbarItem={toolbarItem}
           options={['SMALL', 'CONTENT'] as PluginContainerData_Width_Type[]}
```
danarish4, line 45:
```suggestion
```

packages/plugin-audio/web/src/getToolbarButtons.tsx
```diff
@@ -34,15 +34,27 @@ export const getToolbarButtons = (config, services): ToolbarButton[] => {
...
         />
       ),
     },
     {
       id: PLUGIN_TOOLBAR_BUTTON_ID.ALIGNMENT,
+      modal: {
+        Component: DropdownModal,
```
danarish4, line 52:
```suggestion
        Component: AlignmentPanel,
```

packages/plugin-button/web/src/getToolbarButtons.tsx
```diff
@@ -1,15 +1,24 @@
...
+import { DropdownModal, NodeAlignmentButton } from 'wix-rich-content-toolbars-ui';
+import { buttonsModals, BUTTON_BUTTONS } from './constants';
 import ButtonSettingsModal from './modals/SettingsModal';
 
 export const getToolbarButtons = (config, services, type): ToolbarButton[] => {
   return [
     {
       id: PLUGIN_TOOLBAR_BUTTON_ID.ALIGNMENT,
+      modal: {
+        Component: DropdownModal,
```
danarish4, line 16:
```suggestion
        Component: AlignmentPanel,
```

tamrat-bay:
> I find it strange that alignment modals (for example) are registered as `DropdownModal`. For each standard button (e.g. alignmentPanel, sizePanel, which will also have options as props), I think we should create a unique dropdown. That way, alignmentButton will register alignmentPanel as its modal. By doing so, alignmentButton will be a simple button that opens alignmentPanel when clicked. In addition, DropdownButton component will no longer be required.

@danarish4 
As I understand it, you would like different Dropdown components
For each button ( e.g. AlignmentPanel / SizePanel )  instead of using `DropdownModal.tsx ` is it right?
If so, we will have different Dropdown components, but the code will be duplicated

e2e/tests/plugins.e2e.ts
```diff
@@ -95,7 +95,7 @@ import { getTestPrefix } from '../cypress/utils';
     });
 
     context('divider', () => {
-      it(`${getTestPrefix(useTiptap)} render plugin toolbar and change styling`, function () {
+      it.only(`${getTestPrefix(useTiptap)} render plugin toolbar and change styling`, function () {
```
danarish4, line 98:
```suggestion
      it(`${getTestPrefix(useTiptap)} render plugin toolbar and change styling`, function () {
```

danarish4 (approved):
Looks good!

PR #4289, opened 2022-07-28
Tiptap modals UI html height button values

shlomo-gangul:


e2e/tests/__snapshots__/plugins.e2e.ts.snap
```diff
@@ -6037,7 +6038,6 @@ exports[`plugins > html > [draft] add html with src html #0`] =
       "data": {
         "config": {
           "alignment": "center",
-          "height": 517,
           "textWrap": "wrap",
```
maimonav, line 6041: Why is this deleted?

shlomo-gangul: reverted

maimonav (changes requested):
Looks good 🔥 
Please update changelog

maimonav (approved):

PR #4316, opened 2022-08-03
[ricos-plugins] internal: external, mobile & inline toolbars trigger buttonClick event

maimonav:


packages/ricos-types/web/src/toolbarEnums.ts
```diff
@@ -54,6 +54,8 @@ export enum ToolbarType {
   TEXT = 'ALL-TEXT-TOOLBARS',
   SHORTCUT = 'SHORTCUT',
   LINK = 'LINK',
+  EXTERNAL = 'EXTERNAL',
```
lxgreen, line 57: it is bad idea to use enums

lxgreen: they are either weird non-type objects that impact bundle size like bitches (non-const enums), or non-importable strings (const enums). Please use sum type instead:
```
SIDE | MOBILE | ...
```

lxgreen: this whole file should be converted to sum types

lxgreen: we use eums due to ts-proto and wix-proto-gen conversions, and they are banned in other places.

lxgreen: currently we have many issues with schema-derived types due to enums. e.g. Node_Type cannot be imported outside our monorepo

lxgreen (changes requested):

packages/ricos-types/web/src/toolbarEnums.ts
maimonav: This is used by consumers, if we change this file it might break them, no?

lxgreen: how is this used by consumers? it is ricos-types => tiptap

lxgreen: oh it is moved from one of commons?
shit

lxgreen: nevermind then. we'll take care of this later

PR #4325, opened 2022-08-04
[ricos-modals] a11y close modals on esc

tamrat-bay:


danarish4 (approved):
👍

danarish4:
Move logic to modalPopper

PR #4331, opened 2022-08-07
[ricos-plugins]: pluginAddButton tooltip translation

talevy17:


packages/ricos-plugins/web/src/pluginAddButton.tsx
```diff
@@ -81,11 +81,12 @@ export class RicosPluginAddButton implements PluginAddButton {
   }
 
   toToolbarItemConfig(): IToolbarItemConfigTiptap {
+    const { t } = this.services;
     return {
       id: this.button.id,
       type: this.button.modal ? 'modal' : 'toggle',
       presentation: {
-        tooltip: this.button.tooltip,
+        tooltip: t?.(this.button.tooltip) || this.button.tooltip,
```
danarish4, line 89:
```suggestion
        tooltip: t(this.button.tooltip),
```

talevy17: it causes issues with the unit test since t is not defined.
any ideas?

danarish4: There are already mock services for the tests, just add the mock `t` function

PR #4343, opened 2022-08-07
editor-shortcuts: pass modals to command

lxgreen:


packages/plugin-link/web/src/editor.tsx
```diff
@@ -16,6 +21,12 @@ const LinkModalController = lazy(() =>
   }))
 );
 
+const LinkPanelComponent = props => (
```
lxgreen, line 24: please move component to a separate `tsx`, leaving `editor.ts` to be just logic

lxgreen (commented):
see my comment

PR #4344, opened 2022-08-08
Fix/selected resolver

talevy17:


packages/plugin-divider/web/src/getToolbarButtons.tsx
```diff
@@ -45,9 +45,10 @@ const getNodeStyleResolver = {
 
 const selectedNodeResolver = {
   id: 'selectedNode',
-  resolve: content => {
-    if (Array.isArray(content) && content.length > 0) {
-      return content[0];
+  resolve: (_, __, editor) => {
+    const node = (editor?.state?.selection as Record<string, string> | undefined)?.node;
```
danarish4, line 49:
```suggestion
    const node = editor?.state?.selection?.node;
```
This TS declaration is meaningless and just make it hard to read, please remove it from all places

danarish4 (changes requested):
Lets think of exporting `selectedNodeResolver` from one place which can be imported

packages/plugin-divider/web/src/getToolbarButtons.tsx
talevy17: without the ts declaration we get a compilation error since node is not on type `Selection<any>`

talevy17:
> 

I'm not sure where though, `plugin-commons` maybe?

packages/plugin-divider/web/src/getToolbarButtons.tsx
talevy17: nevermind ):

danarish4 (approved):

PR #4352, opened 2022-08-10
[audio] fix settings image upload

tamrat-bay:


packages/plugin-audio/web/src/editor.ts
```diff
@@ -11,13 +11,20 @@ import { getAddButtons } from './getAddButtons';
 
 export const pluginAudio: EditorPluginCreator<AudioPluginEditorConfig> = config => {
   const pluginConfig: AudioPluginEditorConfig = { ...DEFAULTS.config, ...config };
+  const insertAudioHandlers = {
```
danarish4, line 14: why does it need? `getToolbarButtons` already gets the config you can take it from there instead of add more params

packages/plugin-audio/web/src/editor.ts
```diff
@@ -11,13 +11,20 @@ import { getAddButtons } from './getAddButtons';
...
     createPlugin: createAudioPlugin,
     createPluginData: createAudioData,
     tiptapExtensions,
-    toolbar: { getButtons: (config, services) => getToolbarButtons(config, services) },
-    getAddButtons: (config, services) => getAddButtons(config, services),
+    toolbar: {
+      getButtons: (config, services) =>
+        getToolbarButtons({ ...config, ...insertAudioHandlers }, services),
+    },
+    getAddButtons: (_, services) => getAddButtons(config, services),
```
danarish4, line 28: why did you send the initial config instead of the new one?

packages/plugin-audio/web/src/modals/AudioSettings.tsx
```diff
@@ -57,10 +58,12 @@ const AudioSettings: React.FC<Props> = ({
   t,
   isMobile,
   experiments,
-  blockKey = componentData?.id,
   handleFileSelection,
   handleFileUpload,
+  getComponentData,
+  ...props
 }) => {
+  const blockKey = props.blockKey || componentData?.id;
```
danarish4, line 66: There shouldn't be a situation where blockKey isn't defined.
How does it happen? Is there a specific flow?

PR #4360, opened 2022-08-10
fix android enter bug

justame:


justame:
@lxgreen I don't have a permission to Percy, can you take a look and let me know if it's safe to merge?

oded-soffrin:
Fixed on surge in my android 🤩🎉🎊

tsconfig.json
```diff
@@ -10,7 +13,11 @@
     "moduleResolution": "node",
     "target": "ESNext",
     "jsx": "react",
-    "lib": ["dom", "ESNext"],
+    "lib": [
+      "dom",
+      "ESNext"
+    ],
+    "skipLibCheck": true,
```
lxgreen, line 20: i'd limit this flag to specific packages, using the tsconfig extensibility. let's talk

lxgreen (changes requested):

lxgreen:
@yaronn the resolver unit tests fail. this points to incompatibility we have to handle. `skipLibCheck` is dangerous way to go -- we just replace dependency versions and ignore type errors. The type incompatibility usually points to API incompatibility, and we don't know how many obscure bugs added.

PR #4362, opened 2022-08-10
(fix): fonst size in dropdown for header

justame:


packages/ricos-styles/web/src/styles.ts
```diff
@@ -61,8 +62,10 @@ export class RicosStyles implements Styles {
     return this.theme;
   }
 
-  setTheme(theme: RicosTheme) {
-    this.theme = new TextualTheme(theme);
+  setTheme(theme: RicosTheme, isMobile = false) {
```
justame, line 65: @lxgreen let's find a better solution than isMobile if it's gonna be fast.

PR #4374, opened 2022-08-14
Tiptap modals UI vertical embed popover modal width

shlomo-gangul:


danarish4:
The popover component is relevant to all popover modals.
420px is specific width for social-embed.
The change should be in social-embed modal

PR #4397, opened 2022-08-16
[*] remove prosemirror deps

maimonav:


packages/plugin-table/web/src/tiptap/plugins/controllerPlugin.tsx
```diff
@@ -14,7 +14,8 @@ export const controllerPlugin = editor =>
       init: () => {
         return {};
       },
-      apply(tr, prev, oldState, newState) {
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
maimonav, line 17: Define type

packages/editor-common/web/src/Utils/tiptap/isTextSelection.ts
```diff
@@ -0,0 +1,6 @@
+import type { Selection } from 'prosemirror-state';
+
+// replacing isTextSelection of tiptap which is broken
+// https://github.com/ueberdosis/tiptap/issues/2979
+export const isTextSelection = (selection: Selection): boolean =>
+  selection.toJSON().type === 'text';
```
lxgreen, line 6: why not `instanceOf TextSelection`?

lxgreen: also, why editor-common? we want to deprecate it at all. how about ricos-common?

lxgreen (approved):
we saw it all basically. i skipped the table changes for now

packages/editor-common/web/src/Utils/tiptap/isTextSelection.ts
maimonav: 1. instanceof TextSelection uses prosemirror-state, which costs something like 60KB
https://github.com/wix/ricos/runs/7861688823?check_suite_focus=true#step:7:66
2. We can easy change it to ricos-common when we deprecate editor-common, now it maybe be resulted in bundle increment and I would like to merge it soon.

PR #4403, opened 2022-08-16
[ricos-schema] fix: add isAdsense to html schema

talevy17:


danarish4 (changes requested):
Please add test (add adsense to migration-content)

PR #4430, opened 2022-08-21
[tiptap] fix mention crash on oneApp

danarish4:


packages/plugin-mentions/web/src/tiptap/suggestion.ts
```diff
@@ -47,38 +46,45 @@ export default (
...
         },
 
         onKeyDown(props) {
           if (props.event.key === 'Escape') {
-            popup[0].hide();
-            handleDropdownClose?.();
+            if (handleDropdownClose) {
+              handleDropdownClose();
+            } else {
+              this.onExit();
```
oded-soffrin, line 86: looks like previously when we hit Escape it only hide the tooltip, and now it calls > onExti > component.destory()

do we even need the "component" when we have the external one? (also related to component update props, etc..)

examples/main/shared/editor/EditorPlugins.tsx
```diff
@@ -539,11 +539,11 @@ export const config: RichContentEditorProps['config'] = {
     visibleItemsBeforeOverflow: 5,
     popoverComponent: <div />,
     // eslint-disable-next-line no-console
-    handleDropdownOpen: () => console.log('mentions dropdown opened'),
+    // handleDropdownOpen: () => console.log('mentions dropdown opened'),
```
oded-soffrin, line 542: the comment were actually an easy way to see the callback is working in tiptap, why to remove'em?

PR #4448, opened 2022-08-22
[video] resolve settings button visibility

tamrat-bay:


packages/plugin-video/web/src/getToolbarButtons.tsx
```diff
@@ -59,6 +59,12 @@ export const getToolbarButtons = (config, services): ToolbarButton[] => {
         id: videoModals.settings,
         Component: VideoSettingsModal,
       },
+      attributes: {
+        visible: {
+          id: 'IS_VIDEO_SETTINGS_BUTTON_VISIBLE',
+          resolve: () => !!config.shouldShowSpoiler,
```
danarish4, line 65:
```suggestion
          resolve: () => {
          const spoilerPlugin = services.plugins.asArray().find(plugins => plugin.type === spoiler)
          return !spoilerPlugin.getConfig().supportedPlugins || spoilerPlugin.getConfig().supportedPlugins.includes(video) 
          },
```

danarish4 (changes requested):
You can use the `EditorPlugins` entity for checking if a spoiler plugin exists.
you can revert the changes in `helpers-config` file

packages/plugin-video/web/src/getToolbarButtons.tsx
tamrat-bay: we don't have plugins in services
services.plugins === undefined

danarish4: It should be there
https://github.com/wix/ricos/blob/master/packages/ricos-types/web/src/services.ts#L13

PR #4449, opened 2022-08-23
[ricos-common-plugins, toolbars-v3] fix: bold & italic toggles resolvers & commands

maimonav:


packages/ricos-common-plugins/web/src/bold/extension.ts
```diff
@@ -101,18 +101,18 @@ export const bold: RicosExtension = {
...
             () =>
             ({ commands, state }) => {
-              const isToggleOff =
-                commands.getStylesDecorationBySelectedNode(this.name)?.fontWeightValue ===
-                FONT_WEIGHT_BOLD;
-              const fontWeightValue = isToggleOff ? FONT_WEIGHT_NORMAL : FONT_WEIGHT_BOLD;
+              const decoration: Decoration | undefined = commands.getStylesDecorationBySelectedNode(
+                this.name
+              );
+              const isToggleOff = decoration?.fontWeightValue !== FONT_WEIGHT_BOLD;
```
lxgreen, line 107: it is opposite condition now?

lxgreen: `isToggleOff` name means that current `toggleBold` call is `toggle Bold off`

lxgreen: now it publishes `pluginDelete` on every toggle -- both on and off.

lxgreen (changes requested):
the event publishing is wrong

PR #4451, opened 2022-08-23
[toolbars-v3] undo/redo - handle disabled

noamtwix:


packages/ricos-plugins/web/src/plugin-text-button.ts
```diff
@@ -167,7 +167,7 @@ export class PluginTextButton implements FormattingToolbarButton {
           : this.button.command?.(editorCommands)({});
       },
       isActive: () => !!attributes.active && !!content.resolve(attributes.active),
-      isDisabled: () => false,
+      isDisabled: () => !!attributes.disabled && !!content.resolve(attributes.disabled),
```
maimonav, line 170: Please add this to `toToolbarItemConfig` as well.

maimonav (changes requested):

packages/ricos-plugins/web/src/plugin-text-button.ts
noamtwix: For what purpose?

maimonav: For future buttons that will support disabled state, for example, undo/redo in mega toolbar or indent/outdent on code block.
It can also be added later on, what ever you think is best.

maimonav:
@noamtwix I've checked the undo/redo buttons, they are disabled but they are not appear disabled like draft, can you please check it also?

PR #4456, opened 2022-08-23
[ricos-plugins] internal: open external buttons modals according to layout (drawer/toolbar/dialog)

maimonav:


lxgreen (changes requested):
The `EditorPlugins` aggregate gets the `context` object within its `services`. That object contains all the required contextual info including `isMobile`.  Example usage can be seen [here](https://github.com/wix/ricos/blob/master/packages/ricos-plugins/web/src/pluginAddButton.tsx#L116)

So, every button (both `textButton` or `addButton`) is able to initialize its modal with proper layout without additional parameters. In general, we should think twice before adding parameters to entity methods. Currently, the thumb rule is to inject dependencies via constructor, so typically we should not need to add parameters to their methods.

maimonav:
@lxgreen Amazing, thanks! I was not aware of it, please take a second look :)

packages/ricos-plugins/web/src/pluginAddButton.tsx
```diff
@@ -159,20 +159,24 @@ export class RicosPluginAddButton implements PluginAddButton {
...
         return this.button.modal
           ? modals?.isModalOpen(this.button.modal.id)
             ? modals?.closeModal(this.button.modal.id)
             : modals?.openModal(this.button.modal.id, {
-                layout: 'dialog',
+                positioning: {
+                  placement: 'bottom',
+                  referenceElement: e?.target,
+                },
+                layout: context.isMobile ? 'drawer' : e?.target ? 'toolbar' : 'dialog',
```
danarish4, line 179:
```suggestion
                layout: context.isMobile ? 'drawer' : e?.target ? 'popover' : 'dialog',
```

maimonav: 👑 👑 👑 👑

PR #4464, opened 2022-08-24
(fix): modal position

justame:


maimonav:
@justame the issue is still reproduced in the footer toolbar.

maimonav (approved):

maimonav:
Screen Shot 2022-08-25 at 14 22 25

PR #4497, opened 2022-08-27
(fix): editor height 100% && fullRicosEditor click focus back to editor

justame:


maimonav:
@justame Tests are failing, please check it out.

packages/ricos-editor/web/src/tiptap/FullRicosEditor.tsx
```diff
@@ -76,7 +78,13 @@ export class FullRicosEditor extends React.Component<Props, State> {
...
       () => this.inputRef.current as HTMLInputElement
     );
-
+    this.fullRicosEditorElement?.addEventListener(
+      'click',
+      () => {
+        this.orchestrator.getServices().editor.adapter.focus();
+      },
+      true
+    );
```
maimonav, line 87: Why do we need to add an event listener?
I believe `height: 100%` should be enough, clicking the element should trigger focus.

PR #4520, opened 2022-08-29
[tiptap] node original size style

danarish4:


packages/tiptap-extensions/web/src/styles.tsx
```diff
@@ -61,9 +62,14 @@ export const getComponentStyles = ({ componentData, theme, isFocused, isMobile }
   };
 };
 
+const originalSizeMapper = {
```
justame, line 65: good solution for not but not generic enough, just add a task on the backlog to find a better solution.

danarish4: Another example of why we need the ability to configure extensions externally

PR #4524, opened 2022-08-30
[ricos-plugins] fix size alignment & separator buttons visibility

tamrat-bay:


danarish4:
Looks good :)
I think we can move the separator resolver [to this file ](https://github.com/wix/ricos/blob/master/packages/plugin-commons/web/src/Utils/contentResolvers.ts)- I'm not sure if this is the right package, but let's put all plugin's common resolvers here and decide where they should go later.

It would also be better to change the resolver name to getVisibleOnlyForDesktopResolver - this will make the code clear to everyone who encounters it in the future.

`getVisibleOnlyOnDesktopResolver = isMobile => {
           id: 'VISIBLE_ONLY_ON_DESKTOP'
           resolve: () => !isMobile
`

PR #4527, opened 2022-08-30
[toolbars-v3] disable formatting buttons when plugin selected

noamtwix:


maimonav (approved):
@noamtwix Looks great 🔥

PR #4541, opened 2022-09-01
[ricos-editor] add missing locale prop

tamrat-bay:


danarish4:
Please take locale from context instead of passing it

https://github.com/wix/ricos/blob/master/packages/tiptap-editor/web/src/components/RicosTiptapEditor/RicosTiptapEditor.tsx#L37

PR #4543, opened 2022-09-01
[ricos-styles, text-color] did update theme/doc style, add getColors to theme, fix text/highlight color resolver that will consider theme when selection is collapsed

maimonav:


packages/ricos-styles/web/src/models/styles.ts
```diff
@@ -32,6 +32,13 @@ export interface TextualTheme {
    * @memberof TextualTheme
    */
   toStyleTag(): ReactElement;
+  /**
+   * Returns palette colors for the current theme
+   *
+   * @returns  {PaletteColors}
+   * @memberof TextualTheme
+   */
+  getColors(): PaletteColors | undefined;
```
lxgreen, line 41: why `| undefined`? 
can we have default palette colors, like color/background that you defined in default theme?  every `undefined` makes our types a bit weaker

packages/ricos-editor/web/src/tiptap/RicosStyles.tsx
```diff
@@ -39,6 +49,17 @@ const RicosStylesRenderer: FC<{
     };
   }, []);
 
+  useEffect(() => {
```
lxgreen, line 52: didn't we remove the `usePrevious` yesterday?

lxgreen (approved):
nice

packages/ricos-styles/web/src/models/styles.ts
maimonav: Thanks, fixed this in [here](https://github.com/wix/ricos/pull/4544)

PR #4558, opened 2022-09-04
[ricos-editor] fix: editor with toolbars (side/footer)

maimonav:
This PR aims to solve editor height with side/footer toolbars positions.
The toolbars should be siblings of the editor, so the idea is simple - fix the html structure.
**This is just a simple solution to this issue,** a more fundamental fix should be having all toolbars entities seperated and then we can render each one according to design.


danarish4 (approved):
Looks good as a temporary solution.
We need to extract all the toolbar settings calc  - do we have asana ticket for this?

PR #4579, opened 2022-09-06
[tiptap-table] fix converters

danarish4:


maimonav (changes requested):
Looks good 🔥.
@danarish4 Please add unit tests to make sure we fully test conversion logic, you can add it in the current `table-converter.spec.ts`.

PR #4595, opened 2022-09-09
rceNext API

oded-soffrin:
allow consumers to use `rceNext` to enable tiptap instead of experiment overriden


maimonav:
@oded-soffrin Looks great 🔥 , can you please document this `rceNext` prop?

oded-soffrin:
> @oded-soffrin Looks great 🔥 , can you please document this `rceNext` prop?

Good point. we'll add it officially once we're done with the effort though

PR #4602, opened 2022-09-11
[table] feat: formatting button

talevy17:


packages/ricos-editor/web/src/tiptap/RicosToolbars.tsx
```diff
@@ -84,6 +84,25 @@ class RicosToolbars extends React.Component<
...
+    modalService.register({
+      id: 'ricos-formatting-toolbar',
+      Component: () =>
+        this.floatingToolbarChildren(
+          'ricos-formatting-toolbar',
+          tiptapEditor.view.dom.clientWidth,
+          toolbarItemsConfig
+        ),
+    });
+  };
```
lxgreen, line 104: unacceptable. any service registration should be performed within orchestrator initialization flow, never in UI

lxgreen: the toolbar registration is big deal. please don't merge this PR before we talk. + @danarish4

lxgreen (changes requested):

packages/ricos-editor/web/src/tiptap/RicosToolbars.tsx
talevy17: moved it to `ricos-plugins` more specifically `PluginTextButtons` (the same way we handle the plugin menu registration in `PluginAddButtons`). WDYT? @lxgreen

PR #4607, opened 2022-09-11
[common] fix: link theme text decoration changed from none to underline

maimonav:


oded-soffrin:
Why closed?

maimonav:
I've already fix it in a different way in other PR.

PR #4609, opened 2022-09-12
Video seo

oded-soffrin:
This PR takes care of SEO team request for embedded videos
works in draft, but we need to populate used fields in tiptap as discussed @lxgreen @danarish4


lxgreen (approved):
Just a question about `seoMode` - is it editor-level prop? I don't rememeber we support it in tiptap. If it's a plugin-level config, does it make sense to lift it to editor/viewer level, since it could be cross-plugin setting?

oded-soffrin:
It's a viewer level prop

danarish4:
> This PR takes care of SEO team request for embedded videos works in draft, but we need to populate used fields in tiptap as discussed @lxgreen @danarish4

populate used fields in tiptap was merged

PR #4614, opened 2022-09-13
[plugin-common]fix: color picker bottom margin fix

shlomo-gangul:


maimonav:
@shlomo-gangul Hey, as we discussed, this is not the way to go.
You need to talk to Liam & Judith and understand what's the product & design.

PR #4615, opened 2022-09-13
[editor-query/toolbars-v3] hide plus button on plugin selection

tamrat-bay:


packages/ricos-content-query/web/src/editor-query/queries/is-text-block-node.ts
```diff
@@ -0,0 +1,11 @@
+import type { QueryItem } from 'ricos-types';
+
+const NAME = 'isTextBlockNode';
+const isTextBlockNode: QueryItem =
+  ({ editorQuery }) =>
+  () => {
+    const selectedNode = editorQuery.selection().doc.children[0];
```
danarish4, line 7: what if the doc has no children?
```suggestion
    const selectedNode = editorQuery.selection().doc?.children?.[0];
```

danarish4 (changes requested):
follow comment

danarish4:
1. It makes no sense to call `services.getEditorQueryService` (I would expect it to be `services.editorQuery`). Let's add the 
  EditorQuery service to the toolbar context and use it directly from there.

2. Let's think about a more understandable name for this query (maybe isAtomicPluginSelection?)

3. There is a problem with the query's logic - what if the selection includes both `textBlock` & `atomicPlugin` ? 
    children[0] will return true

packages/ricos-content-query/web/src/editor-query/queries/is-single-root-textblock.ts
```diff
@@ -0,0 +1,10 @@
+import type { QueryItem } from 'ricos-types';
+
+const NAME = 'isSingleRootTextBlock';
+const isSingleRootTextBlock: QueryItem =
+  ({ editorQuery }) =>
+  () => {
+    return !!editorQuery.selection().doc?.children?.[0].isTextBlock();
```
danarish4, line 7: This does not have to be single like this.. you should add
```suggestion
    return editorQuery.isMultipleSelection() && !!editorQuery.selection().doc?.children?.[0].isTextBlock();
```

PR #4620, opened 2022-09-14
chore(schema): remove disallowed optional label

oded-soffrin:
When trying to sync the scehma, I got this error:
`[19:58:06.730] wix/rich_content/v1/plugin_html.proto:22:12: Explicit 'optional' labels are disallowed in the Proto3 syntax. To define 'optional' fields in Proto3, simply remove the 'optional' label, as fields are 'optional' by default.`

when remove the `optinal` I got ts error, fixed here
@lxgreen @talevy17


talevy17 (approved):

packages/ricos-converters/web/src/nodes/html-converter.ts
```diff
@@ -24,6 +26,7 @@ export const htmlConverter: TiptapNodeConverter = {
         nodes: [],
         htmlData: {
           ...(data as HTMLData),
+          source: node.attrs?.source || 'HTML',
```
maimonav, line 29: @oded-soffrin Please change it to `HTMLData_Source.HTML`, we should avoid using string values if we can get it directly from schema.

oded-soffrin: this PR is already merge,
but so here it is - https://github.com/wix/ricos/pull/4628

PR #4621, opened 2022-09-14
[ricos-plugins, toolbars-v3] fix: register horizontal plugin menu in ricos-plugins

talevy17:


packages/ricos-plugins/web/src/plugin-add-buttons.tsx
```diff
@@ -110,6 +109,10 @@ export class RicosPluginAddButtons implements PluginAddButtons {
         <AddPluginMenu addButtons={this} addPluginMenuConfig={config} {...props} />
       ),
     });
+    this.services.modals?.register({
```
tamrat-bay, line 112: This is something to consider
Only one modal can be registered based on the config

(!addPluginMenuConfig || `addPluginMenuConfig?.horizontalMenuLayout` ?AddPluginMenuHorizontal : AddPluginMenu)

talevy17: @tamrat-bay we need them both in some cases (table uses the horizontal plugin menu regardless of the "main" plugin menu)

tamrat-bay (approved):

PR #4623, opened 2022-09-14
[collapsible-list] fix: render plugin menu inside collapsible list body

talevy17:


packages/toolbars-v3/web/src/modals/FloatingPluginMenu/PlusButton.tsx
```diff
@@ -9,12 +9,22 @@ type PlusButtonProps = {
...
     <button
       data-hook={'addPluginFloatingToolbar'}
       ref={ref}
-      onClick={onClick}
+      onClick={event => {
+        handleMouseAction(event);
+        setTimeout(() => onClick(), 0);
+      }}
+      onMouseDown={handleMouseAction}
+      onMouseUp={handleMouseAction}
```
danarish4, line 27: ???

danarish4:
Instead of dirtying the generic plus button, it is better to have the logic to fix the collapsible bug in the collapsible list body component

Please wrap the plus button in logic and give it an informative name, such as:

```
WithModalFocusFix = (
<div onMouseDown={e=>e.preventDefault}>
<PlusButton />
</div>
```

danarish4 (changes requested):
Instead of dirtying the generic plus button, it is better to have the logic to fix the collapsible bug in the collapsible list body component

Please wrap the plus button in logic and give it an informative name, such as:

```
WithModalFocusFix = (
<div onMouseDown={e=>e.preventDefault}>
<PlusButton />
</div>
```

PR #4637, opened 2022-09-18
[ricos-common-plugins] fix: add underline after autoLink

maimonav:


packages/ricos-common-plugins/web/src/underline/extension.ts
```diff
@@ -34,6 +35,32 @@ export const underline: RicosExtension = {
...
+                  const autoLinkStep = steps?.[steps.length - 1]?.toJSON() || {};
+                  from = autoLinkStep.from;
+                  to = autoLinkStep.to;
+                  return;
+                }
+              });
+              if (!!from && !!to) {
+                return newState.tr
+                  .addMark(from, to, this.type.create())
+                  .removeStoredMark(this.type);
```
lxgreen, line 57: what does this `removeStoredMark` do?

lxgreen (approved):

packages/ricos-common-plugins/web/src/underline/extension.ts
maimonav: Remove the underline mark from the further text, if we don't remove it, when you enter more characters, they will also have the underline mark.

PR #4652, opened 2022-09-20
[ricos-common-plugins]fix:qoute font get unwired

shlomo-gangul:


packages/ricos-common-plugins/web/src/blockquote/extension.ts
```diff
@@ -80,7 +80,7 @@ export const blockquote: RicosExtension = {
         return [
           'blockquote',
           attrs,
-          ['span', { ...(shouldAddDir ? { dir: 'auto' } : {}) }, 0],
+          ['span', { class: styles.paragraph, ...(shouldAddDir ? { dir: 'auto' } : {}) }, 0],
```
maimonav, line 83: This looks like it fixes the issue, but this not the best way to do this, you specific use the `.paragraph` css class.
If you move it, or change the name of it, it will break.
What I suggest you is to take a look at [this](https://github.com/wix/ricos/blob/master/packages/ricos-common-plugins/web/src/statics/styles.scss#L17) , this applies the typography over paragraph, and you can do is similar to the `.quote` css class.

maimonav (changes requested):

PR #4653, opened 2022-09-20
[collapsible-list]: fix tiptap state management and plugin menu position

talevy17:


packages/plugin-collapsible-list/web/src/tiptap/utils.ts
```diff
@@ -24,28 +24,37 @@ export function isInCollapsibleList(editor) {
...
 export function isInCollapsibleListBody(editor, nodeId) {
-  return !!findParentNodeClosestToPos(
+  const closestCollapsibleListBody = findParentNodeClosestToPos(
     editor.state.selection.$to,
-    node => node.type.name === TIPTAP_COLLAPSIBLE_ITEM_BODY_TYPE && node.attrs.id === nodeId
+    node =>
+      node.type.name === TIPTAP_COLLAPSIBLE_ITEM_BODY_TYPE ||
+      node.type.name === TIPTAP_COLLAPSIBLE_LIST_TYPE
   );
+  return closestCollapsibleListBody?.node.attrs.id === nodeId;
```
danarish4, line 33: it will return true also for collapsible title no?

talevy17: nope, because the nodeId won't match.
that way we can stop searching if a collapsibleList was found (instead of going to an external list)

packages/plugin-collapsible-list/web/src/tiptap/utils.ts
```diff
@@ -24,28 +24,37 @@ export function isInCollapsibleList(editor) {
...
-  if (node.type.name === TIPTAP_COLLAPSIBLE_LIST_TYPE) {
-    return node.content.content;
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  const nodes: Record<string, any>[] = [];
+  state.doc.descendants(node => {
+    if (node.attrs.id === nodeId) {
+      nodes.push(node);
+    }
+  });
+  if (nodes[0]?.type.name === TIPTAP_COLLAPSIBLE_LIST_TYPE) {
```
danarish4, line 44: we have the function `findNodeById`, the id is unique so you don't need to recheck the node type.
this is a classic example of using editorQuery

packages/ricos-converters/web/src/nodes/collapsible-list-converters.spec.ts
```diff
@@ -28,6 +28,8 @@ describe('collapsibeListNode converter', () => {
         type: Node_Type.COLLAPSIBLE_ITEM,
         attrs: {
           id: '96',
+          isExpanded: true,
```
danarish4, line 31: how does the change not impact on ricos node as well?

packages/ricos-converters/web/src/nodes/collapsible-list-converters.ts
```diff
@@ -5,19 +5,32 @@ import type {
...
+          attrs: {
+            id: collapsibleItem.id,
+            isExpanded:
+              expandState === CollapsibleListData_InitialExpandedItems.ALL ||
+              (index === 0 && expandState === CollapsibleListData_InitialExpandedItems.FIRST),
+            expandOnlyOne: node.collapsibleListData.expandOnlyOne,
+          },
+          content: visit(collapsibleItem),
+        };
+      }),
```
danarish4, line 33: what about fromTiptap?

packages/ricos-converters/web/src/nodes/collapsible-list-converters.spec.ts
talevy17: because, isExpanded is not part of `collapsibleItem` schema

packages/ricos-converters/web/src/nodes/collapsible-list-converters.ts
talevy17: I'm not sure its needed, its not part of the collapsibleItem schema and is derived from the `CollapsibleList` data which already exists there so we don't lose any information - the isExpanded property is a stateful property

PR #4660, opened 2022-09-22
[ricos-editor] create a separate draft portal

tamrat-bay:


oded-soffrin:
אני רק שאלה
why can't both draft & tiptap use ricosPortal?
what's "tiptap' about it? 

also this doesn't yet solve the blog's bug

lxgreen:
can i have some context? what's this portal for and why do we work on draft?

oded-soffrin:
> may i have some context? what's this portal for and why do we work on draft?

https://jira.wixpress.com/browse/RICH-798

PR #4663, opened 2022-09-25
[ricos-common-plugin]feat:added debounce to resizer

shlomo-gangul:


packages/ricos-common-plugins/web/src/resizable/resizer.tsx
```diff
@@ -97,16 +97,17 @@ const ResizerHOC = Component => {
...
-          ...props.componentData.containerData,
-          width: {
-            custom: parseInt(elementRef.style.width),
+    const onResizeStop = () =>
+      debounce((event, direction, elementRef) => {
+        updateAttributes({
+          containerData: {
+            ...props.componentData.containerData,
+            width: {
+              custom: parseInt(elementRef.style.width),
```
lxgreen, line 106: `parseInt` accepts second parameter called `radix`

please check what's this, and pass correct one

packages/ricos-common-plugins/web/src/resizable/resizer.tsx
```diff
@@ -97,16 +97,17 @@ const ResizerHOC = Component => {
...
     }, 30);
 
-    const onResizeStop = (_event, _direction, elementRef) => {
-      updateAttributes({
-        containerData: {
-          ...props.componentData.containerData,
-          width: {
-            custom: parseInt(elementRef.style.width),
+    const onResizeStop = () =>
+      debounce((event, direction, elementRef) => {
```
lxgreen, line 101: `lodash` in general is bad, and `debounce` is unsafe.

Consider implementing `debounce` that is capable of error handling (i.e. if exception is thrown by `debounce`d code, it should be able to catch).

lxgreen (changes requested):

packages/ricos-common-plugins/web/src/resizable/resizer.tsx
lxgreen: in addition, this `debounce` does not fix the performance issue. the root cause is throttled `onResize`. This is cimpletely redundant to update attributes during the resize.

PR #4669, opened 2022-10-03
fix(panels): remove loading message

oded-soffrin:


oded-soffrin:
@noamtwix FYI, no need for "loading" messages (it bothered me enough to take the time to fix it)

PR #4671, opened 2022-10-04
[theme] add editor custom styles

tamrat-bay:


packages/tiptap-editor/web/src/statics/styles/tiptap-editor-styles.scss
```diff
@@ -29,7 +29,10 @@
 }
 
 .ricosTiptapEditor {
-  height: 100%;
+  min-height: var(--ricos-custom-editor-min-height, '100%');
+  max-height: var(--ricos-custom-editor-max-height, '100%');
+  overflow-y: auto;
+  padding: 0 20px;
```
oded-soffrin, line 35: This style need to be applied only when using maxHeight

packages/toolbars-v3/web/src/components/FloatingToolbar/FloatingToolbar.tsx
```diff
@@ -34,8 +34,10 @@ export const FloatingToolbar = ({
   const [dummyUpdate, setForceUpdate] = useState<number>(1);
   const [isModalOpen, setModalOpen] = useState<boolean>(false);
 
+  const domElement = editor.view.dom.parentElement?.parentElement;
```
oded-soffrin, line 37: the maxHeight is implemented on the parent of the parent of the editor dom,
need to either move the css attributes down to contentEditable, or find a better way to get the dom element here

oded-soffrin (changes requested):
some comments above
also - footer toolbar padding is 60px which looks bad in this case. is this by design?

packages/tiptap-editor/web/src/statics/styles/tiptap-editor-styles.scss
tamrat-bay, line 35: @lxgreen @oded-soffrin  @danarish4 
wouldn't it result in double padding for consumers who already add padding?

lxgreen: @tamrat-bay this is just a quick check. we definitely need to add these rules only if min/maxHeight is applied.

tamrat-bay: How about we use JS to check for the custom maxHeight style 
and then use that to apply the rest of the CSS.
I've used it locally and it works fine

lxgreen: agree, we can check the theme and apply certain CSS classes. I'd say it should be responsibility of Styles aggregate (`ricos-styles`)

tamrat-bay: @lxgreen @oded-soffrin 
I have pushed my changes, let me know your thoughts

examples/main/src/ExampleApp.tsx
```diff
@@ -279,13 +279,15 @@ class ExampleApp extends PureComponent<ExampleAppProps, ExampleAppState> {
 
   getThemeObj = (activeFont, activePalette) => {
     const themeObj: RicosTheme = {
+      customStyles: {
+        editor: { minHeight: '200px', maxHeight: '200px' },
```
lxgreen, line 283: TODO: remove this style before merge

examples/main/src/ExampleApp.tsx
```diff
@@ -279,13 +279,15 @@ class ExampleApp extends PureComponent<ExampleAppProps, ExampleAppState> {
...
+      },
       settingsStyles: {
         modals: {
           baseZIndex: 1000,
         },
       },
     };
```
lxgreen, line 290: TODO: revert activeFont

packages/tiptap-editor/web/src/components/RicosTiptapEditor/RicosTiptapEditor.tsx
```diff
@@ -64,6 +64,11 @@ export const RicosTiptapEditor: FunctionComponent<RicosTiptapEditorProps> = ({
 
   const { containerClassName, editorClassName } = editorStyleClasses || {};
 
+  const hasCustomEditorHeight = () => {
+    const styles = window.getComputedStyle(document.body);
+    return styles.getPropertyValue('--ricos-custom-editor-max-height');
+  };
```
lxgreen, line 70: this should be `Styles` [aggregate](https://github.com/wix/ricos/blob/master/packages/ricos-styles/web/src/styles.ts#L32) responsibility

lxgreen (changes requested):

packages/tiptap-editor/web/src/components/RicosTiptapEditor/RicosTiptapEditor.tsx
oded-soffrin: regardless of ricos-styles, there's no point in setting the css and than checking via css
we get the theme as JS object we can check the config directly
in addition, `window` doesn't exist on SSR

maybe it's better to sit on it today if we're all in the office ?

packages/toolbars-v3/web/src/components/FloatingToolbar/FloatingToolbar.tsx
tamrat-bay: we can only move the css attributes on level down to `editor.view.dom.parentElement`

examples/main/shared/editor/Editor.tsx
```diff
@@ -184,7 +184,7 @@ export default class Editor extends PureComponent<ExampleEditorProps> {
             content={contentState}
             injectedContent={injectedContent}
             locale={locale}
-            cssOverride={theme}
+            // cssOverride={theme}
```
tamrat-bay, line 187: TODO: revert before merge

packages/toolbars-v3/web/src/components/FloatingToolbar/FloatingToolbar.tsx
tamrat-bay: @lxgreen 

@oded-soffrin  and I discussed this and decided to leave it this way for now
as well, I am working on improving the solution using refs in the future

packages/common/web/statics/styles/palette.scss
```diff
@@ -47,6 +47,8 @@ $ricos-code-block-font-size: var(--ricos-custom-code-block-font-size, 16px);
 $ricos-code-block-line-height: var(--ricos-custom-code-block-line-height, 26px);
 $ricos-code-block-padding: var(--ricos-custom-code-block-padding, 2px 25px);
 $ricos-code-block-margin: var(--ricos-custom-code-block-margin, 15px 18px);
+$ricos-custom-editor-min-height: var(--ricos-custom-editor-min-height, '100%');
```
tamrat-bay, line 50: remove fallback 100%

packages/ricos-editor/web/src/utils/createEditorStyleClasses.ts
```diff
@@ -16,12 +18,14 @@ export const createEditorStyleClasses: CreateEditorStyleClasses = ({
...
   editorCss = {},
   experiments = {},
+  theme = {},
 }) => {
   const themeDesktopStyle = cssOverride.desktop
     ? { [cssOverride.desktop]: !isMobile && cssOverride && cssOverride.desktop }
     : {};
   const containerClassName = classNames(editorCss.wrapper, cssOverride.wrapper, {
     [editorCss.desktop]: !isMobile,
+    [editorCss.custom_editor]: theme?.customStyles?.editor?.maxHeight,
```
tamrat-bay, line 28: refactor - make it easier to read

PR #4699, opened 2022-10-19
[emoji] fix modal scroll

tamrat-bay:


danarish4 (approved):
Looks good.
Also, we have an emoji button for mobile on tiptap, but we filter it out on draft.
Mobile emoji modal is broken, let's filter this button as well

PR #4736, opened 2022-10-31
Editor x fixes

justame:


packages/tiptap-editor/web/src/components/RicosTiptapEditor/RicosTiptapEditor.tsx
```diff
@@ -55,6 +55,16 @@ export const RicosTiptapEditor: FunctionComponent<RicosTiptapEditorProps> = ({
...
     editor.on('transaction', forceUpdate);
+
+    editor.on('transaction', ({ editor }) => {
+      onUpdate?.({
+        content: tiptapToDraft(
+          editor.getJSON() as JSONContent,
+          experiments.removeRichContentSchemaNormalizer?.enabled
+        ),
+      });
+    });
```
maimonav, line 66: @justame Let's talk about it tomorrow, if I understand correctly, on every transaction it triggers `onChange`, it can very noisy to consumers.

PR #4776, opened 2022-11-09
[image] original width style

danarish4:


talevy17 (commented):
doesnt this mean we should also delete this part from the old viewer and an adjustment on the editor side too? @danarish4

danarish4:
> doesnt this mean we should also delete this part from the old viewer and an adjustment on the editor side too? @danarish4

We can remove it and it will work, but we are going to delete all this code anyway

PR #4779, opened 2022-11-10
[rcv-next] fix: performance

talevy17:


packages/ricos-viewer-next/web/src/rericos/renderers/text/text.tsx
```diff
@@ -18,4 +18,9 @@ const TextRenderer: FC<TextRendererProps> = ({ node }) => {
   );
 };
 
+const TextRenderer = ({ node }: TextRendererProps) => {
+  const key = Math.random().toString(36).slice(2, 9);
+  return <Text key={key} node={node} />;
```
danarish4, line 23: What about this?

```suggestion
  return <Text key={`text-${node.id}`} node={node} />;
```

packages/ricos-viewer-next/web/src/rericos/renderers/text/text.tsx
```diff
@@ -18,4 +18,9 @@ const TextRenderer: FC<TextRendererProps> = ({ node }) => {
   );
 };
 
+const TextRenderer = ({ node }: TextRendererProps) => {
+  const key = Math.random().toString(36).slice(2, 9);
```
danarish4, line 22:
```suggestion
```

packages/ricos-viewer-next/web/src/rericos/renderers/text/text.tsx
talevy17: @danarish4 text nodes have empty ids

PR #4801, opened 2022-11-14
[video ] render iframe player in ssr

tamrat-bay:


packages/plugin-video/web/src/video-viewer.jsx
```diff
@@ -125,7 +137,9 @@ class VideoViewer extends Component {
       ...this.disableDownloadProps(),
     };
 
-    return (
+    return isSSR() ? (
```
danarish4, line 140:
```suggestion
    return isSSR() && isSeo() ? (
```

danarish4 (changes requested):
We want to display the iframe only in case of SSR & SEO

PR #4836, opened 2022-11-24
[ricos-modals] modals navigations using arrow keys

noamtwix:
As part of the a11y improvement, we need to navigate throw modals with arrow keys.
This solution uses an external npm library: [emulate-tab](https://www.npmjs.com/package/emulate-tab) (emulate-tab [bundle size](https://bundlephobia.com/package/emulate-tab@1.2.1)).
The solution is generic and has an impact on all `ricos-modals`.
image

Custom headings - 'update modals' are handled using left / right arrow keys.
image


packages/editor-common/web/src/consts.ts
```diff
@@ -95,6 +95,11 @@ export const KEYS_CHARCODE = {
   ENTER: 13,
   ESCAPE: 27,
   SPACE: 32,
+  TAB: 9,
```
noamtwix, line 98: move to custom-key-codes.ts

PR #4837, opened 2022-11-26
[schema] field deprecation

lxgreen:
Field | Deprecated Name | New Name
----- | --------------- | --------
`AppEmbedData.BookingData.durations: string` | durations | duration
`TableData.dimensions.rowsHeight: number[]` | rowsHeight | rowHeights
`Decoration.italic_data: boolean` | italic_data | italic
`Decoration.underline_data: boolean` | underline_data | underline


oded-soffrin:
First of all, love the diagram. I think it's a good idea to add it to one of our project README,
also, as this a breaking change what's the plan? I assume not V2 so support in API / Converters both old & new values?
lastly, since it has a big "price" and effect, is it a must? should we double check if this consider to be a "blocker' for platformization team ?

lxgreen:
@oded-soffrin we can make it non-breaking, just supporting both names in content. the changes are expensive, not sure how "must" they are for the Blog team

oded-soffrin:
For visibility - we decided not to introduce those non blocker changes, need to keep in mind for V2

PR #4839, opened 2022-11-27
[editor-common] fix: anchor link in draft

maimonav:


oded-soffrin (approved):
When did it break?

maimonav:
@oded-soffrin I'm not sure, by I think long ago, we need to investigate.

PR #4859, opened 2022-12-04
[toolbars-v3] fix: plus icon position when focus move from plugin to text

noamtwix:
When you are in focus on a plugin (image for example), and then click on a text paragraph, the plus icon appears at the top of the editor


packages/toolbars-v3/web/src/modals/FloatingPluginMenu/PlusButton.tsx
```diff
@@ -25,4 +30,11 @@ const PlusButton = React.forwardRef(
   )
 );
 
+const PlusButton = React.memo(PlusButtonForwardRef, (props, nextProps) => {
```
maimonav, line 33: Suggestion:
More new/common way to do so is with `useEffect` hook.

maimonav (approved):
Looks good 🔥 
Please take a look at the comment, and update changelog :)

packages/toolbars-v3/web/src/modals/FloatingPluginMenu/PlusButton.tsx
noamtwix: This is the way to do `shouldComponentUpdate` ([reference](https://reactjs.org/docs/hooks-faq.html#how-do-i-implement-shouldcomponentupdate))

talevy17 (approved):

PR #4873, opened 2022-12-06
chore(image-utils): cleanup

oded-soffrin:
Cleaning up unused features in image-utils and remove preload option all together (I believe it was used by mistake)

@danarish4 once we introduce @wix/image everywhere there's more cleaning up we can do


oded-soffrin:
this PR should be irrelevant due to https://github.com/wix/ricos/pull/4941

oded-soffrin:
@domakas ready to merge

PR #4899, opened 2022-12-14
[tiptap-extensions] fix: android bug - new line does not preserved textStyle

noamtwix:
This fixes a bug with Android when `textStyle` is not preserved in new lines.
The styles arrived in `node.attrs.style` as a string, so we need to create the `textStyle` object manually.


packages/tiptap-extensions/web/src/unique-id.tsx
```diff
@@ -72,11 +107,23 @@ export const uniqueId: RicosExtension = {
                 const shouldUpdate = nodeId !== id;
 
                 if (node.isBlock && shouldUpdate) {
-                  tr.setNodeMarkup(pos, undefined, {
-                    ...node.attrs,
-                    id,
-                    marks,
-                  });
+                  //fix bug in Android when textStyle is not preserved in new lines
+                  if (browser.android && node.attrs.style) {
```
maimonav, line 111: Please check the use case of adding image/ non text node container, in this current impl, the textStyle might be added to all nodes and not only to our desired nodes (text node containers).

packages/tiptap-extensions/web/src/unique-id.tsx
```diff
@@ -24,6 +26,39 @@ const handleAndroidEnterBug = (tr, pos) => {
   tr.setSelection(selection);
 };
 
+/*
+  This fix a bug with Android when textStyle is not preserved in new lines.
+  The styles arrived in node.attrs.style as string so we need to manually
+  create the textStyle object.
+*/
+const handleAndroidTextStyleEnterBug = (nodeStyle: string) => {
+  const style: { paddingTop?: string; paddingBottom?: string } = {};
```
maimonav, line 35: Please reuse these types from ricos-schema

packages/tiptap-extensions/web/src/unique-id.tsx
```diff
@@ -24,6 +26,39 @@ const handleAndroidEnterBug = (tr, pos) => {
...
+  const textStyle: { lineHeight?: string; textAlignment?: string } = {};
+  const stylesAsArray = nodeStyle.split(';').map(x => x.split(':'));
+  stylesAsArray.forEach(nodeStyle => {
```
maimonav, line 38: Since we are building this textStyle object from style string, let's make a converter function, and add unit tests to check edge cases.

packages/tiptap-extensions/web/src/unique-id.tsx
noamtwix: @maimonav I haven't found any problem with this (when your focus is on a plugin and you press enter the plugin is removed).
additionally, `node.attrs.style` is null.

maimonav (approved):

maimonav:
@noamtwix please add changelog :)

PR #4929, opened 2022-12-26
ricos getContentSummary

talevy17:


packages/common/web/src/Utils/ricosContentAnalytics.ts
```diff
@@ -86,7 +87,7 @@ export function getContentSummary(content: RichContent) {
     .map(node => {
       const { type } = node;
       const data = node[`${camelCase(type)}Data`];
-      const details: { type?: string; data?: Record<string, unknown> } = {};
+      const details: { type?: string; data?: any } = {};
```
maimonav, line 90: Why any? can we create a type for it?
What's the usage of this data field in content summary?

maimonav (changes requested):
Only one comment, the rest looks good 🔥

packages/common/web/src/Utils/ricosContentAnalytics.ts
talevy17: its passed as "PluginDetails" to the onPublish BI event.
the function we get from the consumer doesn't specify the data type though, so when I tried using other types the build crashed

talevy17: https://github.com/wix/ricos/blob/V10-viewer-deprecations/packages/ricos-types/web/src/biCallbacksTypes.ts#L149:L149

you can see here, thats the type we exported in the first place
changing it will trigger conflicts with wix-ricos as well @maimonav

maimonav: I see, of course we can update it since onPublish is an event that is being defined in the wix-ricos-common bi-strategy, but yeah let's keep it that way :)

maimonav (approved):
If there are no issues, let's merge it to the branch

PR #4965, opened 2023-01-09
[ricos-content] Fix fromDraft converter when handling sequential list nodes

domakas:
Sequential lists are rendered as nested in tip tap
image


lxgreen (approved):
good catch! how did you find this case?

domakas:
@lxgreen while trying to migrate to ViewerNext in Forum 😄 good thing I had post with all different content

PR #4977, opened 2023-01-11
[tiptap-editor] Remove onTransaction event subscription

viliusmileris:
Improves performance by solving multiple editor re-rerender issue.


lxgreen (approved):
looks legit

PR #4978, opened 2023-01-11
Draft js complete cleanup

oded-soffrin:


oded-soffrin:
This branch attempts to remove all draft-js references OTHER than ricos-content where it's still needed for converting
@danarish4 many occurrences in plugin, probably will be removed when done with plugin cleanup
@lxgreen some references are editor related

@talevy17 @tamrat-bay FYI

examples/main/src/text-color-style-fn.ts
```diff
@@ -1,4 +1,4 @@
-import type { EditorProps } from 'draft-js';
+import type { EditorProps } from 'draft-js'; //TODO
```
lxgreen, line 1: i guess this file can be deleted?

packages/ricos-content/web/package.json
```diff
@@ -71,7 +71,7 @@
     },
     "moduleNameMapper": {
       "\\.(jpg|jpeg|png|gif|svg)$": "<rootDir>/__mocks__/fileMock.js",
-      "\\.(css|scss)$": "<rootDir>/__mocks__/styleMock.js",
+      "\\.(css|scss)$": "<rootDir>/__mocks__/styleMock.js"
       "^draft-js$": "<rootDir>/../../../node_modules/@wix/draft-js",
       "^draft-js/": "<rootDir>/../../../node_modules/@wix/draft-js/"
```
lxgreen, line 76:
```suggestion
```

packages/ricos-content/web/src/comparision/compare.ts
```diff
@@ -1,6 +1,6 @@
 /** Based on https://gist.github.com/Yimiprod/7ee176597fef230d1451 */
 /* eslint-disable no-console, @typescript-eslint/no-explicit-any, fp/no-delete */
-import type { RawDraftEntityRange } from '@wix/draft-js';
+import type { RawDraftEntityRange } from '@wix/draft-js'; //TODO
```
lxgreen, line 3: still needed in compare as we use the `compare` in all the draft converter tests

examples/main/src/text-color-style-fn.ts
oded-soffrin: you mean `customStyleFn` is no longer supported ?

oded-soffrin: I see it's supported but the type is wrong, i've fixed it

PR #4979, opened 2023-01-11
[video] clean draft

danarish4:


packages/plugin-video/web/src/defaults.ts
```diff
@@ -8,3 +8,20 @@ export const DEFAULTS = {
     getVideoUrl: src => `https://video.wixstatic.com/${src.pathname}`,
   },
 };
+
+export const DEFAULT_CONFIG = Object.freeze({
```
tamrat-bay, line 12: delete this if convertNodeDataToDraft is removed

packages/plugin-video/web/src/createVideoData.ts
```diff
@@ -1,5 +1,5 @@
 import { merge } from 'lodash';
-import { DEFAULTS } from './video-component';
+import { DEFAULT_CONFIG } from './defaults';
```
tamrat-bay, line 2: change to RicosDefaults if convertNodeDataToDraft is removed

PR #4985, opened 2023-01-11
[tiptap-extensions] Don't add default dir value for text nodes

domakas:
Basic logic is this:
When blocks are in collapsible list, set direction by collapsible list direction option, unless paragraph has explicit text alignment set.


danarish4 (changes requested):
1. Pay attention that the bundle has grown dramatically
`Error: Packages that grew: ricos-editor-with-plugins: 571KB => 989KB`
[This file](https://github.com/wix/ricos/blob/master/bundle-analyzer/src/bundles/ricos-editor-with-plugins.tsx#L8) analyzes the editor bundle with some plugins consumption (one of which is the collapsible-list plugin).
I'm pretty sure this grew becomes because of this line
`import { getTextDirectionFromAlignment } from 'ricos-converters';`
Try to remove it and see if it helps.

2. The collapsible was hardcoded to search for `PARAGRAPH` and `HEADINGS`,
What about lists? What if a `FANCY_PARAGRAPH` node type will be added in the feature? The goal is to find nodes that belong to the group `text-container`.
This can be done by adding a `reconfigure` which will get all extensions installed from there, fetching all the `text-container` extensions, add to `addOptions`, and then it will be available on `this`
[Here](https://github.com/wix/ricos/blob/master/packages/ricos-common-plugins/web/src/text-align/extension.ts#L34) is an example

danarish4 (changes requested):
It looks good, but there is still a big growth in collapsible-list bundles
This is happening as a result of this line

`import { TextStyle_TextAlignment } from 'ricos-schema';`

Due to the lack of tree-shake, this import imports the entire rich-schema package, which is very large.
We need to find a solution to that, but until then, let's try to avoid this by removing this import

domakas:
@danarish4 tree-shaking is missing from bundleSize tests, right? This is what needs to be solved?

domakas:
Investigated a bit more: bundle size increase was due to `Decoration, DecorationSet` imports, which is not something we can change

danarish4 (approved):
🔥

PR #4998, opened 2023-01-12
[polls] cleanup draft

danarish4:


packages/plugin-social-polls/web/src/defaults.ts
```diff
@@ -3,12 +3,6 @@ export const LAYOUT = {
   GRID: 'grid',
 };
 
-export enum BACKGROUND_TYPE {
-  COLOR = 'color',
-  IMAGE = 'image',
-  GRADIENT = 'gradient',
-}
-
 export const LAYOUTS = {
```
tamrat-bay, line 6: `VISIBILITY_OPTIONS` and `LAYOUT` can be removed as well

PR #5018, opened 2023-01-18
[internal] Fix rollup build to enable tree-shaking

domakas:
Basically we have 5 types of entries:
- editor
- viewer
- viewerLoadable
- lib
- mobileNativeLoader

Each of them should support two formats: cjs and esm.

In esm entry points should support tree shaking, except for: mobileNativeLoader, viewerLoadable and viewer as they have side effects - they should support SSR, so their CSS is extracted to a separate file.

Editor and lib entries don't support SSR, so they have their CSS injected. For editor this is not a problem as it's lazy loaded anyway, for lib it shouldn't be a problem, but who knows. 🤷 
  
For cjs we just take each entry point for: index.ts (used as editor entry point), viewer.ts, viewerLoadable.ts, mobileNativeLoader.ts, build source to concatenated file while producing single cjs file for each. Extracting CSS with postcss plugin and adding CSS imports for each entry point (if needed).

For esm we do the same, for each entry point except for editor entry. For that one we build every source file separately with CSS injected. This way we can support tree shaking and avoid problems with CSS side effects.


dragonofdeath (approved):

rollup.config.common.ts
```diff
@@ -5,20 +5,53 @@ import { cloneDeep } from 'lodash';
...
+ * For cjs we just take each entry point for: index.ts (used as editor entry
+ * point), viewer.ts, viewerLoadable.ts, mobileNativeLoader.ts, build source to
+ * concatenated file while producing single cjs file for each. Extracting CSS
+ * with postcss plugin and adding CSS imports for each entry point (if needed).
+ *
+ * For esm we do the same, for each entry point except for editor entry. For
+ * that one we build every source file separately with CSS injected. This way we
+ * can support tree shaking and avoid problems with CSS side effects.
+ *
+ */
```
oded-soffrin, line 40: amazing summary

oded-soffrin:
amazing!

PR #5024, opened 2023-01-19
[conversion] - Gallery spacing, moved division from conversion to component

Dalcvi:
Currently if we have thumbnailSpacing at 5px in draftJs and we try to convert it to Rich content spacing, we get 5/2=2.5. This causes a few problems:
1. 2.5 doesn't follow our schema, because spacing is defined as integer.
2. 2.5 shows up in layout settings. This means we are lying to our users, because the spacing is actually 2 times bigger.

Fixed these problems by removing multiplication/division between both converters. Moved the division in tiptap version. Now, when in tiptap viewer we give options to Pro Gallery, we just do spacing/2, just like in draftJs version.

Also, the default spacing value for tiptap version was set to 2.5. Doubled it, to match draftJs version.


talevy17 (approved):
seems solid

PR #5040, opened 2023-01-26
[plugins] Fix font size resolver

dragonofdeath:
Fix issues when font size picker was incorrectly displaying
current font size in static toolbars
 * font size indication when editor just loaded
 * font size indication when setting mark
 * font size not being preserved in copy/paste inside editor
   - preservation is done only when copying inside editor - same as in
   draft version

Deduplicate font size resolution function when resolving in toolbar
  and editor command

Write basic integration tests for font size plugin
  * for this to happen made setup for editor/viewer jsdom based testing.
    Took inspiration and some code from remirror project
    Add content builders for test. Maybe later they can become general
    builders exposed to consumers?


packages/toolbars-v3/web/src/resolvers/tiptapResolvers.ts
```diff
@@ -352,75 +353,85 @@ export const getLineSpacingAfterSelectionResolver = TiptapContentResolver.create
...
-          return mark.type.name === Decoration_Type.FONT_SIZE;
-        });
+    const { from, $from, to, empty } = state.selection;
+    if (empty) {
+      // From stored marks
+      const storedFontSize = state.storedMarks?.find(
+        m => m.type.name === Decoration_Type.FONT_SIZE
+      );
+      if (storedFontSize) {
+        return storedFontSize.attrs.value ?? 0;
```
viliusmileris, line 386: Should the return type vary between string and number?

packages/toolbars-v3/web/src/resolvers/tiptapResolvers.ts
```diff
@@ -352,75 +353,85 @@ export const getLineSpacingAfterSelectionResolver = TiptapContentResolver.create
...
+      // From parent block node
+      const parent = state.doc.resolve(from).parent;
+      const ricosNode = fromTiptapNode({ ...parent.toJSON(), type: parent?.type?.name });
+      const decoration = styles.getDecoration(ricosNode as RichTextNode, Decoration_Type.FONT_SIZE);
+
+      if (decoration.fontSizeData?.value) {
+        const fontSizeInDocumentStyle = decoration.fontSizeData?.value;
+        return `${fontSizeInDocumentStyle}`;
+      }
+      // @TODO: should we throw an error here?
```
domakas, line 402: What would that mean? Would we break the editor? Not sure if this would good solution, we should try recovering/fallbacking as much as possible

packages/toolbars-v3/web/src/resolvers/tiptapResolvers.ts
dragonofdeath: Good catch! Why isn't typescript seeing this!??

packages/toolbars-v3/web/src/resolvers/tiptapResolvers.ts
dragonofdeath: Failing fast makes app better, but I'm not sure if its already a time for it

domakas (approved):

PR #5042, opened 2023-01-27
[ricos-content] Fix `fromDraft` converter for legacy font-color style

domakas:
While using `ensureRicosContent` it's not possible to pass `ignoreUnsupportedValues` to `fromDraft` converter. This breaks mobile app as converter throws exception. Added this optional proxy of params.

Also, allowed legacy font-color format conversion, when `style` is just hex color


oded-soffrin:
I believe that's not the only case that breaks due to unsupported value. Why not extend ensure function to use this param?

dragonofdeath (approved):

dragonofdeath (approved):

PR #5054, opened 2023-02-03
[tiptap-editor] Remove default editor padding on small devices

domakas:
Cherrypicked from https://github.com/wix/ricos/pull/5053 🍒


viliusmileris (approved):
screenshots need to be approved

PR #5058, opened 2023-02-07
[converters] - paragraph conversion

Dalcvi:
Testing paragraph conversion.

Found a bug with new line conversion. It just gets converted with other text and we do not support that. New line `(\n)` should have its own TEXT node. If it is with other text on editor update it will remove that illegal `\n`.


domakas (approved):
Seems legit

PR #5093, opened 2023-03-02
[] - Upgrade images to use WowImage

Dalcvi:
Giphy - improved by a lot. Giphy list of items now uses webp instead of .gif and it saves around 2x of data. Also uses WowImage, so half of the gifs dont load instantly on the list.
Giphy block used to have support for video .mp4 format and if video isn't available it went to .gif format. Problem? It checked if the url ends with .mp4 but usually url ends with `.mp4&ct=g`. So video was never actually supported.
Now we have this 
1. Video with .mp4
2. WowImage with .gif
3. img tag with .gif.
Using video saves around. 5-10x data.
Also upgraded giphy SDK as the old one was deprecated a long time ago.


domakas (approved):

domakas:
Code looks good, but it seems that visual regression found an issue in editor where gif size is smaller than in viewer

PR #5098, opened 2023-03-07
[plugin-collapsible-list] Fix drag and drop not working issue

viliusmileris:
Removed custom drag and drop handler plugin which was not working properly. Default tiptap DnD handler works fine, more info about it can be found here - https://tiptap.dev/guide/node-views/react#dragging.
Also, fixed issue where floating add plugin button was displayed during drag event.


integration-tests/drivers/builders.ts
```diff
@@ -121,6 +152,70 @@ export const text = (text: string, ...decorations: Decoration[]): TextNodeBuilde
   };
 };
 
+export const collapsibleList = (
+  ...args: (CollapsibleListItemItemBuilder | CollapsibleListData)[]
+): NodeBuilder => {
+  const collapsibleListData = args.find((a): a is CollapsibleListData => !('type' in a));
+
+  return {
+    type: 'simple_node',
```
dragonofdeath, line 161: collapsible_list?

dragonofdeath (approved):
The test look real nice!

integration-tests/drivers/builders.ts
viliusmileris: changed to block_node

PR #5099, opened 2023-03-08
[ricos-content] Refactor fromDraft video converters

dragonofdeath:
These converters are hard to grasp as the original object is mutated and
we can not know what fields are actually taken from original object and
which are just a side effect of this approach. Typescript fails here too
as the input object has to have both - input and output types.

Second issue is that current types do not cover all the input
possibilities/cover them incorrectly. So to help refactoring I took
500000 of samples of `entityMap` created in different times and generated
type definitions with [quicktype](https://quicktype.io/) tool.

Now its possible to refactor separate plugins to immutable/type safe
style and I already did that for Video/legacy video/audio plugins. Some
more complicated cases might require more tests/research to do that.


packages/ricos-content/src/converters/draft/fromDraft/convertRicosPluginData.ts
```diff
@@ -100,143 +107,140 @@ export const convertBlockDataToRicos = (type: string, data) => {
...
+          }
+        : undefined,
+      containerData,
+      title,
+      options,
+      disableDownload,
     };
   }
-  data.video && (data.video.duration = data.duration);
+  // throw new Error('Failed to convert video data - no src field');
```
viliusmileris, line 205: maybe it's worth to have console.error() ?

packages/ricos-content/src/converters/draft/fromDraft/__tests__/fromDraft.spec.ts
```diff
@@ -973,4 +975,12 @@ describe('migrate from draft', () => {
 
     expect(compare(fileData, expectedFileData)).toEqual({});
   });
+
+  it('crash case 1', () => {
```
viliusmileris, line 979: poor naming

packages/ricos-content/src/converters/draft/fromDraft/convertRicosPluginData.ts
```diff
@@ -100,143 +107,140 @@ export const convertBlockDataToRicos = (type: string, data) => {
...
-  };
-  video;
-  thumbnail;
-  title?;
-  duration?: number;
-}) => {
+const convertVideoData = (data: WixDraftPluginVideo): VideoData => {
+  const duration = typeof data.duration === 'number' ? data.duration : undefined;
+  const containerData = convertContainerData({ config: data.config, src: data.src }, VIDEO_TYPE);
+  const title = data.metadata?.title || data.metadata?.title;
```
viliusmileris, line 152: ?

packages/ricos-content/src/converters/draft/fromDraft/convertRicosPluginData.ts
dragonofdeath: I'm not sure, leaving behaviour as it was

packages/ricos-content/src/converters/draft/fromDraft/__tests__/fromDraft.spec.ts
dragonofdeath: improved!

packages/ricos-content/src/converters/draft/fromDraft/convertRicosPluginData.ts
dragonofdeath: thanks, fixed!

viliusmileris (approved):

dragonofdeath:
In the end I've found out that the plugin converters are actually used as 'ensurers' - either tiptap or draft plugin data is passed to them. It would be possible to handle internal usages, but as they are publicly available( they are used in react native editor), so that makes this change unpredictable. Closing in favor of https://github.com/wix/ricos/pull/5104

